PR #1, opened 2019-12-17
Working prototype of pulling modules from OCI registries

radu-matei:


examples/greet.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: greet
 spec:
   containers:
-    - image: technosophos/greet
+    - image: cnabregistry.azurecr.io/wasm:v2
```
radu-matei: This should be a separate example.

src/main.rs
```diff
@@ -22,4 +22,11 @@ fn main() -> Result<(), failure::Error> {
     let provider = WasmRuntime {};
     let kubelet = Kubelet::new(provider, kubeconfig, namespace);
     kubelet.start(address)
+
+    // oci::pull_wasm(
+    //     String::from("cnabregistry.azurecr.io/wasm-to-oci:v2"),
+    //     String::from("test.wasm"),
```
radu-matei: Remove this.

src/oci.rs
```diff
@@ -0,0 +1,247 @@
...
+        let func: lib::Symbol<unsafe extern "C" fn(r: GoString, f: GoString) -> i64> =
+            lib.get(b"Pull")?;
+        match func(go_str_ref, go_str_file) {
+            0 => return Ok(()),
+            _ => return Err(OCIError::Custom(String::from("cannot pull module"))),
+        }
+    }
+}
+
+pub fn push_wasm(reference: &str, file: &str) -> lib::Result<i64> {
```
radu-matei: Thinking about this a bit more, we probably don't need a push function for the node.

src/wasm.rs
```diff
@@ -38,11 +39,13 @@ impl Provider for WasmRuntime {
...
             .namespace
             .unwrap_or_else(|| "default".into());
-        // Start with a hard-coded WASM file
-        let data = std::fs::read("./examples/greet.wasm")
-            .expect("greet.wasm should be in examples directory");
-        pod_status(client.clone(), pod.clone(), "Running", namespace.as_str());
+        let first_container = pod.spec.containers[0].clone();
+
+        let module_ref = pod.spec.containers[0].image.as_ref().unwrap().to_owned();
+        pull_wasm(module_ref, String::from("pulled.wasm")).unwrap();
```
radu-matei: Use the real module name instead of `pulled.wasm`.

radu-matei: For now we pull the module into `pulled.wasm`.
We don't have a central location for all pulled modules, and we always pull the module from a registry.

We should decide on a cache directory for modules.

src/oci.rs
```diff
@@ -0,0 +1,226 @@
+use std::ffi::CString;
+
+extern crate libloading as lib;
+
+fn lib_file() -> String {
```
rylev: This can just be `const lib_file: &str` instead of a function

src/oci.rs
```diff
@@ -0,0 +1,226 @@
...
+    };
+    let go_str_file = GoString {
+        p: c_file.as_ptr(),
+        n: c_file.as_bytes().len() as isize,
+    };
+
+    let lib = lib::Library::new(lib_file())?;
+    unsafe {
+        let func: lib::Symbol<unsafe extern "C" fn(r: GoString, f: GoString) -> i64> =
+            lib.get(b"Pull")?;
```
rylev: Perhaps unwrapping here would be better? If the `Pull` symbol isn't available, that's a pretty serious dev mistake.

src/wasm.rs
```diff
@@ -38,11 +39,13 @@ impl Provider for WasmRuntime {
             .clone()
             .namespace
             .unwrap_or_else(|| "default".into());
-        // Start with a hard-coded WASM file
-        let data = std::fs::read("./examples/greet.wasm")
-            .expect("greet.wasm should be in examples directory");
-        pod_status(client.clone(), pod.clone(), "Running", namespace.as_str());
+        let first_container = pod.spec.containers[0].clone();
```
rylev: Why clone?

src/oci.rs
radu-matei: I think it would be better to use static linking here.

shared/libwasm2oci.go
```diff
@@ -0,0 +1,33 @@
...
+	"C"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/engineerd/wasm-to-oci/pkg/oci"
+)
+
+//export Pull
+func Pull(ref, outFile string) int64 {
+	err := oci.Pull(ref, outFile)
```
bacongobbler: this can be optimized as

```
if err := oci.Pull(ref, outFile); err != nil {
    ...
}
```

shared/libwasm2oci.go
```diff
@@ -0,0 +1,33 @@
...
+		log.Infof("cannot pull module: %v", err)
+		return 1
+	}
+
+	return 0
+}
+
+//export Push
+func Push(ref, mod string) int64 {
+	err := oci.Push(ref, mod)
```
bacongobbler: same here

```
if err := oci.Push(ref, mod); err != nil {
    ...
}
```

src/wasm.rs
bacongobbler: We can port the ImageStore crate over eventually for storing modules.

radu-matei:
Closing in favor of #48

PR #11, opened 2020-03-02
fix(*): Fixes various Clippy issues and adds an Action pipeline

thomastaylor312:
Closes #10

PR #12, opened 2020-03-02
fix(*): Addresses some follow up from the original PR

thomastaylor312:
Just a few suggestions that @rylev commented on #11 after merge

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
       - name: Build
         run: |
           cargo fmt --all -- --check
-          cargo install just
-          just build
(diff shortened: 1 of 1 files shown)


technosophos:
What is the rationale for changing `to_string()` to `to_owned()`? One is fairly informative, telling readers that the expected type is `String`, so we are changing it accordingly. The other is not terribly informative, as it doesn't necessarily indicate what type we're changing to. I thought this is why we used `to_string()` instead of `into()`.

thomastaylor312:
For reference, I removed the `to_string` changes for now and switched to getting `just` from GH releases

Issue #13, opened 2020-03-02
Listen on all namespaces


technosophos:
When I first created Krustlet, I had it only listening for Pod events on one namespace (since that worked fine for testing). I think this should listen on all namespaces. See `kubelet.rs`

bacongobbler:
After diving into the kubelet code, it looks like it's listening for pods across all namespaces. The pod informer listens for any pod events across all namespaces, so we can pull out some of that namespaced code.

PR #14, opened 2020-03-02
add a few inline docs to the code

technosophos:
This just adds a little bit of inline docs in places that I felt I hadn't covered before.

See also #13, which I filed while writing this.


src/kubelet.rs
```diff
@@ -21,6 +21,8 @@ use std::collections::HashMap;
 pub struct NotImplementedError;
```
radu-matei: The first line in this file has a spelling error: _contans_.

technosophos:
Rebased and fixed `contans`

PR #27, opened 2020-03-04
Default `run` task

rylev:


crates/kubelet/src/lib.rs
```diff
@@ -9,4 +9,4 @@ pub mod node;
 pub mod pod;
 mod server;
 
-pub use self::kubelet::*;
+pub use kubelet::*;
```
thomastaylor312: Hmm...this gave me clippy warnings before

PR #29, opened 2020-03-04
remove wasm and wasi crate dependencies

bketelsen:
wascc-provider doesn't need wasi or wasmtime deps.


thomastaylor312:
I agree with @rylev here. There should only be a lock file at the top level

bketelsen:
interesting.  OK removing!

bketelsen:
> I agree with @rylev here. There should only be a lock file at the top level

There is a only cargo.lock at the top level, is that what you're suggesting I remove?

thomastaylor312:
Oh, I misread that. Actually @rylev are you sure? Our top level has the need for a lock file due to the things we are building

rylev:
@bketelsen sorry, misread the path. All is good!

PR #31, opened 2020-03-04
remove hello world example

bacongobbler:
It doesn't appear as though this can actually be used with Krustlet. `examples/greet.yaml` is the canonical example to run a WASI workload with Krustlet.


bacongobbler:
It's also in the root of the repository... eww

PR #32, opened 2020-03-04
Update wascc-host to 0.5.0

bketelsen:


crates/wascc-provider/src/lib.rs
```diff
@@ -108,7 +108,7 @@ impl Provider for WasccProvider {
             .get(ACTOR_PUBLIC_KEY)
             .map(|a| a.to_string())
             .unwrap_or_else(|| "".into());
-
+        debug!("{:?}", pubkey);
```
thomastaylor312: Did you want this to be echoed during debug?

bketelsen: removed it, but it is really nice seeing the public key of the actor as you go through debugging

bketelsen:
fixes #30

bketelsen:
removing tests - captured in #36

Issue #33, opened 2020-03-04
Fully implement kubelet logs path


thomastaylor312:
The current implementation in the `get_container_logs` function is incomplete. This should parse the proper data from the request and call the provider

PR #35, opened 2020-03-04
feat(wasi_provider): Implements logging for wasi

### LINKED ISSUE #23: Implement logging for WASI
Logging is all set up, but it is not glued together with the log method for the provider

thomastaylor312:
This implements logging for the wasi provider. The actual `logs` function
is untested as there isn't an easy way to do it outside of integration
testing. Also, the logging logic from kubelet isn't fully hooked up yet
(see #33).

Please also note that I made some changes to the logs method on the Provider
trait to more clearly reflect the data available (essentially just the path
from the GET request).

Closes #23


crates/wasi-provider/src/lib.rs
```diff
@@ -141,33 +141,54 @@ impl Provider for WasiProvider {
...
-        // combined stream of stdout and stderr into our hashmap so we can
-        // expose them here
-        unimplemented!()
+    fn logs(
+        &self,
+        namespace: String,
+        pod: String,
+        container: String,
+    ) -> Result<Vec<String>, failure::Error> {
+        let mut handles = self.handles.write().unwrap();
```
thomastaylor312: We need a mutable handle in order to read from the buffer below. If there is some way to do this without a write lock, please let me know

technosophos: If it advances a reader, it is a write operation. So this is likely correct.

crates/wasi-provider/src/lib.rs
```diff
@@ -205,45 +224,49 @@ impl WasiRuntime {
...
-            dirs,
-            stdout,
-            stderr,
-        })
+        //
+        // As warned in the BufReader docs, creating multiple BufReaders on the
+        // same stream can cause data loss. So reopen a new file object so we
+        // don't drop any data
+        let output = NamedTempFile::new_in(&log_file_location)?;
+        let file_handle = output.reopen()?;
```
technosophos: When you create a new temp file, it does not appear to be wrapped in a buffered reader. And `output` is only used once here. So re-opening seems to be unnecessary. The BufReader docs just warn that you shouldn't have two buffered readers on the same file handle because they can independently advance the underlying reader.

crates/wasi-provider/src/lib.rs
```diff
@@ -205,45 +224,49 @@ impl WasiRuntime {
         args: Vec<String>,
         dirs: HashMap<String, Option<String>>,
         log_file_location: L,
-    ) -> Result<Self, failure::Error> {
-        debug!("before parse file: {:?}", std::env::current_dir().unwrap());
+    ) -> Result<(Self, BufReader<File>), failure::Error> {
```
technosophos: I feel uncomfortable with this change.

A constructor isn't generally intended to return a Self _wrapped in another thing_. In fact, that seems contrary to any constructor pattern I've seen. The idea of a constructor is that it creates _an instance of the object_, not an instance of some other thing with an instance of the object embedded somewhere in there.

In addition, returning a file handle seems awkward, since it commits you to having open file handles before the WASM itself has even been run. So you are initializing an object (that consumes system resources) before the system is asking you to commit to running this thing. This very well may paint us into a design corner in the future, where we have heavy objects hanging around early in the game.

When trying this same thing with wasm3, it appears that one can delay creating a file handle until `run()` is called, which means that we don't get a penalty imposed on us until the module is actually executed.

crates/wasi-provider/src/lib.rs
thomastaylor312: This is the file handle for the returned BufReader (which I think you'd like to change anyway). Otherwise I have already moved `output` when I put it in the struct. But let's see how it looks after I address the others

crates/wasi-provider/src/lib.rs
thomastaylor312: Yeah, I wasn't sure myself either. The problem right now is that there is no way to get the handle once we kick things off because the `runtime` is moved onto the thread, so I couldn't call the `output` method anymore. I think the way to do this would be returning the file handle on run along with a future because right now `run` is a blocking call. I'll take another look tomorrow with fresh eyes to see what other ways there are to tackle this

crates/wasi-provider/src/lib.rs
```diff
@@ -107,9 +107,9 @@ impl Provider for WasiProvider {
         let handle = std::thread::spawn(move || runtime.run().unwrap());
         let mut handles = self.handles.write().unwrap();
```
technosophos: `let output = runtime.output()`

thomastaylor312: Runtime has been moved into the thread though

crates/wasi-provider/src/lib.rs
technosophos: To get around the constructor issue, you can just call `let output = runtime.output()`, right? I left a comment inline to show you what I mean.

thomastaylor312:
Ok, so I cleaned things up to use a file we create at module runtime instead, which we then maintain a handle to. Let me know if you like it better

crates/wasi-provider/src/lib.rs
```diff
@@ -199,51 +218,45 @@ impl WasiRuntime {
...
             env,
             args,
             dirs,
-            stdout,
-            stderr,
         })
     }
 
-    fn run(&self) -> Result<(), failure::Error> {
+    fn run(&self, output: File) -> Result<(), failure::Error> {
```
technosophos: I like this apprroach MUCH more!

PR #37, opened 2020-03-04
WIP: add wasm3 runtime

technosophos:
This adds [wasm3](https://github.com/wasm3/wasm3) support via [wasm3-rs](https://github.com/Veykril/wasm3-rs).

The Rust bindings do not have enough of WASI implemented yet. But it is under development. Right now, it can execute a module in-place, but not send anything in via args, env vars, or files. It also cannot directly access the output (which is written directly to FD 1 from C).


technosophos:
I think I am further going to hold work on this until async lands.

rylev:
I'm going to close this for now. We can reopen it when @technosophos is ready to start the work on it again.

Issue #5, opened 2020-02-28
Lease creation fails


bketelsen:
```[2020-02-28T20:11:41Z INFO  krustlet::node] created node just fine
[2020-02-28T20:11:41Z ERROR krustlet::node] Failed to create lease: ApiError BadRequest ("Lease in version \"v1\" cannot be handled as a Lease: v1.Lease.Spec: v1.LeaseSpec.RenewTime: unmarshalerDecoder: parsing time \"2020-02-28T20:11:41.636526400Z\" as \"2006-01-02T15:04:05.000000Z07:00\": cannot parse \"400Z\" as \"Z07:00\", error found in #10 byte of ...|36526400Z\"}}|..., bigger context ...|\":300,\"renewTime\":\"2020-02-28T20:11:41.636526400Z\"}}|...")
[2020-02-28T20:11:41Z INFO  krustlet::server] starting webserver at: V4(0.0.0.0:3000)
[2020-02-28T20:11:42Z ERROR krustlet::node] Failed to create lease: ApiError NotFound ("leases.coordination.k8s.io \"krustlet\" not found")```

**Comment by technosophos** [comment]
What Kubernetes distro are you using? I think I mainly tested on Kind

**Comment by technosophos** [comment]
THat date gets set in `node.rs`:

```rust
/// Defines a new coordiation lease for Kubernetes
///
/// We set the lease times, the lease duration, and the node name.
fn lease_spec_definition() -> LeaseSpec {
    LeaseSpec {
        holder_identity: Some(NODE_NAME.to_string()),
        acquire_time: Some(MicroTime(Utc::now())),
        renew_time: Some(MicroTime(Utc::now())),
        lease_duration_seconds: Some(300),
        ..Default::default()
    }
}
```

bketelsen:
It was whatever the default on AKS is as of yesterday.   I’ll look in a bit.

Brian

On Mar 2, 2020, at 4:14 PM, Matt Butcher <notifications@github.com> wrote:

﻿

What Kubernetes distro are you using? I think I mainly tested on Kind

bketelsen:
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"c52f59bbba5fbf21fbb18e9a06f96e563fe4c20a", GitTreeState:"clean", BuildDate:"2020-01-31T20:00:26Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

[...]

PR #40, opened 2020-03-05
First pass at asyncifying

rylev:
Not only does this make most things async, but it also now uses the latest version of `kube`

PR #41, opened 2020-03-05
Implement workaround for LeaseSpec serialization issue.

peterhuene:
See #5 relating to the issue this attempts to workaround.

Rather than serialize a `LeaseSpec` which has a variable sized timestamp
formatting, this forces the timestamps to the six digits of subsecond precision
currently required by the k8s API.

### DIFF
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -1,7 +1,7 @@
 use chrono::prelude::*;
-use k8s_openapi::api::coordination::v1::{Lease, LeaseSpec};
+use k8s_openapi::api::coordination::v1::Lease;
 use k8s_openapi::api::core::v1::Node;
-use k8s_openapi::apimachinery::pkg::apis::meta::v1::{MicroTime, Time};
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::{
     api::{Api, PatchParams, PostParams},
     client::APIClient,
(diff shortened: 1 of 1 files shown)


peterhuene:
Although it looks like to take the fix from `k8s-openapi` we might need to also bump `kube` and there's been breaking API changes from the version we're currently using.

radu-matei:
I think the easiest way forward would be to update #40 to use this patch.

rylev:
Let's land #40 first and then see where we are.

thomastaylor312:
#40 is merged. I still think we should probably use this fix as I don't know when a new version of `kube` will be released and I don't think we should be using master just for this

bacongobbler:
I'm +1 on going forward with merging this. Would it make sense to add an additional comment under `lease_spec_definition` to remove this code once we update k8s-openapi? That way in case it gets lost we can always reference back to this and pull it out safely.

radu-matei:
Wait - why not update `k8s-openapi` to the revision that specifically fixes our issue?

bacongobbler:
Oh! Is that release out now?

radu-matei:
There isn't a release yet - but there is a fix in `master` that specifically solves our issue - https://github.com/Arnavion/k8s-openapi/commit/35cbddfeb26726bc1a0c882a40584c262fb53726

bacongobbler:
Okay. I'm of the opinion that I'd like to have something that works so we can move forward and test things out on master. Once the new crate version is available, we can update krustlet and remove this section of code.

Does that seem fair?

bacongobbler:
We could also pin to the master branch of `k8s-openapi`, but this seems a little less risky to introduce new issues.

radu-matei:
Ok, then let's merge this now and update once the crate has a new release.

Thanks a lot, @peterhuene!

PR #45, opened 2020-03-06
feat(kubelet): Adds ability to configure nodes

### LINKED ISSUE #34: Proper node configuration
Right now, Krustlet uses a bunch of hardcoded values for creating the `Node` object in Kubernetes (such as the IP address and hostname). We should follow similar patterns to normal kubelet for configuring these values, so part of this task will be investigating how kubelet does it in mainline Kubernetes.

thomastaylor312:
This follows similar defaulting methodology as to what is found in the
normal Kubernetes kubelet. Currently we only support command line flags
and environment variables but it allows a user to configure various parts
of the kubelet

Closes #34

### DIFF
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -26,5 +26,9 @@ tokio  = { version = "0.2.13" }
 kube = { version = "0.28.0"  }
 k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_17"] }
 chrono = { version = "0.4", features = ["serde"] }
-clap = { git = "https://github.com/clap-rs/clap", features = ["wrap_help"] }
+clap = { git = "https://github.com/clap-rs/clap", features = ["wrap_help"], optional = true }
 hostname = "^0.3"
+
+[features]
+default = []
+cli = ["clap"]
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -26,3 +26,5 @@ tokio  = { version = "0.2.13" }
 kube = { version = "0.28.0"  }
 k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_17"] }
 chrono = { version = "0.4", features = ["serde"] }
+clap = { git = "https://github.com/clap-rs/clap", features = ["wrap_help"] }
+hostname = "^0.3"
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -1,12 +1,15 @@
-use clap::derive::Clap;
+#[cfg(feature = "cli")]
+use clap::derive::{FromArgMatches, IntoApp};
 use std::net::IpAddr;
 use std::net::ToSocketAddrs;
 
+const DEFAULT_PORT: u16 = 3000;
+
 /// The configuration needed for a kubelet to run properly. This can be
 /// configured manually in your code or if you are exposing a CLI, use the
-/// [get_from_flags method](get_from_flags). Use
-/// [default_config](Config::default_config) to generate a config with all of
-/// the default values set.
+/// [get_from_flags method](get_from_flags) (this requires the `cli` feature to
+/// be enabled). Use [default_config](Config::default_config) to generate a
+/// config with all of the default values set.
 #[derive(Clone)]
 pub struct Config {
     pub addr: IpAddr,
@@ -35,17 +38,51 @@ impl Config {
                 IpAddr::V4(_) => "0.0.0.0".parse().unwrap(),
                 IpAddr::V6(_) => "::".parse().unwrap(),
             },
-            port: 3000,
-            node_ip: default_node_ip(&hostname, preferred_ip_family)?,
+            port: DEFAULT_PORT,
+            node_ip: default_node_ip(&mut hostname.clone(), preferred_ip_family)?,
             node_name: sanitize_hostname(&hostname),
             hostname,
             arch: String::default(),
         })
     }
+    /// Parses all command line flags and sets the proper defaults. The version
+    /// of your application should be passed to set the proper version for the
+    /// CLI
+    #[cfg(feature = "cli")]
+    pub fn new_from_flags(version: &str) -> Self {
+        // TODO: Support config files too. config-rs and clap don't just work
+        // together, so there is no easy way to merge together everything right
+        // now. This function is here so we can do that data massaging and
+        // merging down the road
+        let app = Opts::into_app().version(version);
+        let opts = Opts::from_arg_matches(&app.get_matches());
+        // Copy the addr to avoid a partial move when computing node_ip
+        let addr = opts.addr;
+        let hostname = opts
+            .hostname
+            .unwrap_or_else(|| default_hostname().expect("unable to get default hostname"));
+        let node_ip = opts.node_ip.unwrap_or_else(|| {
+            default_node_ip(&mut hostname.clone(), &addr)
+                .expect("unable to get default node IP address")
+        });
+        Config {
+            addr,
+            port: 3000,
+            node_ip,
+            node_name: sanitize_hostname(&hostname),
+            hostname,
+            arch: String::default(),
+        }
+    }
 }
 
 // Opts contains the values that can be configured for kubelet
 #[derive(clap::Clap, Clone, Debug)]
+#[cfg(feature = "cli")]
+#[clap(
+    name = "krustlet",
+    about = "A kubelet for running WebAssembly workloads"
+)]
 pub struct Opts {
     #[clap(
         short = "a",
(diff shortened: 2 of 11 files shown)


crates/kubelet/src/node.rs
```diff
@@ -132,21 +128,19 @@ async fn update_lease(node_uid: &str, client: &APIClient) {
 ///
 /// TODO: A lot of the values here are faked, and should be replaced by real
```
rylev: Is this still true? Looks like we're largely not hardcoding things anymore.

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[clap(
+        long = "node-name",
+        env = "KRUSTLET_NODE_NAME",
+        help = "The name for this node in Kubernetes, defaults to the hostname of this machine"
+    )]
+    node_name: Option<String>,
+}
+
+/// Parses all command line flags and sets the proper defaults
+pub fn get_from_flags() -> Config {
```
rylev: Should this be on the config struct itself? `Config::new_from_flags` reads nicer.

crates/kubelet/src/kubelet.rs
```diff
@@ -67,33 +68,39 @@ pub struct Status {
...
     /// This will listen on the given address, and will also begin watching for Pod
     /// events, which it will handle.
-    pub async fn start(&self, address: std::net::SocketAddr) -> Result<(), failure::Error> {
+    pub async fn start(&self) -> Result<(), failure::Error> {
         self.provider.lock().await.init().await?;
         let client = APIClient::new(self.kubeconfig.clone());
         // Create the node. If it already exists, "adopt" the node definition
-        create_node(&client, &self.provider.lock().await.arch()).await;
+        let mut conf = self.config.clone();
+        conf.arch = self.provider.lock().await.arch();
```
rylev: Seems weird that we have an `arch` field that is normally set to "" that we have to remember to update with the arch from provider. Maybe passing around `arch` separately makes more sense?

crates/kubelet/Cargo.toml
```diff
@@ -26,3 +26,5 @@ tokio  = { version = "0.2.13" }
 kube = { version = "0.28.0"  }
 k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_17"] }
 chrono = { version = "0.4", features = ["serde"] }
+clap = { git = "https://github.com/clap-rs/clap", features = ["wrap_help"] }
```
rylev: Will users universally configure from the command line? If not, we should think about putting this behind a feature flag.

crates/kubelet/src/node.rs
```diff
@@ -132,21 +128,19 @@ async fn update_lease(node_uid: &str, client: &APIClient) {
...
 /// numbers post-POC.
-fn node_definition(arch: &str) -> serde_json::Value {
-    let pod_ip = "10.21.77.2";
-    let port = 3000;
+fn node_definition(config: Config) -> serde_json::Value {
```
radu-matei: Could we use the typed object from the API for a `Node` here?

https://github.com/Arnavion/k8s-openapi/blob/05e60741985f5a5c3c5da255fee93aec16157ed7/src/v1_17/api/core/v1/node.rs#L3-L14

(Not blocking this PR though.)

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+        node_name: sanitize_hostname(&hostname),
+        hostname,
+        arch: String::default(),
+    }
+}
+
+fn default_hostname() -> Result<String, failure::Error> {
+    Ok(hostname::get()?
+        .into_string()
+        .map_err(|_| format_err!("invalid hostname string"))?)
```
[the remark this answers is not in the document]
thomastaylor312: The documentation I was reading seemed to specify there could be other failure methods from an os string, but maybe I misread

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Some hostnames (particularly local ones) can have uppercase letters, which is
+// disallowed by the DNS spec used in kubernetes naming. This sanitizes those
+// names
+fn sanitize_hostname(hostname: &str) -> String {
+    // TODO: Are there other sanitation steps we should do here?
+    hostname.to_owned().to_lowercase()
```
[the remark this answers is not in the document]
thomastaylor312: Good catch. I totally glossed over the fact the `to_lowercase` method returns a new String

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+    Ok(hostname::get()?
+        .into_string()
+        .map_err(|_| format_err!("invalid hostname string"))?)
+}
+
+// Some hostnames (particularly local ones) can have uppercase letters, which is
+// disallowed by the DNS spec used in kubernetes naming. This sanitizes those
+// names
+fn sanitize_hostname(hostname: &str) -> String {
+    // TODO: Are there other sanitation steps we should do here?
```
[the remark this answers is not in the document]
thomastaylor312: This is just an open question for people to consider down the line more than a specific issue

crates/kubelet/src/config.rs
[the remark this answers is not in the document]
rylev: From the docs: "Converts the `OsString` into a `String` if it contains valid Unicode data."

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+    // NOTE: As of right now, we don't have cloud providers. In the future if
+    // that is the case, we will need to add logic for looking up the IP and
+    // hostname using the cloud provider as they do in the kubelet
+    // To use the local resolver, we need to add a port to the hostname. Doesn't
+    // matter which one, it just needs to be a valid socket address
+    let mut with_port = hostname.to_owned();
+    with_port.push_str(":80");
+    Ok(with_port
+        .to_socket_addrs()?
+        .find(|i| !&i.ip().is_loopback() && is_same_ip_family(&i.ip(), preferred_ip_family))
```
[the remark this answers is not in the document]
thomastaylor312: Nope! Fixed

crates/kubelet/src/config.rs
thomastaylor312: Yes it should. I had it on `Opts` first and then took it out and didn't move it again

crates/kubelet/src/kubelet.rs
thomastaylor312: I couldn't figure out a nice way to do this originally. The problem is I don't want a user to have to know or set the arch manually (even with a constant). Is there a way I can do this cleanly?

crates/kubelet/Cargo.toml
thomastaylor312: I like this idea. Let me figure out how to do it, because I haven't before

crates/kubelet/src/node.rs
thomastaylor312: This personally reads a little cleaner as is instead of having to deal with converting and cloning strings

crates/kubelet/src/kubelet.rs
rylev: is `create_node` supposed to be public? i.e., will a user call it?

crates/kubelet/src/config.rs
```diff
@@ -0,0 +1,159 @@
...
+// 1. Lookup the IP from node name by DNS
+// 2. Try to get the IP from the network interface used as default gateway
+//    (unimplemented for now because it doesn't work across platforms)
+fn default_node_ip(hostname: &str, preferred_ip_family: &IpAddr) -> Result<IpAddr, failure::Error> {
+    // NOTE: As of right now, we don't have cloud providers. In the future if
+    // that is the case, we will need to add logic for looking up the IP and
+    // hostname using the cloud provider as they do in the kubelet
+    // To use the local resolver, we need to add a port to the hostname. Doesn't
+    // matter which one, it just needs to be a valid socket address
+    let mut with_port = hostname.to_owned();
```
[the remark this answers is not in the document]
thomastaylor312: Fixed

crates/kubelet/src/config.rs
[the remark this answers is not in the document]
thomastaylor312: Nope! turns out it isn't

crates/kubelet/src/kubelet.rs
thomastaylor312: `create_node` should not be public, so I'll go ahead and remove the `pub` modifier

thomastaylor312: That module shouldn't be. I went ahead and removed it. I also just had the underlying functions in the node module accept arch as an argument and removed it from the config

crates/kubelet/src/node.rs
radu-matei: It's fine for now, and there's no point in holding this PR.
But if we're using the typed library, we should use it consistently.

PR #47, opened 2020-03-09
Less blocking

### LINKED ISSUE #21: Making things Async
Bringing up the discussion question of: Should we make kubelet and all providers async? This would involve using tokio as it is the underlying runtime for wasmtime as well.

rylev:
Removes more instances of blocking that will gum up the tokio runtime. 

This fixes #21

### DIFF
--- a/crates/wascc-provider/src/lib.rs
+++ b/crates/wascc-provider/src/lib.rs
@@ -33,10 +33,14 @@ pub struct WasccProvider {}
 #[async_trait::async_trait]
 impl Provider for WasccProvider {
     async fn init(&self) -> Result<(), failure::Error> {
-        let data = NativeCapability::from_file(HTTP_LIB)
-            .map_err(|e| format_err!("Failed to read HTTP capability {}: {}", HTTP_LIB, e))?;
-        host::add_native_capability(data)
-            .map_err(|e| format_err!("Failed to load HTTP capability: {}", e))
+        tokio::task::spawn_blocking(|| {
+            let data = NativeCapability::from_file(HTTP_LIB)
+                .map_err(|e| format_err!("Failed to read HTTP capability {}: {}", HTTP_LIB, e))?;
+            host::add_native_capability(data)
+                .map_err(|e| format_err!("Failed to load HTTP capability: {}", e))
+        })
+        .await
+        .expect("Could not spawn worker for initializing WasccProvider")
     }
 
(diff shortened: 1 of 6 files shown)


crates/wascc-provider/src/lib.rs
```diff
@@ -33,10 +33,14 @@ pub struct WasccProvider {}
...
-        host::add_native_capability(data)
-            .map_err(|e| format_err!("Failed to load HTTP capability: {}", e))
+        tokio::task::spawn_blocking(|| {
+            let data = NativeCapability::from_file(HTTP_LIB)
+                .map_err(|e| format_err!("Failed to read HTTP capability {}: {}", HTTP_LIB, e))?;
+            host::add_native_capability(data)
+                .map_err(|e| format_err!("Failed to load HTTP capability: {}", e))
+        })
+        .await
+        .expect("Could not spawn worker for initializing WasccProvider")
```
thomastaylor312: This should probably just return the error and let the caller handle it

crates/wasi-provider/src/lib.rs
```diff
@@ -1,20 +1,22 @@
...
+use tokio::io::AsyncBufReadExt;
+use tokio::io::BufReader;
+use tokio::stream::StreamExt;
 use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
-use wasi_common::preopen_dir;
-use wasmtime_wasi::old::snapshot_0::Wasi as WasiUnstable;
-use wasmtime_wasi::{Wasi, WasiCtxBuilder};
+
+use wasi_runtime::WasiRuntime;
```
thomastaylor312: Thanks for pulling this out!

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// * `log_file_location` - location for storing logs
+    pub fn new<M: AsRef<Path>>(
+        module_path: M,
+        env: HashMap<String, String>,
+        args: Vec<String>,
+        dirs: HashMap<String, Option<String>>,
+    ) -> Result<Self, failure::Error> {
+        let module_data = wat::parse_file(module_path)?;
+
+        // We need to use named temp file because we need multiple file handles
```
thomastaylor312: I think I forgot to move this comment to where we are constructing the temp files in the Provider as it no longer makes sense here

PR #48, opened 2020-03-09
initial add of OCI Distribution crate

technosophos:
This PR adds a crate for OCI Distribution. It has a working `version` call, and a `pull_manifest` call that hits the correct endpoint. But we need AuthZ implemented before that one will work. I have begun working on that.

We can either merge this now so that others can help, or I can keep pushing it along for a while until we get to something that meets our requirement for viability.

### DIFF
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "oci-distribution"
+version = "0.1.0"
+authors = [
+    "Matt Butcher <matt.butcher@microsoft.com>",
+    "Matthew Fisher <matt.fisher@microsoft.com>",
+    "Radu Matei <radu.matei@microsoft.com>",
+    "Taylor Thomas <taylor.thomas@microsoft.com>",
+    "Brian Ketelsen <Brian.Ketelsen@microsoft.com>",
+    "Brian Hardock <Brian.Hardock@microsoft.com>",
+    "Ryan Levick <rylevick@microsoft.com>",
+]
+edition = "2018"
+
+[dependencies]
+reqwest = { version = "0.10", features = ["json"] }
+anyhow = "1.0.25"
+tokio = { version = "0.2", features = ["full"] }
+serde = "1.0"
+serde_json = "1.0"
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -13,9 +13,9 @@ authors = [
 edition = "2018"
 
 [dependencies]
-reqwest = { version = "0.10", features = ["json"] }
+reqwest = { version = "0.10.4", features = ["json"] }
 failure = "0.1"
-tokio = { version = "0.2", features = ["full"] }
+tokio = "0.2.13"
 serde = "1.0"
 serde_json = "1.0"
 www-authenticate = "0.3.0"
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -18,7 +18,6 @@ anyhow = "1.0.25"
 tokio = { version = "0.2", features = ["full"] }
 serde = "1.0"
 serde_json = "1.0"
-async-oauth2 = "0.1.1"
 www-authenticate = "0.3.0"
-url = "2.1.1"
 hyperx = "0.13"
+chrono = { version = "0.4", features = ["serde"] }
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 reqwest = { version = "0.10", features = ["json"] }
-anyhow = "1.0.25"
+failure = "0.1"
 tokio = { version = "0.2", features = ["full"] }
 serde = "1.0"
 serde_json = "1.0"
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -18,3 +18,7 @@ anyhow = "1.0.25"
 tokio = { version = "0.2", features = ["full"] }
 serde = "1.0"
 serde_json = "1.0"
+async-oauth2 = "0.1.1"
+www-authenticate = "0.3.0"
+url = "2.1.1"
+hyperx = "0.13"
--- a/crates/oci-distribution/src/lib.rs
+++ b/crates/oci-distribution/src/lib.rs
@@ -2,21 +2,37 @@
 extern crate serde;
 
 use chrono::prelude::{DateTime, Utc};
+use failure::format_err;
 use hyperx::header::Header;
+use reqwest::header::{HeaderMap, HeaderValue};
 use serde::Deserialize;
 use std::convert::TryFrom;
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
 use crate::errors::*;
-use crate::reference::Reference;
+pub use crate::reference::Reference;
 
 const OCI_VERSION_KEY: &str = "Docker-Distribution-Api-Version";
 
 pub mod errors;
 pub mod reference;
 
-type OciResult<T> = Result<T, anyhow::Error>;
+type OciResult<T> = Result<T, failure::Error>;
 
+/// The OCI client connects to an OCI registry and fetches OCI images.
+///
+/// An OCI registry is a container registry that adheres to the OCI Distribution
+/// specification. DockerHub is one example, as are ACR and GCR. This client
+/// provides a native Rust implementation for pulling OCI images.
+///
+/// Some OCI registries support completely anonymous access. But most require
+/// at least an Oauth2 handshake. Typlically, you will want to create a new
+/// client, and then run the `auth()` method, which will attempt to get
+/// a read-only bearer token. From there, pulling images can be done with
+/// the `pull_*` functions.
+///
+/// For true anonymous access, you can skip `auth()`. This is not recommended
+/// unless you are sure that the remote registry does not require Oauth2.
 struct Client {
     token: Option<DockerToken>,
 }
@@ -39,7 +55,7 @@ impl Client {
         let res = reqwest::get(&url).await?;
         let disthdr = res.headers().get(OCI_VERSION_KEY);
         let version = disthdr
-            .ok_or_else(|| anyhow::format_err!("no header v2 found"))?
+            .ok_or_else(|| failure::format_err!("no header v2 found"))?
             .to_str()?
             .to_owned();
         Ok(version)
@@ -49,10 +65,10 @@ impl Client {
     ///
     /// This performs authorization and then stores the token internally to be used
     /// on other requests.
-    pub async fn auth(&mut self, host: String, secret: Option<String>) -> OciResult<()> {
+    pub async fn auth(&mut self, image: &Reference, _secret: Option<String>) -> OciResult<()> {
         let cli = reqwest::Client::new();
         // The version request will tell us where to go.
-        let url = format!("https://{}/v2/", host);
+        let url = format!("https://{}/v2/", image.registry());
         let res = cli.get(&url).send().await?;
         let disthdr = res.headers().get(reqwest::header::WWW_AUTHENTICATE);
         if disthdr.is_none() {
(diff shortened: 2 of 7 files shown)


crates/oci-distribution/src/reference.rs
```diff
@@ -0,0 +1,82 @@
...
+        &self.whole[self.slash + 1..self.colon]
+    }
+
+    pub fn tag(&self) -> &str {
+        &self.whole[self.colon + 1..]
+    }
+
+    pub fn to_v2_manifest_url(&self) -> String {
+        format!(
+            "https://{}/v2/{}/manifests/{}",
```
thomastaylor312: Shouldn't this be using the `OCI_VERSION_2` constant? We would need the const to be `pub(crate)` but I think we should probably use it here

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+        let disthdr = res.headers().get(OCI_VERSION_KEY);
+        let version = disthdr
+            .ok_or_else(|| anyhow::format_err!("no header {} found", OCI_VERSION_KEY))?
+            .to_str()?
+            .to_owned();
+        Ok(version)
+    }
+
+    pub async fn pull_manifest(&self, image_name: String) -> OciResult<String> {
+        // We unwrap right now because this try_from literally cannot fail.
```
thomastaylor312: So should this be a FIXME?

crates/oci-distribution/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+    "Taylor Thomas <taylor.thomas@microsoft.com>",
+    "Brian Ketelsen <Brian.Ketelsen@microsoft.com>",
+    "Brian Hardock <Brian.Hardock@microsoft.com>",
+    "Ryan Levick <rylevick@microsoft.com>",
+]
+edition = "2018"
+
+[dependencies]
+reqwest = { version = "0.10", features = ["json"] }
+anyhow = "1.0.25"
```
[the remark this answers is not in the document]
technosophos: All the dependencies were selected by grepping through our cargo lock. Nothing in this file is new or out of sync with our other versions

crates/oci-distribution/src/reference.rs
technosophos: What would be the reasoning for doing string interpolation to insert a constant? That just increases the runtime burden.

radu-matei: As a side note, we might also want to support HTTP-only registries.
(Had to manually change this to HTTP to use a local registry.)

crates/oci-distribution/Cargo.toml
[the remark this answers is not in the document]
thomastaylor312: I wasn't talking so much about the dependency version so much as the fact that we return `failure::Error` from our other crates and `anyhow::Error` here, which can lead to confusion

crates/oci-distribution/src/reference.rs
thomastaylor312: My thinking was more along the DRY route here. But I don't have strong feelings one way or another here.

technosophos: Oh! You can do http-only? The spec made it sound like it had to be https!

crates/oci-distribution/Cargo.toml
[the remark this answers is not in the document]
technosophos: Okay. I thought we had agreed to change to anyhow. I will switch back to failure.

crates/oci-distribution/src/reference.rs
technosophos: DRY would make sense if this was something that could ever change, or if I were removing code paths. But it's not doing either.

The V2 version of the spec cannot switch to different URL formats.

By "drying" here, you'd actually be introducing new code paths. And you get harder-to-read code, longer code, and a small performance delay. So I would argue that DRY would be misapplied in this context.

crates/oci-distribution/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+    "Radu Matei <radu.matei@microsoft.com>",
+    "Taylor Thomas <taylor.thomas@microsoft.com>",
+    "Brian Ketelsen <Brian.Ketelsen@microsoft.com>",
+    "Brian Hardock <Brian.Hardock@microsoft.com>",
+    "Ryan Levick <rylevick@microsoft.com>",
+]
+edition = "2018"
+
+[dependencies]
+reqwest = { version = "0.10", features = ["json"] }
```
rylev: Is there a reason we're not specifying minor versions? I believe the latest is 0.10.4

crates/oci-distribution/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+anyhow = "1.0.25"
+tokio = { version = "0.2", features = ["full"] }
```
rylev: We shouldn't include full features if possible. This significantly hurts compile times.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
+#[macro_use]
+extern crate serde;
```
rylev: personal nit: I prefer referring to macros using fully qualified paths instead of using `#[macro_use]`

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl Client {
+    /// According to the v2 specification, 200 and 401 error codes MUST return the
+    /// version. It appears that any other response code should be deemed non-v2.
+    ///
+    /// For this implementation, it will return v2 or an error result. If the error is a
+    /// `reqwest` error, the request itself failed. All other error messages mean that
+    /// v2 is not supported.
+    pub async fn version(&self, host: String) -> OciResult<String> {
```
rylev: Should `host` be `&str` for all these methods?

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    ///
+    /// This performs authorization and then stores the token internally to be used
+    /// on other requests.
+    pub async fn auth(&mut self, host: String, secret: Option<String>) -> OciResult<()> {
+        let cli = reqwest::Client::new();
+        // The version request will tell us where to go.
+        let url = format!("https://{}/v2/", host);
+        let res = cli.get(&url).send().await?;
+        let disthdr = res.headers().get(reqwest::header::WWW_AUTHENTICATE);
+        if disthdr.is_none() {
```
rylev: Maybe a `match` or an `if let Some(..)` is more appropriate here to avoid the unwrap a few lines below?

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            // The Authenticate header can be set to empty string.
+            return Ok(());
+        }
+        // FIXME: There must be a more elegant way of passing a header from hyper into WwwAuthenticate.
+        let auth = WwwAuthenticate::parse_header(&disthdr.unwrap().to_str()?.to_string().into())?;
```
rylev:
```rust
let auth = WwwAuthenticate::parse_header(&disthdr.unwrap().as_bytes().into())?;
```

`as_bytes` is basically free (unlike `to_str` which checks for ascii), and there's no need for a string as you can get a `Raw` from `bytes`.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        let challenge_opt = auth.get::<BearerChallenge>();
+        if challenge_opt.is_none() {
```
rylev: Same as above. One other thing you could do is:

`let challenge_opt = challenge_opt.ok_or(Ok(())?`

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            return Ok(());
+        }
+
+        let challenge = challenge_opt.as_ref().unwrap()[0].clone();
```
rylev: I believe the clone here isn't necessary.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        let realm = challenge.realm.unwrap();
+        let mut oauth = OauthClient::new(
+            "krustlet",
+            Url::parse(realm.as_str())?,
```
rylev: Why parse twice?

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            Url::parse(realm.as_str())?,
+        );
+        oauth.add_scope(challenge.scope.unwrap().as_str());
+        oauth.set_client_secret(secret.unwrap_or_else(|| "".to_owned()));
```
rylev: I think `unwrap_or_default` is a better choice here

crates/oci-distribution/src/lib.rs
rylev: And why cannot it not fail?

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        let status = res.status();
+        if res.status().is_client_error() {
+            // According to the OCI spec, we should see an error in the message body.
+            let err = res.json::<OciEnvelope>().await?;
+            // FIXME: This should not have to wrap the error.
+            return Err(anyhow::format_err!("{}", err.errors[0]));
+        } else if status.is_server_error() {
+            return Err(anyhow::format_err!("Server error at {}", url));
+        }
+        let text = res.text().await?;
```
rylev: I'd just make this the first if condition instead of doing the early returns before.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub realm: Option<String>,
+    pub service: Option<String>,
+    pub scope: Option<String>,
+}
+
+impl Challenge for BearerChallenge {
+    fn challenge_name() -> &'static str {
+        "Bearer"
+    }
+    fn from_raw(raw: RawChallenge) -> Option<Self> {
```
rylev: nit: spaces between functions

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        match raw {
+            RawChallenge::Token68(_) => None,
+            RawChallenge::Fields(mut map) => Some(BearerChallenge {
+                // NB (mpb): This follows the `remove` pattern from the existing
```
rylev: This is to make ownership of the items that were in the map. Otherwise you'd need to clone.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+                // implementations. Not sure why they do this, though.
+                realm: map.remove("realm"),
+                scope: map.remove("scope"),
+                service: map.remove("service"),
+            }),
+        }
+    }
+    fn into_raw(self) -> RawChallenge {
+        let mut map = ChallengeFields::new();
+        self.realm
```
rylev: perhaps it's more code but I wonder if `if let Some` would be clearer...

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+const OCI_VERSION_KEY: &str = "Docker-Distribution-Api-Version";
+
+pub mod errors;
+pub mod reference;
+
+type OciResult<T> = Result<T, anyhow::Error>;
+
+struct Client {
+    token: Option<StandardToken>,
```
rylev: To prevent users from using an unauthenticated Client, I would make this not optional and have a constructor that does the authencating and populates this field.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        RawChallenge::Fields(map)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[tokio::test]
+    async fn test_version() {
+        let c = Client::default();
```
rylev: Do we really want to make network requests in tests?

crates/oci-distribution/src/reference.rs
```diff
@@ -0,0 +1,82 @@
...
+            self.registry(),
+            self.repository(),
+            self.tag()
+        )
+    }
+}
+
+impl TryFrom<String> for Reference {
+    type Error = ();
```
rylev: Do we want to give this a more descriptive name?

```rust
pub struct ReferenceParseError {
    _priv: ()
}
```

crates/oci-distribution/Cargo.toml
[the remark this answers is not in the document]
thomastaylor312: Created #52 so we actually make the change

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            .expect("result from version request");
+        assert_eq!("registry/2.0".to_owned(), ver);
+    }
+
+    #[tokio::test]
+    async fn test_auth() {
+        let mut c = Client::default();
+        c.auth("webassembly.azurecr.io".to_owned(), None)
+            .await
+            .expect("result from version request");
```
radu-matei: nit: _from `auth` request_
This still seems to fail though:

```
running 1 test
thread 'test::test_auth' panicked at 'called `Option::unwrap()` on a `None` value', /rustc/f3e1a954d2ead4e2fc197c7da7d71e6c61bad196/src/libcore/macros/mod.rs:15:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test test::test_auth ... FAILED

failures:

failures:
    test::test_auth
```

crates/oci-distribution/src/lib.rs
technosophos: It can. I was wrong. Fixed it up.

crates/oci-distribution/Cargo.toml
technosophos: I've been copying what the docs say, figuring they were right. I will start using full versions if that is better.

crates/oci-distribution/Cargo.toml
technosophos: I'm confused on this one... the async_oauth2 said it needed the above. Do I just ignore the docs in that case? (At this point, I removed async_oauth, so I think I can go back to whatever)

crates/oci-distribution/Cargo.toml
technosophos: I'm more confused, now. I went to look at the other Cargo.toml files in this project, and most of them do major, minor for at least some. How do we know when to include the patch version and when not to?

crates/oci-distribution/src/lib.rs
technosophos: This is probably my Go-brain speaking, but to avoid apparent cyclomatic complexity, there are cases where we want to return early without increasing the indent. What is the right way to do that in Rust? Both `match` and `if let Some()` would increase the indent (and thus the apparent cyclomatic complexity). `if let None` seems about the same as `if is_none()`.

crates/oci-distribution/src/lib.rs
technosophos: Ooo... that is much cleaner!

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: fixed

crates/oci-distribution/src/lib.rs
technosophos: fixed

crates/oci-distribution/src/lib.rs
technosophos: Clients are supposed to have the choice of authenticating (because some registries don't support an oauth flow). I made this clearer in the inline docs

crates/oci-distribution/src/lib.rs
technosophos: Again, I would normally say no... but we're doing it elsewhere. Today we talked about setting up some integration tests, in which case I will move these. But for now, I would prefer to leave these so that we have some sanity check on a protocol that actually does not follow the published standard.

crates/oci-distribution/src/reference.rs
technosophos: This is moved over from Wok. I rewrote it today. I don't know the history behind it. But it was far more helpful with some actual errors instead of returning `()` on failure. I found it very confusing (see earlier thing about this not returning an error)

crates/oci-distribution/src/lib.rs
technosophos: Fixed

technosophos:
@rylev raised a good point: We've allowed some of our unit tests to do network calls (which means tests require network to pass, etc. etc). For Docker Registry stuff (which is where most of the offenders are), we should probably write functional tests for this. I'll refactor this one later.

crates/oci-distribution/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+        } else if status.is_server_error() {
+            return Err(format_err!("Server error at {}", url));
+        }
+        let text = res.text().await?;
+        Ok(text)
+    }
+}
+
+#[derive(serde::Deserialize)]
+struct DockerToken {
```
radu-matei: nit: Personally I would not use the `Docker*` nomenclature.

technosophos: OciToken? Who standardizes this token? I thought it was Docker in this case (not OCI)

radu-matei: It is a registry token, so I would use `RegistryToken` for this.

crates/oci-distribution/src/lib.rs
radu-matei: +1 for keeping the tests until we have some integration testing in place.

technosophos:
Okay, so that this point I have the manifest code in-flight. I tried to implement close to the OCI spec, but am finding that the real objects are less than identical to what the spec says. So I have had to make a few minor adjustments.

Next after manifest will be layers. I think we are really close to being able to pull a WASM from a registry.

crates/oci-distribution/Cargo.toml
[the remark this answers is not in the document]
technosophos: Removed anyhow

crates/oci-distribution/src/lib.rs
technosophos: I tried this out, and I agree. It is actually more readable all around.

crates/oci-distribution/src/lib.rs
technosophos: Fixed.

crates/oci-distribution/src/lib.rs
technosophos: Oh! That totally makes sense!

technosophos:
@rylev said this morning that this PR is still good to go. I will follow up with him on his remaining suggestions. But for now, I'll merge this to unblock related work.

PR #51, opened 2020-03-09
feat(kubelet): Fully glues together logs logic

### LINKED ISSUE #33: Fully implement kubelet logs path

thomastaylor312:
As part of this, I changed the provider trait to instead return a vec
of bytes instead of worrying about all the allocations of copying each
line into a string. Instead the raw data is just passed back as part of
the HTTP response body.

I also found that the wasi logs implementation needed to reset the seek
pointer to the beginning of the file for future reads, so that has now
been fixed

NOTE: This is not currently tested with a running master because we do
not have HTTPS set up yet. It has been tested with manual calls to the
endpoint

Closes #33

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -214,7 +214,7 @@ pub trait Provider {
         _namespace: String,
         _pod: String,
         _container: String,
-    ) -> Result<Vec<String>, failure::Error> {
+    ) -> Result<Vec<u8>, failure::Error> {
         Err(NotImplementedError {}.into())
     }
 
--- a/crates/kubelet/src/server.rs
+++ b/crates/kubelet/src/server.rs
@@ -4,7 +4,7 @@
 use hyper::server::conn::AddrStream;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Error, Method, Request, Response, Server, StatusCode};
-use log::{error, info};
+use log::{debug, error, info};
 use tokio::sync::Mutex;
 
 use std::net::SocketAddr;
@@ -68,15 +68,46 @@ fn get_ping() -> Response<Body> {
 /// Implements the kubelet path /containerLogs/{namespace}/{pod}/{container}
 async fn get_container_logs<T: Provider + Sync>(
     provider: &T,
-    _req: &Request<Body>,
+    req: &Request<Body>,
 ) -> Response<Body> {
-    // TODO: extract the right data from the request.
+    // Basic validation steps
+    let path: Vec<&str> = req.uri().path().split('/').collect();
+    // Because of the leading slash, index 0 is an empty string. Index 1 is the
+    // container logs path
+    if path.len() != 5 {
+        let mut res = Response::new(Body::from(format!(
+            "Resource {} not found",
+            req.uri().path()
+        )));
+        *res.status_mut() = StatusCode::NOT_FOUND;
+        return res;
+    }
+
+    let namespace = path[2];
+    let pod = path[3];
+    let container = path[4];
+    if namespace.is_empty() || pod.is_empty() || container.is_empty() {
+        let mut res = Response::new(Body::from(format!(
+            "Resource {} not found",
+            req.uri().path()
+        )));
+        *res.status_mut() = StatusCode::NOT_FOUND;
+        return res;
+    }
+
+    // END validation
+
+    debug!(
+        "Got container log request for container {} in pod {} in namespace {}",
+        container, pod, namespace
+    );
+
     match provider
-        .logs("".to_string(), "".to_string(), "".to_string())
+        .logs(namespace.into(), pod.into(), container.into())
         .await
     {
-        Ok(lines) => Response::new(Body::from(lines.join("\n"))),
-        // TODO: THis should detect not implemented vs. regular error
+        Ok(data) => Response::new(Body::from(data)),
+        // TODO: This should detect not implemented vs. regular error (pod not found, etc.)
         Err(e) => {
             error!("Error fetching logs: {}", e);
             let mut res = Response::new(Body::from("Not Implemented"));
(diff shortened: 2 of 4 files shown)


crates/kubelet/src/server.rs
```diff
@@ -68,15 +68,46 @@ fn get_ping() -> Response<Body> {
...
+    if path.len() != 5 {
+        let mut res = Response::new(Body::from(format!(
+            "Resource {} not found",
+            req.uri().path()
+        )));
+        *res.status_mut() = StatusCode::NOT_FOUND;
+        return res;
+    }
+
+    let namespace = path[2];
```
rylev:
```rust
let [_, _, namespace, pod, container] = path;
```

crates/kubelet/src/server.rs
```diff
@@ -68,15 +68,46 @@ fn get_ping() -> Response<Body> {
...
-    _req: &Request<Body>,
+    req: &Request<Body>,
 ) -> Response<Body> {
-    // TODO: extract the right data from the request.
+    // Basic validation steps
+    let path: Vec<&str> = req.uri().path().split('/').collect();
+    // Because of the leading slash, index 0 is an empty string. Index 1 is the
+    // container logs path
+    if path.len() != 5 {
+        let mut res = Response::new(Body::from(format!(
```
rylev: Should we do this instead:

```
return Response::builder().status(StatusCode::NOT_FOUND).body(...);
```

crates/kubelet/src/kubelet.rs
```diff
@@ -214,7 +214,7 @@ pub trait Provider {
         _namespace: String,
         _pod: String,
         _container: String,
-    ) -> Result<Vec<String>, failure::Error> {
+    ) -> Result<Vec<u8>, failure::Error> {
         Err(NotImplementedError {}.into())
```
rylev: Is there a reason we default to this? It seems we should require that the implementing provider implement this method. If they choose to implement it with a `NotImplementedError` than that's their decision.

crates/wasi-provider/src/lib.rs
```diff
@@ -172,12 +172,11 @@ impl Provider for WasiProvider {
                 pod_name: pod,
                 container_name: container,
             })?;
-        Ok(handle
-            .0
-            .by_ref()
-            .lines()
-            .map(|l| l.unwrap_or_else(|_| "<error while reading line>".to_string()))
-            .collect::<Vec<String>>())
+        let mut output = Vec::default();
```
rylev: `Vec::new` is shorter and the same thing as `Vec::default`. 

It's probably not worth it, but we could read the file metadata and find the length of the file and preallocate that much space in the Vec using `Vec::with_capacity`.

crates/kubelet/src/kubelet.rs
thomastaylor312: Right now it serves as more of a reminder to us. I’ll create an issue to change this before we release

crates/kubelet/src/server.rs
thomastaylor312: Most definitely. I was just copy pasting and didn’t even think of looking

crates/kubelet/src/server.rs
thomastaylor312: Oh so that’s how it’s done 😁. Had a feeling you might know

crates/kubelet/src/kubelet.rs
rylev: Leaving it like this means we need to change two crates in the future instead of making the trait definition how it should be from the start.

thomastaylor312: As it is out of scope in this PR, I'll do it in a follow up

thomastaylor312: I've thought about this some more and since implementing logs and exec is completely optional, I think this should remain as is

rylev: Should it return an empty Vec then? This will cause other work flows to error out.

crates/kubelet/src/server.rs
```diff
@@ -68,15 +68,47 @@ fn get_ping() -> Response<Body> {
...
-    _req: &Request<Body>,
+    req: &Request<Body>,
 ) -> Response<Body> {
-    // TODO: extract the right data from the request.
+    // Basic validation steps
+    let path: Vec<&str> = req.uri().path().split('/').collect();
+    // Because of the leading slash, index 0 is an empty string. Index 1 is the
+    // container logs path
+    let (namespace, pod, container) = match path.as_slice() {
+        [_, _, namespace, pod, container] => (*namespace, *pod, *container),
```
rylev: Maybe add: `!if namespace. is_empty() & ...` to this? You can then get rid of the other check below

crates/wasi-provider/src/lib.rs
```diff
@@ -188,12 +187,13 @@ impl Provider for WasiProvider {
...
-
-        Ok((&mut handle.0)
-            .lines()
-            .map(|l| l.unwrap_or_else(|_| "<error while reading line>".to_string()))
-            .collect::<Vec<String>>()
-            .await)
+        let mut output = Vec::new();
+        (&mut handle.0).read_to_end(&mut output).await?;
+        // Reset the seek location for the next call to read from the file
+        // NOTE: This is a little janky, but the Tokio BufReader does not
```
rylev: I don't think this is very janky... What do we want to say with this comment?

thomastaylor312: I don't like pulling out an inner mutable ref to a read object just to seek, especially when the standard lib equivalent has the ability to do this without the inner mutable ref

crates/kubelet/src/server.rs
thomastaylor312: I think for now I'll leave it as I am knee deep in another task.

crates/wasi-provider/src/lib.rs
rylev: True. Not sure it warrants a comment, but I agree

Issue #36, opened 2020-03-04
Update echo provider, actor & tests


bketelsen:
rebuild the echo provider & actor with the new wascc SDK

PR #53, opened 2020-03-10
chore: update LINUX only echo provider

bketelsen:
This PR updates the wascc echo-provider linux ".so" file, as well as the signed wasm actor for the echo actor.  It does not update the macOS version which will be done by @technosophos in a later PR.


bketelsen:
fixes #36

PR #54, opened 2020-03-10
Update build action to build and test on macOS and Ubuntu

radu-matei:
This PR updates our build action to also build for macOS.

It uses a simple matrix config for the operating system, and introduces a dependency on [this GitHub Action](https://github.com/marketplace/actions/engineerd-configurator) to simplify the download of `just`.

Note that testing on macOS is still failing:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error(HostCallFailure(DecodeError { description: "buffer underflow", stack: [] }))', src/libcore/result.rs:1188:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test test::test_wascc_run ... FAILED

failures:

---- test::test_wascc_run stdout ----
thread 'test::test_wascc_run' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError', src/libcore/result.rs:1188:5


failures:
    test::test_wascc_run

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,13 +4,21 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+        - {os: "ubuntu-latest", url: "https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz", name: "just", pathInArchive: "just" }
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -4,13 +4,21 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
```
radu-matei: This ensures other jobs are not cancelled if one fails ([see documentation](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)).

PR #56, opened 2020-03-10
docs(intro): set an IP address before starting krustlet

bacongobbler:
Krustlet requires a KRUSTLET_NODE_IP to run locally or an error will occur.

### DIFF
--- a/docs/community/developers.md
+++ b/docs/community/developers.md
@@ -26,8 +26,7 @@ If you want to build the Docker image, you'll need [Docker](https://docs.docker.
 We use `just` to build our programs, but you can use `cargo` if you want:
 
 ```console
-$ just
-$ cargo build
+$ just build
 ```
 
 Building a Docker image is easy, too:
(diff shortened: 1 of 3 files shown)


docs/community/developers.md
```diff
@@ -43,22 +42,21 @@ That will take a LOOONG time the first build, but the layer cache will make it m
 To run Krustlet locally, you can run
 
 ```console
-$ just run-wasi
+$ just run
```
radu-matei: This runs the wasCC provider by default.
Is this intented?

bacongobbler: I figured that whatever's the default for `just run` is probably the right choice for devs. SHould we instead re-write this section to tell them that they should use `just run-wasi` or `just run-wascc` depending on their needs?

bacongobbler: Note: in the tutorial we actually call this out and inform the user that `run-wasi` is the right choice. Happy to add more clarification here as well. Should be a quick copy n' paste job.

Issue #25, opened 2020-03-03
Implement init containers for all runtimes


thomastaylor312:
Each pod's init containers should be run in order and to completion before starting the others. We likely want to finish discussion on #21 before we do this as this would be a blocking operation as we wait for each container

itowlson:
At the moment this looks like it would be a provider-by-provider thing: all the logic for running pods is implemented within the provider.  As far as I can see, if we want to handle this generically at the kubelet level, we'd need to make providers responsible only for running _containers_, not pods.  Am I missing something?  (I expect I am!)

itowlson:
For avoidance of duplication, I'm looking at this - if someone else is already doing this then please sing out!

thomastaylor312:
Yeah, it is a per provider thing. The provider trait tries to make no assumptions about what is running the underlying workload

PR #58, opened 2020-03-10
feat(*): Adds support for multiple containers

### LINKED ISSUE #24: Implement multiple containers in pods for all runtimes
Right now we only pull the first container from the containers array. We need to pull all containers and start a process for each of them

thomastaylor312:
Both WasCC and WASI will now spin up a running wasmtime instance/actor
for each container in a pod. Right now the actual module is still
hard coded, but will be replaced when the work around OCI is done.

Closes #24

### DIFF
--- a/crates/wasi-provider/src/lib.rs
+++ b/crates/wasi-provider/src/lib.rs
@@ -59,10 +59,6 @@ impl Provider for WasiProvider {
         // and then execute the WASM, passing in the relevant data.
         // When the pod finishes, we update the status to Succeeded unless it
         // produces an error, in which case we mark it Failed.
-        debug!(
-            "Pod added {:?}",
-            pod.metadata.as_ref().and_then(|m| m.name.as_ref())
-        );
         let namespace = pod
             .metadata
             .as_ref()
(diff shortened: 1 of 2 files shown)


crates/wasi-provider/src/lib.rs
```diff
@@ -59,21 +59,13 @@ impl Provider for WasiProvider {
...
         let namespace = pod
             .metadata
             .as_ref()
             .and_then(|m| m.namespace.as_deref())
             .unwrap_or_else(|| "default");
 
         // TODO: Implement this for real.
-        // Okay, so here is where things are REALLY unfinished. Right now, we are
-        // only running the first container in a pod. And we are not using the
-        // init containers at all. And they are not executed on their own threads.
```
bacongobbler: From what I can read, the comment about init containers comment still rings true. We still don't fire off any init containers. Should we keep that part of the comment in here (and in wascc-provider) so we know we still need to implement this?

thomastaylor312: This is captured in #25 so I thought it wasn't necessary here any more

bacongobbler: Great! As long as we got that captured then I'm happy. :)

PR #59, opened 2020-03-11
add e2e tests

bacongobbler:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,3 +32,23 @@ jobs:
         run: |
           just build
           just test
+  e2e:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: engineerd/setup-kind@v0.3.0
+    - uses: engineerd/configurator@v0.0.1
+      with:
+        name: just
+        url: https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz
+        pathInArchive: just
+    - name: Run e2e tests
+      run: |
+        cd crates/wascc-provider
+        cargo build --bin krustlet-wascc --manifest-path ../../Cargo.toml
+        ../../target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 &
+        cd ../wasi-provider
+        cargo build --bin krustlet-wasi --manifest-path ../../Cargo.toml
+        ../../target/debug/krustlet-wasi --node-name krustlet-wasi --port 3001 &
+        cd ../..
+        just test-e2e
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,11 @@ kubelet = { path = "./crates/kubelet", version = "0.1.0", features = ["cli"] }
 wascc-provider = { path = "./crates/wascc-provider", version = "0.1.0" }
 wasi-provider = { path = "./crates/wasi-provider", version = "0.1.0" }
 
+[dev-dependencies]
+serde_derive = "1.0"
+serde_json = "1.0"
+k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_17"] }
+
 [workspace]
 members = [
     "crates/*"
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -71,12 +71,15 @@ impl Config {
             default_node_ip(&mut hostname.clone(), &addr)
                 .expect("unable to get default node IP address")
         });
+        let node_name = opts
+            .node_name
+            .unwrap_or_else(|| sanitize_hostname(&hostname));
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
         let pfx_password = opts.pfx_password.unwrap_or_default();
         Config {
             node_ip,
-            node_name: sanitize_hostname(&hostname),
+            node_name,
             hostname,
             server_config: ServerConfig {
                 addr,
--- a/docs/community/developers.md
+++ b/docs/community/developers.md
@@ -71,6 +71,36 @@ To set the node IP, run:
 $ export KRUSTLET_NODE_IP=<the ip address>
 ```
 
+## Testing
+
+Krustlet contains both integration and unit tests. For convenience, there are `just` targets for running one or the other.
+
+For unit tests:
+
+```console
+$ just test
+```
+
+For the integration tests, start a wascc and wasi node in separate terminals before running the tests.
+
+In terminal 1:
+
+```console
+$ just run-wascc
+```
+
+In terminal 2:
+
+```console
+$ just run-wasi
+```
+
+And in terminal 3:
+
+```
+$ just test-e2e
+```
+
 ## Creating your own Kubelets with Krustlet
 
 If you want to create your own Kubelet based on Krustlet, all you need to do is implement a `Provider`.
--- a/justfile
+++ b/justfile
@@ -12,16 +12,19 @@ prefetch:
 test:
     cargo fmt --all -- --check
     cargo clippy --workspace
-    cargo test --workspace
+    cargo test --workspace --lib
+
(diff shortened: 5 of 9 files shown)


justfile
```diff
@@ -11,7 +11,10 @@ prefetch:
 test:
     cargo fmt --all -- --check
     cargo clippy --workspace
-    cargo test --workspace
+    cargo test --workspace --lib
```
rylev: I believe in the README we mention that using `cargo` instead of `just` is also fine. Perhaps we should note that `cargo test` will run integration tests so it's advisable to run with the `--lib` option when testing locally (unless of course the dev wants to run integration tests too)

justfile
```diff
@@ -11,7 +11,10 @@ prefetch:
...
+
+test-e2e:
```
rylev: I personally think calling this `integration_test` might be more clear

bacongobbler:
unmarking as a WIP. This is now ready for review!

Cargo.toml
```diff
@@ -21,6 +21,9 @@ env_logger = "0.7.1"
 kubelet = { path = "./crates/kubelet", version = "0.1.0", features = ["cli"] }
 wascc-provider = { path = "./crates/wascc-provider", version = "0.1.0" }
 wasi-provider = { path = "./crates/wasi-provider", version = "0.1.0" }
+serde_json = "1.0"
```
thomastaylor312: These should be added under `dev-dependencies`. Also, I think that it is better to add the `features = ["derive"]` to the `serde` dependency instead of importing `serde_derive` directly

docs/community/developers.md
```diff
@@ -71,6 +71,24 @@ To set the node IP, run:
...
+For unit tests:
+
+```console
+$ just test
+```
+
+For the integration tests, start a wascc and wasi node in separate terminals before running the tests:
+
+```console
+$ just run-wascc
```
thomastaylor312: Nit: I would add comments like `# run in terminal 1` to each of these just to be really clear

tests/integration_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+        "apiVersion": "v1",
+        "kind": "Pod",
+        "metadata": {
+            "name": "helloworld"
+        },
+        "spec": {
+            "containers": [
+                {
+                    "name": "helloworld",
+                    "image": "bacongobbler.azurecr.io/helloworld:v1.0.0",
```
thomastaylor312: I know that right now it isn't doing anything, but should this be the hello world we have in webassembly ACR?

tests/integration_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+    let node_status = node.status.expect("node reported no status");
+    assert_eq!(
+        node_status
+            .node_info
+            .expect("node reported no information")
+            .architecture,
+        "wasm-wasi",
+        "expected node to support the wasm-wasi architecture"
+    );
+
```
thomastaylor312: There a reason we aren't creating a pod here?

tests/integration_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    });
+    let pod_data = serde_json::to_vec(&p).expect("pod should always be serializable to JSON");
+
+    let pod = pods
+        .create(&PostParams::default(), pod_data)
+        .await
+        .expect("could not create pod");
+
+    assert_eq!(pod.status.unwrap().phase.unwrap(), "Pending");
```
thomastaylor312: We definitely should get past Pending here. It should probably be Running (and updated to `Completed` once we get status working)

tests/integration_tests.rs
bacongobbler: sure

tests/integration_tests.rs
bacongobbler: no particular reason

tests/integration_tests.rs
bacongobbler: This is just an initial submission. Happy to go further here but I figured I should stop somewhere before I go down the whole rabbit hole. A few tests were failing previously so I wanted to incrementally add more tests in follow-up PRs.

.github/workflows/build.yml
```diff
@@ -22,3 +22,23 @@ jobs:
...
+    - uses: engineerd/configurator@v0.0.1
+      with:
+        name: just
+        url: https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz
+        pathInArchive: just
+    - name: Run e2e tests
+      run: |
+        cd crates/wascc-provider
+        cargo build --bin krustlet-wascc --manifest-path ../../Cargo.toml
+        ../../target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 &
```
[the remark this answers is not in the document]
bacongobbler: yes, the docker container actions run in a container, so these processes are killed when the task is finished.

tests/integration_tests.rs
bacongobbler: Added a comment as a TODO item for follow-up

Cargo.toml
bacongobbler: > the package `krustlet` depends on `serde_json`, with features: `derive` but `serde_json` does not have these features.

Reverting back to previous code. Note: this is the same as what's in `crates/kubelet/Cargo.toml`. Let's revisit later

crates/kubelet/src/config.rs
```diff
@@ -71,12 +71,16 @@ impl Config {
             default_node_ip(&mut hostname.clone(), &addr)
                 .expect("unable to get default node IP address")
         });
+        let node_name = match opts.node_name {
```
rylev: Is it clearer to write? 
```rust
let node_name = opts.node_name.unwrap_or_else(|| sanitize_hostname(&hostname));
```
FWIW I'm fine with the match expression

crates/kubelet/src/config.rs
```diff
@@ -71,12 +71,16 @@ impl Config {
...
+            Some(name) => name,
+            None => sanitize_hostname(&hostname),
+        };
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
         let pfx_password = opts.pfx_password.unwrap_or_default();
         Config {
             node_ip,
-            node_name: sanitize_hostname(&hostname),
+            node_name: node_name,
```
rylev: This can just be `node_name,` instead of `node_name: node_name`

tests/integration_tests.rs
```diff
@@ -0,0 +1,158 @@
...
+    assert_eq!(
+        node_meta
+            .labels
+            .expect("node had no labels")
+            .get("kubernetes.io/arch")
+            .expect("node did not have kubernetes.io/arch label"),
+        "wasm32-wascc"
+    );
+
+    let pods: Api<Pod> = Api::namespaced(client.clone(), "default");
```
rylev: This call to `clone` isn't necessary

Cargo.toml
thomastaylor312: Yeah, I meant that you'd need to bring in `serde` and `serde_json`

crates/kubelet/src/config.rs
bacongobbler: agreed

tests/integration_tests.rs
bacongobbler: Ah! Thank you for catching this. It was necessary earlier but after a few refactors it's no longer necessary.

bacongobbler:
Holding off on merging; I think I need to rebase this now #77 was merged (the integration tests used 0.28 of kube-rs as well

bacongobbler:
:cry: 

> thread 'main' panicked at 'failed to deserialize node from node definition JSON: Error("missing field `bootID`", line: 0, column: 0)', src/libcore/result.rs:1188:5

I'm fairly certain this was due to the introduction of #77, sadly. Now we need to populate the API with ALL the fields, rather than just the mandatory ones. Sigh.

Will look at this later today

PR #62, opened 2020-03-12
feat(wasi): Adds runtime handle wrapper

thomastaylor312:
Introduces the new `RuntimeHandler` type for the WASI runtime. The purpose
of this handle is to wrap all of the logic for managing the process and
getting its logs/status/etc. after it has started. Ideally, this could be
something we generalize across all providers in the future after we hammer
the kinks out. There are still some clippy warnings due to some of the code
not yet being used

Note: This also introduces some basic messaging for each "container" to
inform the provider that its status has changed

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -7,8 +7,14 @@ use crate::{
     server::start_webserver,
 };
 use async_trait::async_trait;
+use chrono::{DateTime, Utc};
 use futures::{StreamExt, TryStreamExt};
-use k8s_openapi::api::core::v1::{ConfigMap, Container, EnvVar, Secret};
+use k8s_openapi::api::core::v1::ContainerStatus as KubeContainerStatus;
+use k8s_openapi::api::core::v1::{
+    ConfigMap, Container, ContainerState, ContainerStateRunning, ContainerStateTerminated,
+    ContainerStateWaiting, EnvVar, Secret,
+};
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::{
     api::{Api, ListParams, WatchEvent},
     client::APIClient,
@@ -49,6 +55,97 @@ pub enum Phase {
 pub struct Status {
     pub phase: Phase,
     pub message: Option<String>,
+    pub container_statuses: Vec<ContainerStatus>,
+}
+
+/// A struct that contains all necessary information for a waiting status, meant
+/// to be used with the [ContainerStatus] enum
+#[derive(Clone, Debug)]
+pub struct WaitingStatus {
+    /// The timestamp of when this status was reported
+    pub timestamp: DateTime<Utc>,
+    /// A human readable string describing the why it is in a waiting status
+    pub message: String,
+}
+
+/// A struct that contains all necessary information for a running status, meant
+/// to be used with the [ContainerStatus] enum
+#[derive(Clone, Debug)]
+pub struct RunningStatus {
+    /// The timestamp of when this status was reported
+    pub timestamp: DateTime<Utc>,
+}
+
+/// A struct that contains all necessary information for a terminated status,
+/// meant to be used with the [ContainerStatus] enum
+#[derive(Clone, Debug)]
+pub struct TerminatedStatus {
+    /// The timestamp of when this status was reported
+    pub timestamp: DateTime<Utc>,
+    /// A human readable string describing the why it is in a terminating status
+    pub message: String,
+    /// Should be set to true if the process exited with an error
+    pub failed: bool,
+}
+
+/// ContainerStatus is a simplified version of the Kubernetes container status
+/// for use in providers. It allows for simple creation of the current status of
+/// a "container" (a running wasm process) without worrying about a bunch of
+/// Options. Use the [ContainerStatus::to_kubernetes] method for converting it
+/// to a Kubernetes API status
+#[derive(Clone, Debug)]
+pub enum ContainerStatus {
+    Waiting(WaitingStatus),
+    Running(RunningStatus),
+    Terminated(TerminatedStatus),
+}
+
+impl ContainerStatus {
+    pub fn to_kubernetes(&self, pod_name: &str) -> KubeContainerStatus {
+        let state = match self {
+            Self::Waiting(w) => ContainerState {
+                waiting: Some(ContainerStateWaiting {
+                    message: Some(w.message.clone()),
+                    ..Default::default()
+                }),
+                ..Default::default()
+            },
+            Self::Running(r) => ContainerState {
+                running: Some(ContainerStateRunning {
+                    started_at: Some(Time(r.timestamp)),
+                }),
+                ..Default::default()
+            },
+            Self::Terminated(t) => {
+                let exit_code = if t.failed { 1 } else { 0 };
+                ContainerState {
+                    terminated: Some(ContainerStateTerminated {
+                        finished_at: Some(Time(t.timestamp)),
+                        message: Some(t.message.clone()),
+                        exit_code,
+                        ..Default::default()
+                    }),
+                    ..Default::default()
+                }
+            }
+        };
+        let ready = state.running.is_some();
+        KubeContainerStatus {
+            state: Some(state),
+            name: pod_name.to_string(),
+            // Right now we don't have a way to probe, so just set to ready if
+            // in a running state
+            ready,
+            // This is always true if startupProbe is not defined. When we
+            // handle probes, this should be updated accordingly
+            started: Some(true),
+            // The rest of the items in status (see docs here:
+            // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#containerstatus-v1-core)
+            // either don't matter for us or we have not implemented the
+            // functionality yet
+            ..Default::default()
+        }
+    }
 }
 
 /// Kubelet provides the core Kubelet capability.
@@ -499,6 +596,7 @@ mod test {
             Ok(Status {
                 phase: Phase::Succeeded,
                 message: None,
+                container_statuses: Vec::new(),
             })
         }
     }
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -58,36 +58,6 @@ pub struct Status {
     pub container_statuses: Vec<ContainerStatus>,
 }
 
-/// A struct that contains all necessary information for a waiting status, meant
-/// to be used with the [ContainerStatus] enum
-#[derive(Clone, Debug)]
-pub struct WaitingStatus {
-    /// The timestamp of when this status was reported
-    pub timestamp: DateTime<Utc>,
-    /// A human readable string describing the why it is in a waiting status
-    pub message: String,
-}
-
-/// A struct that contains all necessary information for a running status, meant
-/// to be used with the [ContainerStatus] enum
-#[derive(Clone, Debug)]
-pub struct RunningStatus {
-    /// The timestamp of when this status was reported
-    pub timestamp: DateTime<Utc>,
-}
-
-/// A struct that contains all necessary information for a terminated status,
-/// meant to be used with the [ContainerStatus] enum
-#[derive(Clone, Debug)]
-pub struct TerminatedStatus {
-    /// The timestamp of when this status was reported
-    pub timestamp: DateTime<Utc>,
-    /// A human readable string describing the why it is in a terminating status
-    pub message: String,
-    /// Should be set to true if the process exited with an error
-    pub failed: bool,
-}
-
 /// ContainerStatus is a simplified version of the Kubernetes container status
 /// for use in providers. It allows for simple creation of the current status of
 /// a "container" (a running wasm process) without worrying about a bunch of
@@ -95,44 +65,58 @@ pub struct TerminatedStatus {
 /// to a Kubernetes API status
 #[derive(Clone, Debug)]
 pub enum ContainerStatus {
-    Waiting(WaitingStatus),
-    Running(RunningStatus),
-    Terminated(TerminatedStatus),
+    Waiting {
+        /// The timestamp of when this status was reported
+        timestamp: DateTime<Utc>,
+        /// A human readable string describing the why it is in a waiting status
+        message: String,
+    },
+    Running {
+        /// The timestamp of when this status was reported
+        timestamp: DateTime<Utc>,
+    },
+    Terminated {
+        /// The timestamp of when this status was reported
+        timestamp: DateTime<Utc>,
+        /// A human readable string describing the why it is in a terminating status
+        message: String,
+        /// Should be set to true if the process exited with an error
+        failed: bool,
+    },
 }
 
 impl ContainerStatus {
-    pub fn to_kubernetes(&self, pod_name: &str) -> KubeContainerStatus {
-        let state = match self {
-            Self::Waiting(w) => ContainerState {
-                waiting: Some(ContainerStateWaiting {
-                    message: Some(w.message.clone()),
+    pub fn to_kubernetes(&self, pod_name: String) -> KubeContainerStatus {
+        let mut state = ContainerState::default();
+        match self {
+            Self::Waiting { message, .. } => {
+                state.waiting.replace(ContainerStateWaiting {
+                    message: Some(message.clone()),
                     ..Default::default()
-                }),
-                ..Default::default()
-            },
-            Self::Running(r) => ContainerState {
-                running: Some(ContainerStateRunning {
-                    started_at: Some(Time(r.timestamp)),
-                }),
-                ..Default::default()
-            },
-            Self::Terminated(t) => {
-                let exit_code = if t.failed { 1 } else { 0 };
-                ContainerState {
-                    terminated: Some(ContainerStateTerminated {
-                        finished_at: Some(Time(t.timestamp)),
-                        message: Some(t.message.clone()),
-                        exit_code,
-                        ..Default::default()
-                    }),
+                });
+            }
+            Self::Running { timestamp } => {
+                state.running.replace(ContainerStateRunning {
+                    started_at: Some(Time(*timestamp)),
+                });
+            }
+            Self::Terminated {
+                timestamp,
+                message,
+                failed,
+            } => {
+                state.terminated.replace(ContainerStateTerminated {
+                    finished_at: Some(Time(*timestamp)),
+                    message: Some(message.clone()),
+                    exit_code: *failed as i32,
                     ..Default::default()
-                }
+                });
             }
         };
         let ready = state.running.is_some();
         KubeContainerStatus {
             state: Some(state),
-            name: pod_name.to_string(),
+            name: pod_name,
             // Right now we don't have a way to probe, so just set to ready if
             // in a running state
             ready,
--- a/crates/wasi-provider/src/handle.rs
+++ b/crates/wasi-provider/src/handle.rs
@@ -1,0 +1,70 @@
+use std::io::SeekFrom;
+
+use tokio::io::{AsyncReadExt, AsyncSeekExt, BufReader};
+use tokio::sync::watch::Receiver;
+use tokio::task::JoinHandle;
+
+use kubelet::ContainerStatus;
+
+/// Represents a handle to a running WASI instance. Right now, this is
+/// experimental and just for use with the [crate::WasiProvider]. If we like
+/// this pattern, we will expose it as part of the kubelet crate
+pub struct RuntimeHandle<R: AsyncReadExt + AsyncSeekExt + Unpin> {
+    output: BufReader<R>,
+    handle: JoinHandle<Result<(), failure::Error>>,
+    status_channel: Receiver<ContainerStatus>,
+}
+
+impl<R: AsyncReadExt + AsyncSeekExt + Unpin> RuntimeHandle<R> {
+    /// Create a new handle with the given reader for log output and a handle to
+    /// the running tokio task. The sender part of the channel should be given
+    /// to the running process and the receiver half passed to this constructor
+    /// to be used for reporting current status
+    pub fn new(
+        output: R,
+        handle: JoinHandle<Result<(), failure::Error>>,
+        status_channel: Receiver<ContainerStatus>,
+    ) -> Self {
+        RuntimeHandle {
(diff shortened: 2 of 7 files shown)


crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -1,9 +1,17 @@
 use std::collections::HashMap;
-use std::fs::File;
+use std::io::SeekFrom;
 use std::path::Path;
 
 use failure::{bail, format_err};
-use log::info;
+use k8s_openapi::api::core::v1::{
```
thomastaylor312: I'm not sure if we should be bringing the k8s types into providers, but it felt like it would work here. The other option I thought of was adding a simplified version of the status types into `kubelet` with a `From` implementation to turn them into the k8s types

thomastaylor312:
@rylev Your review would be much appreciated here as I did a bunch of stuff in Rust I hadn't done before and probably did a lot of things inefficiently

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+            handle,
+            status_recv,
+        ))
+    }
+}
+
+/// Represents a handle to a running WASI instance. Right now, this is
+/// experimental and just for use with the [crate::WasiProvider]. If we like
+/// this pattern, we will expose it as part of the kubelet crate
+pub struct RuntimeHandle<R: AsyncReadExt + AsyncSeekExt + Unpin> {
```
rylev: Can we put this in its own file?

crates/wasi-provider/src/wasi_runtime.rs
rylev: This seems trivial enough that adding this dependency to the crate doesn't seem worth it. Duplicating that definition seems like a better idea

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -34,27 +45,33 @@ impl WasiRuntime {
...
+        // temp file getting cleaned out from underneath us while running. If we
+        // think it necessary, we can make these permanent files with a cleanup
+        // loop that runs elsewhere. These will get deleted when the reference
+        // is dropped
         Ok(WasiRuntime {
             module_data,
             env,
             args,
             dirs,
+            output: NamedTempFile::new_in(log_dir)?,
```
rylev: If `run` is async then `new` should probably be too. `NamedTempFile::new_in` blocks I believe

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
-                match export {
-                    Some(export) => Ok(export.clone().into()),
-                    None => bail!(
-                        "import `{}` was not found in module `{}`",
-                        i.name(),
-                        i.module()
-                    ),
+        // Clone the module data so it can be moved and we don't have to worry
+        // about the lifetime of the struct data
+        let module_data = self.module_data.clone();
```
rylev: Does cloning here make more sense then wrapping the `module_data` in an `Arc`?

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+                                finished_at: Some(Time(chrono::Utc::now())),
+                                ..Default::default()
+                            }),
+                            ..Default::default()
+                        })
+                        .expect("status should be able to send");
+                    // Converting from anyhow
+                    Err(format_err!("{}: {}", message, e))
+                }
+            }?;
```
rylev: I think using an early return in the `Err` case is clearer than returning back a result only to then return with `?`

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+            // Iterate through the module includes and resolve imports
+            let imports = match module
```
rylev: I think it would be easier to read if you separated the iteration of the imports from the match expression

crates/wasi-provider/src/wasi_runtime.rs
rylev: After seeing it in use, I definitely think we should not be using the k8 types here.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+                .broadcast(ContainerState {
+                    running: Some(ContainerStateRunning {
+                        started_at: Some(Time(chrono::Utc::now())),
+                    }),
+                    ..Default::default()
+                })
+                .expect("status should be able to send");
+            match instance
+                .get_export("_start")
+                .expect("export")
```
rylev: Should this `expect` message be a bit more descriptive?

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+                        ..Default::default()
+                    }),
+                    ..Default::default()
+                })
+                .expect("status should be able to send");
+            Ok(())
+        });
+
+        Ok(RuntimeHandle::new(
+            tokio::fs::File::from_std(self.output.reopen()?),
```
rylev: I'm pretty sure that reopen blocks.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+/// Represents a handle to a running WASI instance. Right now, this is
+/// experimental and just for use with the [crate::WasiProvider]. If we like
+/// this pattern, we will expose it as part of the kubelet crate
+pub struct RuntimeHandle<R: AsyncReadExt + AsyncSeekExt + Unpin> {
+    output: BufReader<R>,
+    handle: JoinHandle<Result<(), failure::Error>>,
+    status_channel: Receiver<ContainerState>,
+}
+
+impl<R: AsyncReadExt + AsyncSeekExt + Unpin> RuntimeHandle<R> {
```
rylev: I believe we can just require `AsyncRead` and `AsyncSeek` since there are blanket impls for the extension traits.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+        status_channel: Receiver<ContainerState>,
+    ) -> Self {
+        RuntimeHandle {
+            output: BufReader::new(output),
+            handle,
+            status_channel,
+        }
+    }
+
+    pub async fn output(&mut self) -> Result<Vec<u8>, failure::Error> {
```
rylev: Thinking out loud. I wonder if it will make sense for this method to take the buffer it will write into instead of creating a new one.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+    pub async fn stop(&mut self) -> Result<(), failure::Error> {
+        // TODO: Send an actual stop signal once there is support in wasmtime
+        self.wait().await?;
+        unimplemented!("There is currently no way to stop a running wasmtime instance")
+    }
+
+    pub async fn status(&self) -> Result<ContainerState, failure::Error> {
+        // NOTE: For those who modify this in the future, borrow must be as
+        // short lived as possible. We do not use the recv method because it
+        // uses the value each time and blocks on the next call, whereas we want
```
rylev: Isn't that it blocks on the `send` side? This comment makes it seem like borrow would block which I don't believe is true.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -71,64 +88,246 @@ impl WasiRuntime {
...
+        // to return the last sent value until updated
+        Ok((*self.status_channel.borrow()).clone())
```
rylev: I believe this can just be:

```rust
Ok(self.status_channel.borrow().clone())
```

crates/wasi-provider/src/wasi_runtime.rs
technosophos: Normally I don't weigh in on style issues, but I agree with Ryan here: I found this difficult to read. There were 19 lines between the `match` and the beginning of the match block.

crates/wasi-provider/src/wasi_runtime.rs
technosophos: I was gonna disagree with Ryan on this one, on the grounds that there's not case in which this isn't Kubernetes centric. But after reading the PR, I agree: It is probably better to just do a more concise status message inside a provider, and have an the Kubelet layer translate that into the clunky Kubernetes/OpenAPI format. Seems like a nice clean `enum` could do the trick.

That said, I'm definitely not set on this. If it turns out that importing Kube stuff makes the code less repetitive...

technosophos:
In general, I'm a big fan of the new pattern. And I think it should generalize across waSCC.

crates/wasi-provider/src/lib.rs
```diff
@@ -164,6 +141,42 @@ impl Provider for WasiProvider {
         //     let handles = self.handles.read().await;
         //     let containers = handles.get(key_from_pod(&pod));
         // };
+        let pod_name = pod
```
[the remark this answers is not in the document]
thomastaylor312: Yay for kubernetes APIs!

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: This is all done

crates/wasi-provider/src/lib.rs
```diff
@@ -92,37 +88,18 @@ impl Provider for WasiProvider {
...
-                        // temp file getting cleaned out from underneath us while running. If we
-                        // think it necessary, we can make these permanent files with a cleanup
-                        // loop that runs elsewhere. These will get deleted when the reference
-                        // is dropped
-                        // TODO: Actual log path configuration
-                        let tempfile = NamedTempFile::new_in(std::env::current_dir()?)?;
-                        Ok((runtime, tempfile.reopen()?, tempfile.reopen()?))
-                    })
-                    .await??;
+                let runtime = WasiRuntime::new(
```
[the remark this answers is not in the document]
thomastaylor312: Yes, due to the tempfile handling. Is there a way to do this more nicely?

crates/wasi-provider/src/lib.rs
```diff
@@ -92,37 +88,18 @@ impl Provider for WasiProvider {
...
                 debug!("Starting container {} on thread", container.name);
-                let handle = tokio::task::spawn_blocking(move || runtime.run(output_handle1));
-                entry.insert(
-                    container.name.clone(),
-                    (
-                        BufReader::new(tokio::fs::File::from_std(output_handle2)),
-                        handle,
-                    ),
-                );
+                let handle = runtime.run().await?;
```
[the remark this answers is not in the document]
thomastaylor312: That is fair. I'm going to rename the method to start, since that is actually what it is doing

crates/wasi-provider/src/lib.rs
```diff
@@ -92,37 +88,18 @@ impl Provider for WasiProvider {
...
+                entry.insert(container.name.clone(), handle);
```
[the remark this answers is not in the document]
thomastaylor312: Yes, because `container` is a shared reference

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: In order to change this, we'd need to glue together our own temp file handling. There another way around this?

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: See other comments related to blocking

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Without using the `Ext` traits, I couldn't access the `read_to_end` or `seek` methods

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Oooh yes, much better idea

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: `recv` "blocks" in the await sense. According to the docs, borrowing for a long time _can_ create a block though. I'll reword

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Done

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: I think I did it right. It required an `as_ref` while passing it to the function to satisfy the trait bound, but I could have done it wrong.

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Done

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Updated it with something clearer

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Yep, this was a nicer way of doing it

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Completely right. IT has been separated out

crates/kubelet/src/kubelet.rs
```diff
@@ -49,6 +55,97 @@ pub enum Phase {
...
+                ..Default::default()
+            },
+            Self::Running(r) => ContainerState {
+                running: Some(ContainerStateRunning {
+                    started_at: Some(Time(r.timestamp)),
+                }),
+                ..Default::default()
+            },
+            Self::Terminated(t) => {
+                let exit_code = if t.failed { 1 } else { 0 };
```
rylev: `let exit_code = t.failed as i32`

crates/kubelet/src/kubelet.rs
```diff
@@ -49,6 +55,97 @@ pub enum Phase {
...
+pub enum ContainerStatus {
+    Waiting(WaitingStatus),
+    Running(RunningStatus),
+    Terminated(TerminatedStatus),
+}
+
+impl ContainerStatus {
+    pub fn to_kubernetes(&self, pod_name: &str) -> KubeContainerStatus {
+        let state = match self {
+            Self::Waiting(w) => ContainerState {
```
rylev: I think it would be more readable to initialize a mutable `ContainerState` with all `Default::default` and then in each branch set the appropriate item.

crates/kubelet/src/kubelet.rs
```diff
@@ -49,6 +55,97 @@ pub enum Phase {
...
+/// to a Kubernetes API status
+#[derive(Clone, Debug)]
+pub enum ContainerStatus {
+    Waiting(WaitingStatus),
+    Running(RunningStatus),
+    Terminated(TerminatedStatus),
+}
+
+impl ContainerStatus {
+    pub fn to_kubernetes(&self, pod_name: &str) -> KubeContainerStatus {
```
rylev: If we need this to be a `String` anyway, we should just require a `String`. This prevents `String => &str => String` conversions.

crates/wasi-provider/src/handle.rs
```diff
@@ -0,0 +1,70 @@
...
+            status_channel,
+        }
+    }
+
+    /// Write all of the output from the running process into the given buffer.
+    /// Returns the number of bytes written to the buffer
+    pub async fn output(&mut self, buf: &mut Vec<u8>) -> Result<usize, failure::Error> {
+        let bytes_written = self.output.read_to_end(buf).await?;
+        // Reset the seek location for the next call to read from the file
+        // NOTE: This is a little janky, but the Tokio BufReader does not
```
rylev: I still don't think this is a helpful comment. If you feel like we should keep it, can you replace "This is a little janky, but" with "We must call `get_mut()` because"

crates/wasi-provider/src/handle.rs
```diff
@@ -0,0 +1,70 @@
...
+        // the recv method because it uses the value each time and then waits
+        // for a new value on the next call, whereas we want to return the last
+        // sent value until updated
+        Ok(self.status_channel.borrow().clone())
+    }
+
+    // For now this is private (for use in testing and in stop). If we find a
+    // need to expose it, we can do that later
+    pub(crate) async fn wait(&mut self) -> Result<(), failure::Error> {
+        (&mut self.handle).await.unwrap()
```
rylev: Is the additional `&mut` here necessary?

crates/wasi-provider/src/lib.rs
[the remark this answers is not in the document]
rylev: I would recommend keeping the `spawn_blocking` call. This informs the tokio runtime that the call will block so it can put it on its own dedicated worker thread which won't block the rest of the runtime.

crates/wasi-provider/src/wasi_runtime.rs
rylev: Using `spawn_blocking`

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -34,27 +43,33 @@ impl WasiRuntime {
...
+            output: NamedTempFile::new_in(log_dir)?,
         })
     }
 
-    pub fn run(&self, output: File) -> Result<(), failure::Error> {
-        let engine = wasmtime::Engine::default();
-        let store = wasmtime::Store::new(&engine);
-
+    pub async fn start(&self) -> Result<RuntimeHandle<tokio::fs::File>, failure::Error> {
+        let output = self.output.reopen()?;
```
rylev: Do these calls to `reopen` also block? We might also need to wrap these in `spawn_blocking`

crates/kubelet/src/kubelet.rs
```diff
@@ -49,6 +55,97 @@ pub enum Phase {
...
+}
+
+/// ContainerStatus is a simplified version of the Kubernetes container status
+/// for use in providers. It allows for simple creation of the current status of
+/// a "container" (a running wasm process) without worrying about a bunch of
+/// Options. Use the [ContainerStatus::to_kubernetes] method for converting it
+/// to a Kubernetes API status
+#[derive(Clone, Debug)]
+pub enum ContainerStatus {
+    Waiting(WaitingStatus),
```
rylev: Do the associated data of each enum variant need to be separate structs? Would it be simpler to do:

```rust
pub enum ContainerStatus {
    Waiting {
        /// The timestamp of when this status was reported
        pub timestamp: DateTime<Utc>,
        /// A human readable string describing the why it is in a waiting status
        pub message: String,
     },
    ....
```

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -137,14 +256,25 @@ mod test {
...
-        let mut stdout = String::default();
+        let mut handle = wr.start().await.expect("runtime handle");
+        handle.wait().await.expect("successful run");
+
+        let mut output = Vec::new();
+        handle.output(&mut output).await.unwrap();
+        assert_eq!("Hello, world!\n".to_string().into_bytes(), output);
+
+        let status = handle.status().await.unwrap();
+        assert!(match status {
```
rylev: Can't wait for CI to update to the latest Rust so we can use `matches!`

crates/kubelet/src/kubelet.rs
thomastaylor312: I think I’ll change this, but for discussion sake: What are your thoughts when it comes to convenience for API use though? Others who consume the crate might pass a literal or only have a borrow and it feels nicer to not force the to clone or use into string. Is there general guidance for this when making rust apis?

crates/wasi-provider/src/handle.rs
thomastaylor312: I’ll double check, but I think it didn’t like other options

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: I had that in there but their next release is 4-5 days out

crates/kubelet/src/kubelet.rs
thomastaylor312: I think I’ll change this, but for discussion sake: What are your thoughts when it comes to convenience for API use though? Others who consume the crate might pass a literal or only have a borrow and it feels nicer to not force the to clone or use into string. Is there general guidance for this when making rust apis?

crates/kubelet/src/kubelet.rs
thomastaylor312: I honestly forgot this was an option. Will switch out

crates/kubelet/src/kubelet.rs
rylev: By taking &str here you're forcing a clone even if there doesn't need to be one. By taking a `String` here you're letting the caller know what you need and allowing them to give you it as efficiently as possible. With `&str` you'll have to clone every time even if the caller had an owned string they could give you.

crates/wasi-provider/src/handle.rs
thomastaylor312: Yeah, it needs this because `Future` is only implemented for `mut` according to the compiler

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -64,25 +75,33 @@ impl WasiRuntime {
...
+            output: Arc::new(temp),
         })
     }
 
     pub async fn start(&self) -> Result<RuntimeHandle<tokio::fs::File>, failure::Error> {
-        let output = self.output.reopen()?;
+        let temp = self.output.clone();
+        // Because a reopen is blocking, run in a blocking task to get new
+        // handles to the tempfile
+        let (output_write, output_read) = tokio::task::spawn_blocking(
```
rylev: Like this names 👍

PR #63, opened 2020-03-12
Small changes to OCI repo client

rylev:


crates/oci-distribution/src/lib.rs
```diff
@@ -63,37 +62,36 @@ impl Client {
...
-        }
-        let auth = WwwAuthenticate::parse_header(&disthdr.unwrap().as_bytes().into())?;
-        let challenge_opt = auth.get::<BearerChallenge>();
-        if challenge_opt.is_none() {
-            // This means that no challenge was present, even though the header was present.
-            // Since we do not handle basic auth, it could be the case that the upstream service
-            // is in compatibility mode with a Docker v1 registry.
-            return Ok(());
-        }
+        let dist_hdr = match res.headers().get(reqwest::header::WWW_AUTHENTICATE) {
```
technosophos: Oh! I see! This is a great way of doing things.

crates/oci-distribution/src/lib.rs
```diff
@@ -129,34 +126,29 @@ impl Client {
...
-            // According to the OCI spec, we should see an error in the message body.
-            let err = res.json::<OciEnvelope>().await?;
-            // FIXME: This should not have to wrap the error.
-            Err(format_err!("{} on {}", err.errors[0], url))
-        } else if status.is_server_error() {
-            Err(format_err!("Server error at {}", url))
-        } else {
-            let text = res.text().await?;
-            Err(format_err!(
+        match res.status() {
```
technosophos: Ooo! I like this, too! I definitely have not gotten the hang of the match/if syntax.

crates/oci-distribution/src/lib.rs
```diff
@@ -201,14 +183,18 @@ impl Challenge for BearerChallenge {
...
+
     fn into_raw(self) -> RawChallenge {
         let mut map = ChallengeFields::new();
-        self.realm
-            .and_then(|realm| map.insert_static_quoting("realm", realm));
-        self.scope
-            .and_then(|item| map.insert_static_quoting("scope", item));
-        self.service
-            .and_then(|item| map.insert_static_quoting("service", item));
+        if let Some(realm) = self.realm {
```
technosophos: Okay, I have to say that I find the if/map thing less readable than the and_then() thing. Is this an established pattern or a preference?

crates/oci-distribution/Cargo.toml
```diff
@@ -16,7 +16,7 @@ edition = "2018"
 reqwest = { version = "0.10.4", features = ["json"] }
 failure = "0.1"
 tokio = "0.2.13"
-serde = "1.0"
+serde = { version = "1.0", features = ["derive"] }
```
technosophos: Still trying to learn about how features work... does this mean "the feature `derive` is added" or "all the features except `derive` are disabled"?

crates/oci-distribution/src/lib.rs
rylev: I would say using `if let` is much more common than `and_then` in this case. `and_then` is really best when wanting to do the same as a `map` then `flatten`. We simply want to perform an action if the item is set.

crates/oci-distribution/Cargo.toml
rylev: Features are generally additive (i.e., they _should_ be additive but not every crate follows that recommended pattern). So putting the `derive` feature here adds additional functionality (namely the ability to derive `Serialize` and `Deserialize`). You may see `default-features = false` which turns off all features that are marked as default in that dependency's Cargo.toml file. You can then add features on top of that.

PR #64, opened 2020-03-12
SSL Web Server

rylev:
Provides SSL bindings for the web server


crates/kubelet/src/server.rs
```diff
@@ -20,7 +26,14 @@ pub async fn start_webserver<T: 'static + Provider + Send + Sync>(
     provider: Arc<Mutex<T>>,
     address: &SocketAddr,
 ) -> Result<(), failure::Error> {
-    let service = make_service_fn(move |_conn: &AddrStream| {
+    let identity = tokio::fs::read("/home/rylevick/krustlet/certificate.pfx")
```
rylev: Obviously this is not correct 😛 - The question is how should this be configured.

thomastaylor312: Definitely pull it from the `Config` object and expose it as a flag and env variable. As for local dev, we should probably have a `just` target that makes a self-signed cert and puts it in a default location

thomastaylor312: Did we want to use a pfx bundle or a separate cert and key combo?

crates/kubelet/src/server.rs
```diff
@@ -49,7 +62,21 @@ pub async fn start_webserver<T: 'static + Provider + Send + Sync>(
...
-    let server = Server::bind(address).serve(service);
+
+    let mut listener = TcpListener::bind(address).await.unwrap();
+    let mut incoming = listener.incoming();
+    let accept = hyper::server::accept::from_stream(stream! {
+        loop {
+            match incoming.next().await {
+                Some(Ok(stream)) => match acceptor.clone().accept(stream).await {
+                    result @ Ok(_) => yield result,
+                    Err(e) => break,
```
thomastaylor312: Should we log and continue the loop instead of breaking here?

crates/kubelet/src/server.rs
technosophos: Does OpenSSL's CLI allow you to generate pfx bundles?

rylev: This is what I did locally on my Ubuntu box:
```
openssl genrsa 2048 > host.key
chmod 400 host.key
openssl req -new -x509 -nodes -sha256 -days 365 -key host.key -out host.cert
openssl pkcs12 -export -out certificate.pfx -inkey host.key -in host.cert
```

crates/kubelet/src/server.rs
rylev: Yea that makes sense

crates/kubelet/src/config.rs
```diff
@@ -30,16 +35,20 @@ impl Config {
...
+            server_config: ServerConfig {
+                addr: match preferred_ip_family {
+                    // Just unwrap these because they are programmer error if they
+                    // don't parse
+                    IpAddr::V4(_) => "0.0.0.0".parse().unwrap(),
+                    IpAddr::V6(_) => "::".parse().unwrap(),
+                },
+                port: DEFAULT_PORT,
+                pfx_password: String::new(),
+                pfx_path: PathBuf::new(),
```
thomastaylor312: This should have the default path set as well

crates/kubelet/src/config.rs
```diff
@@ -62,12 +71,20 @@ impl Config {
             default_node_ip(&mut hostname.clone(), &addr)
                 .expect("unable to get default node IP address")
         });
+        let pfx_path = opts
+            .pfx_path
+            .unwrap_or_else(|| PathBuf::from("./config/certificate.pfx"));
```
thomastaylor312: Not something that needs to be addressed now, but we should probably define a common path (like `~/.krustlet`) that things like certs and pod logs can go into by default

crates/kubelet/src/server.rs
```diff
@@ -18,38 +25,66 @@ use crate::kubelet::Provider;
...
+                                pod.to_string(),
+                                container.to_string(),
+                            )
+                            .await
+                        }
+                        (&Method::POST, [_, "exec", _, _, _]) => {
+                            post_exec(&*provider.lock().await, &req)
+                        }
+                        _ => {
+                            let mut response = Response::new(Body::from("Not Found"));
```
thomastaylor312: Should this use the builder pattern instead?

justfile
```diff
@@ -13,21 +13,29 @@ test:
...
 
 dockerize:
     docker build -t technosophos/krustlet:latest .
 
 push:
     docker push technosophos/krustlet:latest
 
+bootstrap-ssl:
+    mkdir -p config
+    rm -rf config/*
```
thomastaylor312: This should probably check if the cert already exists before doing the steps every time. Not a blocker though

PR #68, opened 2020-03-13
Switch to anyhow

### LINKED ISSUE #52: Move from failure to anyhow
We agreed awhile back that we wanted to switch to the anyhow crate over the failure crate. This will need to be done in all of our crates simultaneously to avoid error conversions or other conflicts

rylev:
Fixes #52

### DIFF
--- a/crates/oci-distribution/src/reference.rs
+++ b/crates/oci-distribution/src/reference.rs
@@ -48,7 +48,10 @@ impl TryFrom<&str> for Reference {
     type Error = anyhow::Error;
     fn try_from(string: &str) -> Result<Self, Self::Error> {
         let slash = string.find('/').ok_or_else(|| {
-            anyhow::anyhow!("Failed to pare {}. Expected at least one slash (/)", string)
+            anyhow::anyhow!(
+                "Failed to parse {}. Expected at least one slash (/)",
+                string
+            )
         })?;
         let colon = string[slash + 1..].find(':').ok_or_else(|| {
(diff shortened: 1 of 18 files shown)


crates/oci-distribution/src/reference.rs
```diff
@@ -39,20 +38,20 @@ impl Reference {
...
     }
 }
 
 impl TryFrom<&str> for Reference {
-    type Error = failure::Error;
+    type Error = anyhow::Error;
     fn try_from(string: &str) -> Result<Self, Self::Error> {
         let slash = string.find('/').ok_or_else(|| {
-            format_err!("Failed to pare {}. Expected at least one slash (/)", string)
+            anyhow::anyhow!("Failed to pare {}. Expected at least one slash (/)", string)
```
thomastaylor312: Spelling: parse

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -156,7 +156,7 @@ impl WasiRuntime {
                         })
                         .expect("status should be able to send");
                     // Converting from anyhow
```
thomastaylor312: Nit: these comments are no longer needed and I actually think you can return `Err(e)` because they are already `anyhow` errors

rylev: I'll get rid of the comment but the call to `anyhow!` allows us to add the message as context so I'll keep that.

PR #69, opened 2020-03-13
fix(justfile): Fixes "wrong password" issue with generated certs

thomastaylor312:
Based on testing from the team and what I found on the internet (always
a reliable source of information), having an empty password with a
pkcs bundle can cause errors. This exports a default password to be
used with the certificate.

Also, I removed the kubelet=debug setting for logging because kubelet is
a _really_ verbose default

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,4 +1,5 @@
-export RUST_LOG := "wascc_host=debug,wascc_provider=debug,wasi_provider=debug,main=debug,kubelet=debug"
+export RUST_LOG := "wascc_host=debug,wascc_provider=debug,wasi_provider=debug,main=debug"
+export PFX_PASSWORD := "testing"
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -1,4 +1,5 @@
-export RUST_LOG := "wascc_host=debug,wascc_provider=debug,wasi_provider=debug,main=debug,kubelet=debug"
+export RUST_LOG := "wascc_host=debug,wascc_provider=debug,wasi_provider=debug,main=debug"
```
rylev: Why did you remove the `kubelet` log level? too noisy?

thomastaylor312: > Also, I removed the kubelet=debug setting for logging because kubelet is
a really verbose default

PR #70, opened 2020-03-13
Add image layer pull

technosophos:
This adds support for _layer_ pulls on OCI images. OCI images have two artifact types that have to be pulled: manifests and layers. This adds support for pulling layers into local writers.

I am still working on this, but need help with one particular thing having to do with whether to use `tokio::io::AsyncWrite` or `std::io::Write`. Left a note inline.

The rest is not yet ready for review. Still need:

- [ ] ~A high-level `pull()` function~
- [ ] ~A way to describe where/how to store the full image~

*UPDATE:* To keep this from blocking other work, I have pared it back down to the minimum. I have addressed almost all of @rylev 's comments, but have reached out to him for a follow-up PR on the AsyncWrite in the loop.

### DIFF
--- a/crates/oci-distribution/src/lib.rs
+++ b/crates/oci-distribution/src/lib.rs
@@ -35,8 +35,18 @@ pub struct Client {
     client: reqwest::Client,
 }
 
+/*
 impl Default for Client {
     fn default() -> Self {
+        Client {
+            token: None,
+            client: reqwest::Client::new(),
+        }
+    }
+}
+*/
+impl Client {
+    fn new() -> Self {
         Client {
             token: None,
             client: reqwest::Client::new(),
@@ -112,22 +122,11 @@ impl Client {
         }
     }
 
-    fn auth_headers(&self) -> HeaderMap {
-        let mut headers = HeaderMap::new();
-        headers.insert("Accept", "application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.list.v2+json".parse().unwrap());
-
-        if let Some(bearer) = self.token.as_ref() {
-            headers.insert("Authorization", bearer.bearer_token().parse().unwrap());
-        }
-        headers
-    }
-
     /// Pull a manifest from the remote OCI Distribution service.
     ///
     /// If the connection has already gone through authentication, this will
     /// use the bearer token. Otherwise, this will attempt an anonymous pull.
     pub async fn pull_manifest(&self, image: &Reference) -> OciResult<OciManifest> {
-        // We unwrap right now because this try_from literally cannot fail.
         let url = image.to_v2_manifest_url();
         let request = self.client.get(&url);
 
@@ -160,7 +159,7 @@ impl Client {
     /// repository and the registry, but it is not used to verify that
     /// the digest is a layer inside of the image. (The manifest is
     /// used for that.)
-    pub async fn pull_layer<T: std::io::Write>(
+    pub async fn pull_layer<T: std::io::Write + tokio::io::AsyncWrite>(
         &self,
         image: &Reference,
         digest: &str,
@@ -176,10 +175,26 @@ impl Client {
             .bytes_stream();
 
         while let Some(bytes) = stream.next().await {
-            out.write_all(&bytes?.to_vec())?;
+            //out.write_all(&bytes?.to_vec())?;
+            out.write_all(&bytes?)?;
         }
 
         Ok(out)
+    }
+
+    /// Generate the headers necessary for authentication.
+    ///
+    /// If the struct has Some(bearer), this will insert the bearer token in an
+    /// Authorization header. It will also set the Accept header, which must
+    /// be set on all OCI Registry request.
+    fn auth_headers(&self) -> HeaderMap {
+        let mut headers = HeaderMap::new();
+        headers.insert("Accept", "application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.list.v2+json".parse().unwrap());
+
+        if let Some(bearer) = self.token.as_ref() {
+            headers.insert("Authorization", bearer.bearer_token().parse().unwrap());
+        }
+        headers
     }
 }
 
@@ -244,7 +259,7 @@ mod test {
 
     #[tokio::test]
     async fn test_version() {
-        let c = Client::default();
+        let c = Client::new();
         let ver = c
             .version("webassembly.azurecr.io")
             .await
@@ -254,8 +269,8 @@ mod test {
 
     #[tokio::test]
     async fn test_auth() {
-        let image = Reference::try_from(HELLO_IMAGE).expect("parsed reference");
-        let mut c = Client::default();
+        let image = Reference::try_from(HELLO_IMAGE).expect("failed to parse reference");
+        let mut c = Client::new();
         c.auth(&image, None)
             .await
             .expect("result from auth request");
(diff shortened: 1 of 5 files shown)


crates/oci-distribution/src/lib.rs
```diff
@@ -145,6 +153,35 @@ impl Client {
...
     }
+
+    /// Pull a single layer from an OCI registy.
+    ///
+    /// This pulls the layer for a particular image that is identified by
+    /// the given digest. The image reference is used to find the
+    /// repository and the registry, but it is not used to verify that
+    /// the digest is a layer inside of the image. (The manifest is
+    /// used for that.)
+    pub async fn pull_layer<T: std::io::Write>(
```
technosophos: @rylev So here's where I am a little stuck: I want to be able to write simple unit tests that consume the stream into a buffer so I can test the size and such. But in most real-world cases, the `Write` implementation should be async. I can't figure out how to satisfy the first using `tokio::io::AsyncWrite`. But it does look like `tokio::fs::File` and family do support `Write`. Any suggestions on what I should do? I've also marked the related test with a comment.

crates/oci-distribution/src/lib.rs
```diff
@@ -252,4 +289,24 @@ mod test {
...
+
+    #[tokio::test]
+    async fn test_pull_layer() {
+        let image = Reference::try_from(HELLO_IMAGE).expect("parsed reference");
+        let mut c = Client::default();
+        c.auth(&image, None).await.expect("authenticated");
+        let manifest = c.pull_manifest(&image).await.expect("pull manifest");
+
+        // Pull one specific layer
+        let file: Vec<u8> = Vec::new();
```
technosophos: My test is basically writing the data to a vector, where I can run a few trivial tests to make sure the right amount od data was sucked down. (Ignore, for the moment, the network-bound nature of this test.)

crates/oci-distribution/src/lib.rs
```diff
@@ -108,23 +113,26 @@ impl Client {
...
-    /// Pull a manifest from the remote OCI Distribution service.
-    ///
-    /// If the connection has already gone through authentication, this will
-    /// use the bearer token. Otherwise, this will attempt an anonymous pull.
-    pub async fn pull_manifest(&self, image: &Reference) -> OciResult<OciManifest> {
-        let client = reqwest::Client::new();
-        let url = image.to_v2_manifest_url();
-        let request = client.get(&url);
-
+    fn auth_headers(&self) -> HeaderMap {
```
rylev: nit: might be more readable to have private functions come after public ones, but I'm not sure.

crates/oci-distribution/src/lib.rs
```diff
@@ -108,23 +113,26 @@ impl Client {
...
+        headers
+    }
 
-        let res = request.headers(headers).send().await?;
+    /// Pull a manifest from the remote OCI Distribution service.
+    ///
+    /// If the connection has already gone through authentication, this will
+    /// use the bearer token. Otherwise, this will attempt an anonymous pull.
+    pub async fn pull_manifest(&self, image: &Reference) -> OciResult<OciManifest> {
+        // We unwrap right now because this try_from literally cannot fail.
```
rylev: Does this comment still make sense? There's no unwrapping here.

crates/oci-distribution/src/lib.rs
rylev: This should be `AsyncWrite`. What errors do you get when you try with `AsyncWrite`?

crates/oci-distribution/src/lib.rs
```diff
@@ -145,6 +153,35 @@ impl Client {
...
+        let mut stream = self
+            .client
+            .get(&url)
+            .headers(self.auth_headers())
+            .send()
+            .await?
+            .bytes_stream();
+
+        while let Some(bytes) = stream.next().await {
+            out.write_all(&bytes?.to_vec())?;
```
rylev: Do you need the call to `to_vec`? I believe `bytes?.as_ref()` would work as well and avoids an allocation

crates/oci-distribution/src/lib.rs
```diff
@@ -215,8 +255,7 @@ mod test {
 
     #[tokio::test]
     async fn test_auth() {
-        let image =
-            Reference::try_from("webassembly.azurecr.io/hello-wasm:v1").expect("parsed reference");
+        let image = Reference::try_from(HELLO_IMAGE).expect("parsed reference");
```
rylev: Nit: the expect message is what gets printed when unwrapping fails. It should read like "could not parse reference".

crates/oci-distribution/src/lib.rs
```diff
@@ -32,11 +33,15 @@ type OciResult<T> = Result<T, failure::Error>;
 /// unless you are sure that the remote registry does not require Oauth2.
 pub struct Client {
     token: Option<RegistryToken>,
+    client: reqwest::Client,
 }
 
 impl Default for Client {
     fn default() -> Self {
```
rylev: In my opinion, this makes more sense as a `new` function rather than a `Default` implementation.

crates/oci-distribution/src/lib.rs
rylev: `Vec<u8>` implements `AsyncWrite` so this should work with `AsyncWrite` instead of `Write`

crates/oci-distribution/src/lib.rs
```diff
@@ -252,4 +289,24 @@ mod test {
...
+        let mut c = Client::default();
+        c.auth(&image, None).await.expect("authenticated");
+        let manifest = c.pull_manifest(&image).await.expect("pull manifest");
+
+        // Pull one specific layer
+        let file: Vec<u8> = Vec::new();
+        let layer0 = manifest.layers[0].clone();
+
+        let file = c
+            .pull_layer(&image, layer0.digest.as_str(), file)
```
rylev: nit: `as_str` isn't necessary. You can just pass the digest `String` by reference and it will automatically be coerced into a `&str`.

crates/oci-distribution/src/lib.rs
```diff
@@ -252,4 +289,24 @@ mod test {
...
+    #[tokio::test]
+    async fn test_pull_layer() {
+        let image = Reference::try_from(HELLO_IMAGE).expect("parsed reference");
+        let mut c = Client::default();
+        c.auth(&image, None).await.expect("authenticated");
+        let manifest = c.pull_manifest(&image).await.expect("pull manifest");
+
+        // Pull one specific layer
+        let file: Vec<u8> = Vec::new();
+        let layer0 = manifest.layers[0].clone();
```
rylev: The `clone` here isn't neccessary. You can just borrow like this: `&manifest.layers[0]`

crates/oci-distribution/src/lib.rs
technosophos: Oh, good catch! Removing that.

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: Done.

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: I just realized that I needed `Write + AsyncWrite`. Then it started working. Thanks!

crates/oci-distribution/src/lib.rs
technosophos: Fixed

crates/oci-distribution/src/lib.rs
technosophos: Fixed

technosophos:
@rylev Okay, I switched to using `tokio::io::AsycWriteExt`, and that seems to work fine. There may still be a more elegant solution. But this one appears to at least be non-blocking as-is.

PR #71, opened 2020-03-16
Fix environment variables from secrets.

peterhuene:
The `string_data` field of `Secret` is a write-only convenience and will always
be `None` when read.

This commit fixes it so that the `data` field is used instead.

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -438,18 +438,17 @@ async fn on_missing_value(
             match Api::<Secret>::namespaced(client, ns).get(name).await {
                 Ok(secret) => {
                     // I am not totally clear on what the outcome should
-                    // be of a cfgmap key miss. So for now just return an
+                    // be of a secret key miss. So for now just return an
                     // empty default.
-
                     return secret
-                        .string_data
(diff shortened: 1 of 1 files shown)


crates/kubelet/src/kubelet.rs
```diff
@@ -438,18 +438,17 @@ async fn on_missing_value(
...
                     // empty default.
-
                     return secret
-                        .string_data
+                        .data
                         .unwrap_or_default()
-                        .get(&seckey.key)
-                        .cloned()
+                        .remove(&seckey.key)
+                        .map(|s| String::from_utf8(s.0).unwrap_or_default())
```
rylev: Is non-ut8 data possible in this field?

peterhuene: Secrets in k8s are binary blobs, so it is possible. I think we should just error pod creation if the specified secret can't be represented as an environment variable.  The error handling in this method is currently pretty lax, so this didn't attempt to fix it.

thomastaylor312: I agree with @peterhuene here, we can improve the error handling down the road as we need it

Issue #61, opened 2020-03-12
Get Status from provider


thomastaylor312:
Right now each provider updates the pod status itself. This should be done in the Kubelet logic by calling the provider's `status` method. ~Also, status should return an array of container statuses that should be set on the pod as well~

thomastaylor312:
I am taking care of adding container statuses to the Status object in #62

thomastaylor312:
I think I have a good initial way for doing this. Going to give it a whirl and see what people think

PR #72, opened 2020-03-16
Use Phase enum everywhere

rylev:


thomastaylor312:
@rylev and I talked about this on a call and it is not needed as we will be using the `Status` object to patch things as we implement #61

PR #73, opened 2020-03-16
Wrap the Pod struct from k8s-openapi

rylev:


crates/kubelet/src/pod.rs
```diff
@@ -1,33 +1,149 @@
...
-/// Patch the pod status to update the phase.
-pub async fn pod_status(client: APIClient, pod: &Pod, phase: &str, ns: &str) {
-    let status = serde_json::json!(
-        {
-            "metadata": {
-                "resourceVersion": "",
-            },
-            "status": {
-                "phase": phase
+/// A Kubertnetes Pod
```
thomastaylor312: Nit: Spelling -> Kubernetes

PR #74, opened 2020-03-16
fix(actions): Fixes GH actions checkout with version upgrade

thomastaylor312:
The old v1 had issues re-running builds if master had changed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,10 +9,20 @@ jobs:
       fail-fast: false
       matrix:
         config:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -9,10 +9,20 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - {os: "ubuntu-latest", url: "https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz", name: "just", pathInArchive: "just" }
-        - {os: "macos-latest", url: "https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-apple-darwin.tar.gz", name: "just", pathInArchive: "just" }
+          - {
```
thomastaylor312: This was just formatting that ran automatically, but I think it looks nicer anyway, so I kept it

Issue #49, opened 2020-03-09
Better error handling for logs


thomastaylor312:
Right now the error returned to the user with `kubectl logs` if something fails during the call to `Provider.logs` is always "Not Implemented." It should be smarter and return the proper error code and a useful message

PR #80, opened 2020-03-18
feat(e2e): wait for pod to enter the Running state

bacongobbler:

### DIFF
--- a/crates/kubelet/src/server.rs
+++ b/crates/kubelet/src/server.rs
@@ -17,7 +17,7 @@ use tokio::sync::Mutex;
 use std::sync::Arc;
 
 use crate::config::ServerConfig;
-use crate::kubelet::Provider;
+use crate::kubelet::{NotImplementedError, Provider};
 
 /// Start the Krustlet HTTP(S) server
 ///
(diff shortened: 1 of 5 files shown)


tests/integration_tests.rs
```diff
@@ -74,7 +76,40 @@ async fn test_wascc_provider() {
...
+        .expect("failed to poll for pod events")
+        .boxed();
+
+    while let Some(event) = watcher
+        .try_next()
+        .await
+        .expect("failed to poll for a pod event")
+    {
+        match event {
+            WatchEvent::Modified(o) => {
```
thomastaylor312: There should probably be a timeout here so that if it doesn't reach running that it doesn't keep going forever

bacongobbler: There's a ListParams timeout set above! Does that satisfy it, or is there another timeout we need to set?

bacongobbler:
added more e2e tests to check the output of `kubectl logs`. Also verifies and closes #49

bacongobbler:
Oh... I also refactored the tests to return a `Result`, cleaning up a lot of the code. Let me know if I should pull that out into a separate PR.

tests/integration_tests.rs
thomastaylor312: Oh I skipped right over that part. I'm not used to list params having timeouts

crates/kubelet/src/server.rs
```diff
@@ -122,11 +122,14 @@ async fn get_container_logs<T: Provider + Sync>(
     }
     match provider.logs(namespace, pod, container).await {
         Ok(data) => Response::new(Body::from(data)),
-        // TODO: This should detect not implemented vs. regular error
         Err(e) => {
             error!("Error fetching logs: {}", e);
-            let mut res = Response::new(Body::from("Not Implemented"));
-            *res.status_mut() = StatusCode::NOT_IMPLEMENTED;
+            let mut res = Response::new(Body::from(format!("Server error: {}", e)));
```
thomastaylor312: Nit: This and the status should probably use the builder patter like we are doing elsewhere

bacongobbler: Do you have a link to an example? Still wading my way around the codebase

PR #81, opened 2020-03-18
Feat/status updates

### LINKED ISSUE #61: Get Status from provider

thomastaylor312:
Updating status is now working for wasi. This has all been added as part
of the `wasi_runtime::handle` module, and so should be portable to
wascc when it is ready. It introduces a new `PodHandle` which is a higher
level (and possibly better?) abstraction for interacting with a running Pod.

I also removed the `status` method from the provider interface as it should
be done with handles and not by the kubelet calling the provider.

Closes #61

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,20 +35,24 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: engineerd/setup-kind@v0.3.0
-    - uses: engineerd/configurator@v0.0.1
-      with:
-        name: just
-        url: https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz
-        pathInArchive: just
-    - name: Run e2e tests
-      run: |
-        cd crates/wascc-provider
-        cargo build --bin krustlet-wascc --manifest-path ../../Cargo.toml
-        ../../target/debug/krustlet-wascc --node-name krustlet-wascc --node-ip 172.17.0.1 --port 3000 &
-        cd ../wasi-provider
-        cargo build --bin krustlet-wasi --manifest-path ../../Cargo.toml
-        ../../target/debug/krustlet-wasi --node-name krustlet-wasi --node-ip 172.17.0.1 --port 3001 &
-        cd ../..
-        just test-e2e
+      - uses: actions/checkout@v2
+      - uses: engineerd/setup-kind@v0.3.0
+      - uses: engineerd/configurator@v0.0.1
+        with:
+          name: just
+          url: https://github.com/casey/just/releases/download/v0.5.8/just-v0.5.8-x86_64-unknown-linux-musl.tar.gz
+          pathInArchive: just
+      - name: Run e2e tests
+        env:
+          KRUSTLET_NODE_IP: "172.17.0.1"
+          # The default location for the cert is not accessible on build hosts, so do it in the current dir
+          KEY_DIR: "./config"
+          PFX_PASSWORD: "testing"
+        run: |
+          just bootstrap-ssl
+          just build
+          cd crates/wascc-provider
+          ../../target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 --pfx-path ../../config/certificate.pfx &
+          cd ../wasi-provider
+          ../../target/debug/krustlet-wasi --node-name krustlet-wasi --port 3001 --pfx-path ../../config/certificate.pfx &
+          just test-e2e
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -1,5 +1,5 @@
-/// This library contains the Kubelet shell. Use this to create a new Kubelet
-/// with a specific handler. (The handler included here is the WASM handler.)
+/// This library contains code for running a kubelet. Use this to create a new
+/// Kubelet with a specific handler (called a `Provider`)
 use crate::{
     config::Config,
     node::{create_node, update_node},
@@ -48,12 +48,18 @@ pub enum Phase {
     Unknown,
 }
 
+impl Default for Phase {
+    fn default() -> Self {
+        Self::Unknown
+    }
+}
+
 /// Describe the status of a workload.
 ///
 /// Phase captures the lifecycle aspect of the workload, while
 /// the message provides a human-readable description of the
 /// state of the workload.
-#[derive(Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct Status {
     pub phase: Phase,
     pub message: Option<String>,
@@ -269,12 +275,6 @@ pub trait Provider {
     /// Pods that are sent to this function have already met certain criteria for modification.
     /// For example, updates to the `status` of a Pod will not be sent into this function.
     async fn modify(&self, pod: Pod, client: APIClient) -> anyhow::Result<()>;
-
-    /// Given a pod, determine the status of the underlying workload.
-    ///
-    /// This information is used to update Kubernetes about whether this workload is running,
-    /// has already finished running, or has failed.
-    async fn status(&self, pod: Pod, client: APIClient) -> anyhow::Result<Status>;
 
     /// Given the definition of a deleted Pod, remove the workload from the runtime.
     ///
@@ -533,13 +533,6 @@ mod test {
         }
         async fn modify(&self, _pod: Pod, _client: APIClient) -> anyhow::Result<()> {
             Ok(())
-        }
-        async fn status(&self, _pod: Pod, _client: APIClient) -> anyhow::Result<Status> {
-            Ok(Status {
-                phase: Phase::Succeeded,
-                message: None,
-                container_statuses: Vec::new(),
-            })
         }
     }
 
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -64,14 +64,14 @@ pub struct Status {
     /// Allows a provider to set a custom message, otherwise, kubelet will infer
     /// a message from the container statuses
     pub message: Option<String>,
-    pub container_statuses: Vec<ContainerStatus>,
+    pub container_statuses: HashMap<String, ContainerStatus>,
 }
 
 /// ContainerStatus is a simplified version of the Kubernetes container status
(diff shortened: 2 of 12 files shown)


bacongobbler:
Looks like there are a few failing CI tests due to calling `.unwrap` on an `&str`. Let me know when this is ready for review

crates/kubelet/src/kubelet.rs
```diff
@@ -37,7 +37,7 @@ pub struct NotImplementedError;
 ///
 /// This is specified by Kubernetes itself.
 #[derive(Clone, Debug, serde::Serialize)]
-pub enum Phase {
+pub(crate) enum Phase {
```
bacongobbler: would it be useful for providers to inspect the pod's phase?

crates/kubelet/src/server.rs
```diff
@@ -49,9 +49,9 @@ pub async fn start_webserver<T: 'static + Provider + Send + Sync>(
                             get_container_logs(
                                 &*provider.lock().await,
                                 &req,
-                                namespace.to_string(),
-                                pod.to_string(),
-                                container.to_string(),
+                                (*namespace).to_string(),
```
bacongobbler: out of curiosity, what's the reason for this change? How come we have to deref here?

crates/wasi-provider/src/lib.rs
```diff
@@ -69,34 +66,39 @@ impl Provider for WasiProvider {
...
+                PodHandle::new(container_handles, pod, client)?,
+            );
         }
         info!(
             "All containers started for pod {:?}. Updating status",
-            pod.name()
+            pod_name
         );
-        pod.patch_status(client, &Phase::Running).await;
+        // pod.patch_status(client, &Phase::Running).await;
```
bacongobbler: should this be removed completely?

crates/kubelet/src/kubelet.rs
thomastaylor312: Based on how I have it implemented now, there should be no need to check the phase outside of the crate

crates/kubelet/src/server.rs
thomastaylor312: This is a clippy suggestion. Basically `namespace` is a `&&` which has a slower implementation of to_string

crates/wasi-provider/src/lib.rs
thomastaylor312: Oops, yes

thomastaylor312:
I'd love to add e2e tests for this! I was actually going to rebase on your PR once you've merged

tests/integration_tests.rs
```diff
@@ -174,12 +174,17 @@ async fn test_wasi_provider() -> Result<(), Box<dyn std::error::Error>> {
...
+    let mut found_running = false;
     while let Some(event) = watcher.try_next().await? {
         match event {
             WatchEvent::Modified(o) => {
                 let phase = o.status.unwrap().phase.unwrap();
                 if phase == "Running" {
+                    found_running = true;
+                }
+                if phase == "Succeeded" && !found_running {
+                    panic!("Reached completed phase before receiving Running phase")
```
bacongobbler: could there ever be a case where we see the completed phase before the running phase from the events system, or is the order guaranteed?

thomastaylor312: I think eventing order is guaranteed. This was mostly to check that we actually got a running status reported at some point

thomastaylor312:
@bacongobbler Wascc doesn't have any status updating yet, so it is not included in tests. It was just updated to match the new format

.github/workflows/build.yml
```diff
@@ -35,20 +35,23 @@ jobs:
...
+          pathInArchive: just
+      - name: Run e2e tests
+        env:
+          KRUSTLET_NODE_IP: "172.17.0.1"
+          # The default location for the cert is not accessible on build hosts, so do it in the current dir
+          KEY_DIR: "./config"
+          PFX_PATH: "./config/certificate.pfx"
+        run: |
+          ifconfig
+          nslookup `hostname`
```
rylev: Can we add a comment to the `ifconfig/nslookup` magic

crates/kubelet/src/pod.rs
```diff
@@ -85,27 +86,88 @@ impl Pod {
...
+            }
+        };
+
+        // This section figures out what the current phase of the pod should be
+        // based on the container statuses
+        let current_statuses = status
+            .container_statuses
+            .into_iter()
+            .map(|s| s.to_kubernetes(name.to_owned()))
+            .collect::<Vec<KubeContainerStatus>>();
```
rylev: Should we collect this into a `HashMap<String, KubeContainerStatus>` where the keys are the `name`s so we can do quick lookups in the filter below?

crates/kubelet/src/pod.rs
```diff
@@ -85,27 +86,88 @@ impl Pod {
...
+            .into_iter()
+            .filter(|s| {
+                current_statuses
+                    .iter()
+                    .find(|&other| other.name == s.name)
+                    .is_none()
+            })
+            .collect::<Vec<KubeContainerStatus>>();
+        container_statuses.extend(current_statuses);
+        let mut num_terminated: usize = 0;
```
rylev: `terminated` can have a connotation of failure. Is `num_succeed` a better name?

crates/kubelet/src/server.rs
rylev: Out of curiosity, does `to_owned` lead to the same clippy warning?

crates/wascc-provider/src/lib.rs
```diff
@@ -105,10 +105,30 @@ impl Provider for WasccProvider {
                 tokio::task::spawn_blocking(move || wascc_run_http(data, env, &pub_key)).await?;
             match http_result {
                 Ok(_) => {
-                    pod.patch_status(client.clone(), &Phase::Running).await;
+                    pod.patch_status(
+                        client.clone(),
+                        Status {
```
rylev: I think all these `patch_status` calls in this PR would be much easier to read if the status arguments were not inlined and instead moved to a local variable.

crates/wasi-provider/src/handle.rs
```diff
@@ -46,25 +52,125 @@ impl<R: AsyncReadExt + AsyncSeekExt + Unpin> RuntimeHandle<R> {
...
+    ) -> anyhow::Result<usize> {
+        let mut handles = self.container_handles.write().await;
+        let handle =
+            handles
+                .get_mut(container_name)
+                .ok_or_else(|| ProviderError::ContainerNotFound {
+                    pod_name: self.pod.name().to_owned(),
+                    container_name: container_name.to_owned(),
+                })?;
+        handle.output(buf).await
```
rylev: The way that this is architected, we'll hold on to the write lock while `handle.output` is running. That's ok right?

crates/wasi-provider/src/lib.rs
```diff
@@ -69,34 +66,38 @@ impl Provider for WasiProvider {
         //   - mount any volumes (popen)
         //   - run it to completion
         //   - bail if it errors
-        info!("Starting containers for pod {:?}", pod.name());
+        let pod_name = pod.name().to_owned();
```
rylev: Is to_owned needed? Is this to get around borrowing?

tests/integration_tests.rs
```diff
@@ -193,9 +198,20 @@ async fn test_wasi_provider() -> Result<(), Box<dyn std::error::Error>> {
     let mut logs = pods.log_stream("hello-wasi", &LogParams::default()).await?;
 
     while let Some(line) = logs.try_next().await? {
-        assert_eq!("Hello World!\n", String::from_utf8_lossy(&line));
+        assert_eq!("Hello, world!\n", String::from_utf8_lossy(&line));
     }
 
+    let pod = pods.get("hello-wasi").await?;
+
+    let state = pod.status.unwrap().container_statuses.unwrap()[0]
```
rylev: This is super gnarly. I know that this is a test but can we clean this up?

crates/wasi-provider/src/handle.rs
```diff
@@ -46,25 +52,125 @@ impl<R: AsyncReadExt + AsyncSeekExt + Unpin> RuntimeHandle<R> {
...
-    // need to expose it, we can do that later
     pub(crate) async fn wait(&mut self) -> anyhow::Result<()> {
         (&mut self.handle).await.unwrap()
     }
 }
+
+/// PodHandle is the top level handle into managing a pod. It manages updating
+/// statuses for the containers in the pod and can be used to stop the pod and
+/// access logs
+pub struct PodHandle<R: AsyncReadExt + AsyncSeekExt + Unpin> {
```
rylev: I don't think you can just use `AsyncRead + AsyncSeek`

tests/integration_tests.rs
thomastaylor312: Have a way of doing it that doesn't involve another function? I tried a couple different things and none of them worked

crates/wasi-provider/src/handle.rs
thomastaylor312: Yes, because we need the mutable lock on that handle until we are done

crates/wasi-provider/src/lib.rs
thomastaylor312: Yes. This was because we need to move the pod into the new `PodHandle` further down

.github/workflows/build.yml
thomastaylor312: This was just debug material while I am getting the e2e tests working

crates/kubelet/src/pod.rs
thomastaylor312: Yes it is. Will change

crates/wasi-provider/src/handle.rs
thomastaylor312: I think I need Ext here as well. As in the previous PR, I don't think you can use the methods I do without those traits in scope

crates/wascc-provider/src/lib.rs
thomastaylor312: This better?

thomastaylor312:
@rylev Going to merge this for now. Completely open to a follow up PR or an issue for handling any other things you think we should fix

PR #82, opened 2020-03-19
oci_distribution::Client::pull

### LINKED ISSUE #76: OCI Registry: Add top-level `pull` method
RIght now, to use the OCI registry, you need to create a registry, then auth, then pull the manifest, then open a destination location, and fetch each layer.

We should have a top-level `pull()` function that takes an image and some kind of "where to store stuff" object, and can do the entire OCI thing.

rylev:
Fixes #76

### DIFF
--- a/crates/oci-distribution/src/lib.rs
+++ b/crates/oci-distribution/src/lib.rs
@@ -1,4 +1,6 @@
+use async_trait::async_trait;
 use chrono::prelude::{DateTime, Utc};
+use futures_util::future;
 use futures_util::stream::StreamExt;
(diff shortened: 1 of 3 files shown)


crates/oci-distribution/src/lib.rs
```diff
@@ -188,6 +222,43 @@ impl Client {
         headers
     }
 }
+#[async_trait]
+pub trait ImageStore {
```
radu-matei: I think in the CRI experiment we decided to call this a `ModuleStore`, as it is more accurate as to what it stores.

PR #97, opened 2020-03-23
Pull image in the wasi provider instead of hard coded

### LINKED ISSUE #96: Use non-hardcoded modules in wasi-provider
We are currently using a [hard coded WASM module](https://github.com/deislabs/krustlet/blob/master/crates/wascc-provider/src/lib.rs#L103) in the wasi-provider. This should be replaced with a module store that fetches the module if it's not there and returns the module from a cache.

rylev:
Fixes #96

### DIFF
--- a/crates/wasi-provider/src/lib.rs
+++ b/crates/wasi-provider/src/lib.rs
@@ -10,7 +10,7 @@ use kube::client::APIClient;
 use kubelet::pod::Pod;
 use kubelet::{Provider, ProviderError};
 use log::{debug, info};
-use oci_distribution::{Client, FileModuleStore, Reference};
+use oci_distribution::{FileModuleStore, Reference};
 use tokio::fs::File;
 use tokio::sync::RwLock;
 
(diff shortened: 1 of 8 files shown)


examples/greet.yaml
```diff
@@ -6,15 +6,15 @@ metadata:
     deislabs.io/wascc-action-key: MB4OLDIC3TCZ4Q4TGGOVAZC43VXFE2JQVRAXQMQFXUCREOOFEKOKZTY2
```
bacongobbler: this example appears to use the WasCC provider. should this be removed, or maybe a new example provided for the WASI provider instead?

crates/wasi-provider/src/lib.rs
```diff
@@ -71,9 +73,20 @@ impl Provider for WasiProvider {
...
         let mut container_handles = HashMap::new();
         for container in pod.containers() {
+            let mut c = Client::default();
+
+            let image = container
+                .image
+                .clone()
+                .expect("Container must have an image");
+            let image = Reference::try_from(image).unwrap();
+            let store = FileModuleStore::new(&PathBuf::from("/home/rylevick"));
```
bacongobbler: should this reference a user's homedir rather than your home? :)

examples/greet.yaml
thomastaylor312: I would prefer a wasi and wascc example

crates/wasi-provider/src/lib.rs
technosophos: Good catch. Is this about the time when we define an env var for "KRUSTLET_HOME" or something

thomastaylor312: Should the store be a data member of the `WasiProvider` struct so we don't need to initialize it on every loop?

rylev: Yes. This should have been marked as WIP.

PR #99, opened 2020-03-23
feat(*): Adds configurable data path

thomastaylor312:
Allows a user to specify what directory logs and modules should be downloaded to

### DIFF
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -17,6 +17,7 @@ pub struct Config {
     pub hostname: String,
     pub node_name: String,
     pub server_config: ServerConfig,
+    pub data_dir: PathBuf,
 }
 
 #[derive(Clone, Debug)]
@@ -38,6 +39,7 @@ impl Config {
             node_ip: default_node_ip(&mut hostname.clone(), preferred_ip_family)?,
             node_name: sanitize_hostname(&hostname),
             hostname,
+            data_dir: default_dir()?,
             server_config: ServerConfig {
                 addr: match preferred_ip_family {
                     // Just unwrap these because they are programmer error if they
@@ -77,10 +79,14 @@ impl Config {
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
         let pfx_password = opts.pfx_password.unwrap_or_default();
+        let data_dir = opts
+            .data_dir
+            .unwrap_or_else(|| default_dir().expect("unable to get default directory"));
         Config {
             node_ip,
             node_name,
             hostname,
+            data_dir,
             server_config: ServerConfig {
                 addr,
                 port,
@@ -152,12 +158,25 @@ pub struct Opts {
         help = "The name for this node in Kubernetes, defaults to the hostname of this machine"
     )]
     node_name: Option<String>,
+
+    #[clap(
+        long = "data-dir",
+        env = "KRUSTLET_DATA_DIR",
+        help = "The data path (logs, container images, etc) for krustlet storage. Defaults to $HOME/.krustlet"
+    )]
+    data_dir: Option<PathBuf>,
 }
 
 fn default_hostname() -> anyhow::Result<String> {
     Ok(hostname::get()?
         .into_string()
         .map_err(|_| anyhow::anyhow!("invalid utf-8 hostname string"))?)
+}
+
+fn default_dir() -> anyhow::Result<PathBuf> {
+    Ok(dirs::home_dir()
+        .ok_or_else(|| anyhow::anyhow!("Unable to get home directory"))?
+        .join(".krustlet"))
 }
 
 // Some hostnames (particularly local ones) can have uppercase letters, which is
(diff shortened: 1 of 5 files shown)


crates/wasi-provider/src/lib.rs
```diff
@@ -18,22 +18,35 @@ use handle::PodHandle;
...
+    /// (including creating it if necessary)
+    pub fn new<P: AsRef<Path>>(data_dir: P) -> anyhow::Result<Self> {
+        // Make sure we have a log dir and containers dir created
+        let data_path = data_dir.as_ref().to_path_buf();
+        // This is temporary as we should probably be passing in a ModuleStore
+        // as a parameter or the oci client as a whole
+        // NOTE: We do not have to create the dir here as the FileModuleStore already does this
+        let container_path = data_path.join("containers");
+        let log_path = data_path.join(LOG_DIR_NAME);
+        std::fs::create_dir_all(&log_path)?;
```
rylev: We should use `tokio::fs::create_dir_all` and make the function `async`

crates/wasi-provider/src/lib.rs
```diff
@@ -18,22 +18,35 @@ use handle::PodHandle;
...
 
 impl WasiProvider {
-    pub fn new() -> Self {
-        let store = FileModuleStore::new(&dirs::home_dir().expect("Cannot get home directory"));
-        Self {
+    /// Returns a new WASI provider configured to use the proper data directory
+    /// (including creating it if necessary)
+    pub fn new<P: AsRef<Path>>(data_dir: P) -> anyhow::Result<Self> {
+        // Make sure we have a log dir and containers dir created
+        let data_path = data_dir.as_ref().to_path_buf();
```
rylev: We should probably just take a `PathBuf` if we're just going to turn it into one anyway

crates/kubelet/src/config.rs
```diff
@@ -160,6 +173,12 @@ fn default_hostname() -> anyhow::Result<String> {
         .map_err(|_| anyhow::anyhow!("invalid utf-8 hostname string"))?)
 }
 
+fn default_dir() -> anyhow::Result<PathBuf> {
+    Ok(dirs::home_dir()
```
rylev: Can we make `dirs` only available when the `cli` feature is on?

crates/kubelet/src/config.rs
```diff
@@ -160,6 +173,12 @@ fn default_hostname() -> anyhow::Result<String> {
         .map_err(|_| anyhow::anyhow!("invalid utf-8 hostname string"))?)
 }
 
+fn default_dir() -> anyhow::Result<PathBuf> {
```
rylev: If this is only used for `data_dir` then this should probably be `default_data_dir`

crates/wasi-provider/src/lib.rs
```diff
@@ -18,22 +18,35 @@ use handle::PodHandle;
...
+        // This is temporary as we should probably be passing in a ModuleStore
+        // as a parameter or the oci client as a whole
+        // NOTE: We do not have to create the dir here as the FileModuleStore already does this
+        let container_path = data_path.join("containers");
```
rylev: Can we put `"containers"` in a const?

crates/wasi-provider/src/lib.rs
thomastaylor312: I actually did that first and it made no sense as the provider should be created before passing it in anyway.

However, with that said, because `main` is async, is it safer to just do a Tokyo call here?

crates/wasi-provider/src/lib.rs
thomastaylor312: I didn't on purpose because this is temporary. If you feel strongly about it, I'll add it in

crates/wasi-provider/src/lib.rs
thomastaylor312: Ha, I predicted this comment would come 😄. So my reasoning here is that it makes this clean for testing or hard coded paths (when someone just wants to use a `&str`) and still allows us to use the `PathBuf` from the config. My goal was to make this as easy as possible for users (e.g. just passing in anything that implements `AsRef<Path>`) who use this crate.

crates/wasi-provider/src/lib.rs
rylev: Basically anything that's being executed in an async context should be an async call.

thomastaylor312: I guess the other consideration here is around whether or not users should have to be in an async block to create this provider and my initial feeling is no

crates/wasi-provider/src/lib.rs
rylev: That's fine.  If we want ultimate flexibility this is ok.

crates/kubelet/src/config.rs
rylev: Ok then we can't do that

crates/wasi-provider/src/lib.rs
rylev: Nope. Shall we create an issue?

crates/wasi-provider/src/lib.rs
rylev: But we use it in an async block currently so at that site when we need to call spawn_blocking

crates/wasi-provider/src/lib.rs
thomastaylor312: Yeah, I'll create one as an output of the discussion we have today offline

PR #106, opened 2020-03-24
feat(demos): Adds wasi rust example

### LINKED ISSUE #103: Create wasi demo in Rust
This example should have the code we use to compile the module and a pod spec that loads user specified env vars, and env vars from the downward API/config maps/secrets. These env vars should be printed out and then retrieved via kubectl logs

thomastaylor312:
This adds the demo directory with a simple rust example with environment
variables. I also removed a directory change that was no longer needed
in the Justfile

Closes #103


demos/wasi/hello-world-rust/README.md
```diff
@@ -0,0 +1,47 @@
+# Hello World Rust for WASI
+A simple hello world example in Rust that will print the environment variables
```
technosophos: nit: This sentence is a little confusing. Maybe break into a bullet list?

thomastaylor312: Good point. Breaking out

PR #108, opened 2020-03-24
fix the release workflow

bacongobbler:
This commit fixes two things:

We want the release workflow to start when the tag is pushed rather than when it is published on github. That way, the release assets are ready and available when the release is published.

It turns out that you cannot have two conditional jobs with the same ID. Go figure. Instead, we now inject the release version into the environment and run the step if the reference is valid.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,9 +4,8 @@ on:
   push:
     branches:
       - master
-  release:
-    types:
-      - published
+    tags:
+      - 'v*'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -4,9 +4,8 @@ on:
   push:
     branches:
       - master
-  release:
-    types:
-      - published
+    tags:
+      - '*'
```
thomastaylor312: Should we make this `v*` in a half-hearted attempt to match semver tags? I don't know of another way to do it right now

bacongobbler: I think it'd be best to keep this as is for now. I don't imagine we'd be publishing any tags other than of the form `v1.2.3`, and this would keep our options open should we decide to change our version scheme at some point.

bacongobbler: meh, went ahead with the change

PR #113, opened 2020-03-25
feat(demos): Adds WASI demo written in C

### LINKED ISSUE #104: Create wasi demo in C or C++
This example should have the code we use to compile the module and a pod spec that loads user specified env vars, and env vars from the downward API/config maps/secrets. These env vars should be printed out and then retrieved via kubectl logs

thomastaylor312:
Also contains a tiny fix in the webassembly README

Closes #104

### DIFF
--- a/demos/wasi/hello-world-c/README.md
+++ b/demos/wasi/hello-world-c/README.md
@@ -1,5 +1,5 @@
-# Hello World AssemblyScript for WASI
-A simple hello world example in AssemblyScript that will print:
+# Hello World C for WASI
+A simple hello world example in C that will print:
 - The environment variables available to the process
 - Text to both stdout and stderr.
 - Any args passed to the process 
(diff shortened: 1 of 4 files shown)


demos/wasi/hello-world-c/README.md
```diff
@@ -0,0 +1,57 @@
...
+$ wget https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-8/wasi-sdk-8.0-${OS_NAME}.tar.gz
+$ tar -xzf wasi-sdk-8.0-${OS_NAME}.tar.gz
+```
+
+If you don't have Krustlet with the WASI provider running locally, see the
+instructions in the [tutorial](../../../docs/intro/tutorial03.md) for running
+locally.
+
+## Building
+Simply run:
```
radu-matei: Personally I would avoid "simply".

thomastaylor312: In this case I agree

PR #115, opened 2020-03-26
Abstract over module store and client

### LINKED ISSUE #95: Use non-hardcoded modules in wascc-provider
We are currently using a [hard coded WASM module](https://github.com/deislabs/krustlet/blob/master/crates/wascc-provider/src/lib.rs#L103) in the wascc-provider. This should be replaced with a module store that fetches the module if it's not there and returns the module from a cache.

rylev:
This abstracts over which store and image client is used giving the user of the library control over how they store modules (with a convenience for a file based store which we expect to be the most common) and how modules are fetched (with only support for OCI currently implemented). 

Fixes #95


crates/kubelet/src/module_store.rs
```diff
@@ -0,0 +1,88 @@
...
+            let contents = self.client.lock().await.pull(image_ref).await?;
+            self.store(image_ref, &contents).await?;
+            return Ok(contents);
+        }
+
+        Ok(tokio::fs::read(path).await?)
+    }
+}
+
+impl<C> Clone for FileModuleStore<C> {
```
thomastaylor312: Can this not be automatically derived? Or is the fact that C is generic what is giving problems?

crates/oci-distribution/src/lib.rs
```diff
@@ -3,23 +3,18 @@ use chrono::prelude::{DateTime, Utc};
 use futures_util::future;
 use futures_util::stream::StreamExt;
 use hyperx::header::Header;
+use kubelet::{ImageClient, Reference};
```
thomastaylor312: Because we want this to eventually be a fully featured OCI client, I don't think we want to have a dependency on Kubelet in here. `Reference` should most likely be in this crate instead. I'm still thinking through how we'd want to structure things for `ImageClient` though

crates/kubelet/src/module_store.rs
rylev: Yea, `#[derive(Clone)]` automatically puts a `: Clone` on generic params which we don't want.

crates/wascc-provider/src/lib.rs
```diff
@@ -25,10 +26,20 @@ type EnvVars = std::collections::HashMap<String, String>;
 /// TODO: In the future, we will look at loading capabilities using the "sidecar" metaphor
 /// from Kubernetes.
 #[derive(Clone)]
-pub struct WasccProvider {}
+pub struct WasccProvider<S> {
+    store: S,
+}
+
+impl<S: ModuleStore + Send + Sync> WasccProvider<S> {
+    /// Returns a new wasCC provider configured to use the proper data directory
```
thomastaylor312: Note: This is not valid here now, but we will need it anyway when we implement logging

PR #117, opened 2020-03-26
Simplfy imports

### LINKED ISSUE #116: AssemblyScript example doesn't have to use relative import path
This:
https://github.com/deislabs/krustlet/blob/1c945e8af2f661d747184f4ccb934df30ea19421/demos/wasi/hello-world-assemblyscript/assembly/index.ts#L1

Can just use `as-wasi`

willemneal:
fixes #116


willemneal:
No problem glad to see you're using AssemblyScript!

PR #118, opened 2020-03-27
Improve docs

rylev:
Screenshot 2020-03-27 at 13 49 56


crates/kubelet/src/provider.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    /// Resolve the environment variables for a container.
+    ///
+    /// This generally should not be overwritten unless you need to handle
+    /// environment variable resolution in a special way, such as allowing
+    /// custom Downward API fields.
+    ///
+    /// It is safe to call from within your own providers.
+    // TODO: Finish secrets, configmaps, and resource fields
```
thomastaylor312: I think this TODO is no longer needed

crates/kubelet/src/status.rs
```diff
@@ -0,0 +1,122 @@
...
+use k8s_openapi::api::core::v1::{
+    ContainerState, ContainerStateRunning, ContainerStateTerminated, ContainerStateWaiting,
+};
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
+
+use std::collections::HashMap;
+
+/// Describe the status of a workload.
+///
+/// Phase captures the lifecycle aspect of the workload, while
```
thomastaylor312: This comment is no longer valid as we don't capture phase directly. It should be updated to reflect that container statuses is part of it

crates/oci-distribution/src/client.rs
```diff
@@ -0,0 +1,394 @@
+//! OCI distribution client
```
thomastaylor312: We should probably add a note that this is not a full implementation yet, but we want it to be

crates/oci-distribution/src/manifest.rs
```diff
@@ -1,3 +1,4 @@
+//! Oci Manifest
```
thomastaylor312: Nit: Capitalize OCI

crates/kubelet/src/module_store.rs
```diff
@@ -8,13 +11,47 @@ use std::convert::TryFrom;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+/// A store of container modules.
+///
+/// This provides the ability to get a module's bytes given an image `Reference`.
```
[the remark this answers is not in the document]
rylev: Yep, will do

PR #119, opened 2020-03-27
Change `arch` to be an associated const

rylev:


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Build and Test
 
-on: [push, pull_request]
```
radu-matei: Good catch!

PR #120, opened 2020-03-30
No client in provider

rylev:
There's no need to pass a client to every Provider method. If the implementing type needs a client they can provide it themselves.

### DIFF
--- a/src/krustlet-wascc.rs
+++ b/src/krustlet-wascc.rs
@@ -25,7 +25,7 @@ async fn main() -> anyhow::Result<()> {
     module_store_path.push("modules");
     let store = FileModuleStore::new(client, &module_store_path);
 
-    let provider = WasccProvider::new(store, &config).await?;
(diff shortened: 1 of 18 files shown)


src/krustlet-wascc.rs
```diff
@@ -25,7 +25,7 @@ async fn main() -> anyhow::Result<()> {
     module_store_path.push("modules");
     let store = FileModuleStore::new(client, &module_store_path);
 
-    let provider = WasccProvider::new(store, &config).await?;
```
rylev: @thomastaylor312 should the Krustlet config object just contain a kube config?

thomastaylor312: So the one problem with moving the config loading logic to the config (unless we make it a lazy load) is that you end up requiring a kubeconfig to even show the `--help` text. Because that code would be run while parsing the flags and returning the config

thomastaylor312:
Merging now per @rylev's request

PR #128, opened 2020-03-30
feat(docs): Adds how-to guides with complete AKS guide

### LINKED ISSUE #88: Create instructions for running a Krustlet node with AKS
We should have detailed instructions for configuring and using Krustlet as an AKS node. This first set of instructions does not need to be a fully production ready node (as the project itself isn't ready for full production yet either). Learning how to do it with AKS should make it easy for us to write instructions for other Kubernetes providers

thomastaylor312:
This also stubs out the rest of this documentation and includes a tiny
bug fix to make sure flags load before the kubeconfig in binaries.

Closes #88

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,3 +9,4 @@ Krustlet has a lot of documentation. A high-level overview of how it’s organiz
 - [The introduction](intro/README.md) takes you by the hand through a series of steps to run an application on Krustlet. Start here if you’re new to Krustlet.
 - [Topic guides](topics/README.md) discuss key topics and concepts at a fairly high level and provide useful background information and explanation.
 - [Community Guides](community/README.md) teach you about the development process for Krustlet and how you can contribute to the project.
+- [How To Guides](how-to/README.md) explain how to run Krustlet with various types of Kubernetes clusters.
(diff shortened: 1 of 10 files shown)


docs/README.md
```diff
@@ -9,3 +9,4 @@ Krustlet has a lot of documentation. A high-level overview of how it’s organiz
 - [The introduction](intro/README.md) takes you by the hand through a series of steps to run an application on Krustlet. Start here if you’re new to Krustlet.
 - [Topic guides](topics/README.md) discuss key topics and concepts at a fairly high level and provide useful background information and explanation.
 - [Community Guides](community/README.md) teach you about the development process for Krustlet and how you can contribute to the project.
+- [How To Guides](how-to/README.md) explain how to run Krustlet with various types of Kubernetes clusters.
```
bacongobbler: I can imagine us writing other HOWTOs unrelated to various Kubernetes cluster setups. How to mount volumes into your WASI pod, running with multiple krustlets, etc.

docs/how-to/aks.md
```diff
@@ -0,0 +1,247 @@
...
+docs](https://docs.microsoft.com/en-us/azure/aks/tutorial-kubernetes-deploy-cluster)
+for more information. Make sure you have the credentials to your cluster
+downloaded by running:
+
+```shell
+$ az aks get-credentials --name $CLUSTER_NAME --resource-group $RESOURCE_GROUP
+```
+
+This specific tutorial will be running krustlet on another Azure VM; however,
+you can follow these steps from any device that can access the internet.
```
radu-matei: Not sure if this is quite accurate - the current setup requires a _device that can start a web server on an IP accessible from the Kubernetes control plane_ rather than _a device that can access the internet_, right?

docs/README.md
thomastaylor312: Fixed!

docs/how-to/aks.md
thomastaylor312: You are totally right. This has been fixed

PR #134, opened 2020-03-31
chore(*): Removes dockerfile, as we no longer use it

thomastaylor312:


thomastaylor312:
Oh, I forgot about the docs. Fixing

PR #135, opened 2020-03-31
fix(wasi): Removes unimplemented panics

### LINKED ISSUE #129: Remove panic from wasi stop
Although we cannot actually stop a running wasi runtime, we can drop the handle from the map and _not_ panic so we don't stop the whole thing from running

thomastaylor312:
These panics were causing the process to crash when a pod was deleted.
We now just log when we attempt to delete a pod and let the pod be
deleted from the API

Closes #129


bacongobbler:
seems sane

PR #137, opened 2020-03-31
ref(howto): formatting

bacongobbler:


bacongobbler:
merging since this follows the same line length as the rest of the docs. Will follow up later with a change to 100 characters per line across the board. It looks like github PRs word wraps at around 115 characters, resulting in some words wrapping on another line.

PR #139, opened 2020-04-01
ref(handle): Moves handle to kubelet crate and makes it fully generic

thomastaylor312:
This adds a new `Stop` trait that can be implemented per provider while still preserving
all the pod and container management goodness of the handles. Yes, there are
now 2 generic types on the handle structs, but after discussing it with @rylev
we are ok with this as a middleground for now. This should enable any provider to
take advantage of these handles.

Please note that I removed the test for the wasi runtime because with the new
implementation, we can't wait on things to complete right now. Also, the same
functionality is being covered in the e2e tests

### DIFF
--- a/crates/wasi-provider/src/wasi_runtime.rs
+++ b/crates/wasi-provider/src/wasi_runtime.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 use std::sync::Arc;
 
 use anyhow::bail;
-use log::{error, info};
+use log::{error, info, warn};
 use tempfile::NamedTempFile;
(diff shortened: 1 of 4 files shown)


crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -253,41 +272,3 @@ impl WasiRuntime {
         })
     }
 }
-
-#[cfg(test)]
-mod test {
-    use super::*;
-
-    #[tokio::test]
-    async fn test_run() {
```
bacongobbler, line 262: is there somewhere else this test is being run?

PR #148, opened 2020-04-01
Update kind instruction for Mac

jingweno:
For Mac, `kubectl logs` accesses the Krustlet process running on the host.
The default gateway is the en0 interface IP on the host.

Discussion: https://bytecodealliance.zulipchat.com/#narrow/stream/227959-wasmtime-on.20k8s/topic/Krustlet/near/192599630
Ref 1: https://github.com/moby/moby/issues/22753#issuecomment-234078631
Ref 2: https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds

### DIFF
--- a/docs/howto/krustlet-on-kind.md
+++ b/docs/howto/krustlet-on-kind.md
@@ -76,6 +76,9 @@ We can use this IP address from the guest Operating System (the KinD host) to co
 Operating System (where Krustlet is running). If this was changed, check `ip addr show docker0` from
 the host OS to determine the default gateway.
 
+For Mac, if you are running Krustlet on the host, the default gateway should be the `en0` interface IP.
+You can get it with `ifconfig en0 | grep "inet " | cut -d " " -f2`
+
 ## Step 3: Install and run Krustlet
 
 First, install the latest release of Krustlet following [the install guide](../intro/install.md).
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hey @jingweno! I grabbed my laptop and started looking into this issue further. I wrote up my findings over in #147, which looks to cover the documentation provided here with additional context.

I found the following:

- Because a container can reach the `en0` network, you should not need to provide a `--node-ip`, as Krustlet will start listening on the 192.168.1.* address by default
- I attached some links that should help the next user caught up in this mess.

Would it be okay if we went ahead with #147 instead? I feel that PR provides a little more information for the user to dig into this issue further and hopefully equip them with more tools to dive into the deep end with this.

Thanks again!

jingweno:
@bacongobbler Yes, feel free to close this one. 

> Because a container can reach the en0 network, you should not need to provide a --node-ip, as Krustlet will start listening on the 192.168.1.* address by default

This doesn't seem to always work for me. As discussed in https://bytecodealliance.zulipchat.com/#narrow/stream/227959-wasmtime-on.20k8s/topic/Krustlet/near/192597480, sometimes I got:

```
./krustlet-wasi --pfx-password password
thread 'main' panicked at 'unable to get default node IP address: failed to lookup address information: nodename nor servname provided, or not known', crates/kubelet/src/config.rs:92:13
```

bacongobbler:
Thanks for reminding me about that particular case. I'll make sure to call that out.

PR #151, opened 2020-04-02
fix(*): Adds node tainting to control pod scheduling

### LINKED ISSUE #126: Add taints to krustlet nodes
Right now, normal pods can get scheduled on krustlet nodes by the API, this is likely due to there being no taints that describe it as a wasi or wascc node. We should add a taint and update demos to specify a toleration for that taint

thomastaylor312:
This change forces pods to add a toleration for the specified architecture
before scheduling. Other normal pods were getting scheduled to the attached
krustlet nodes and ever starting before this.

I also found that we were listing _all_ pods instead of pods scoped to the
node, so now the informer has a field selector to only get pods that are
assigned to the node.

With this change, we no longer have a need for the `can_schedule` method, so
it has been removed from the provider trait

NOTE: I have also included a bugfix for getting nodes to go back to ready
on restart. It is hacky, but sufficient for now

Closes #126

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -65,10 +65,14 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
 
         // This informer listens for pod events.
         let provider = self.provider.clone();
-
+        let node_name = self.config.node_name.clone();
         let pod_informer = tokio::task::spawn(async move {
             // Create our informer and start listening.
-            let informer = Informer::new(client, ListParams::default(), Resource::all::<KubePod>());
+            let params = ListParams {
+                field_selector: Some(format!("spec.nodeName={}", node_name)),
+                ..Default::default()
+            };
+            let informer = Informer::new(client, params, Resource::all::<KubePod>());
             loop {
                 let mut stream = informer.poll().await.expect("informer poll failed").boxed();
                 while let Some(event) = stream.try_next().await.unwrap() {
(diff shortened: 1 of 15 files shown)


crates/kubelet/src/kubelet.rs
```diff
@@ -65,10 +65,14 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
 
         // This informer listens for pod events.
         let provider = self.provider.clone();
-
+        let node_name = self.config.node_name.clone();
```
rylev, line 68: nit: you can avoid an extra allocation if you do the formatting (with `format!`) here.

`let field_selector = format!("spec.nodeName={}", self.config.node_name)`

crates/kubelet/src/node.rs
```diff
@@ -38,11 +38,28 @@ pub async fn create_node(client: &kube::Client, config: &Config, arch: &str) {
...
+                    // is a hacky workaround for now where we delete and
+                    // recreate the node. This is being tracked in https://github.com/deislabs/krustlet/issues/150
                     info!("node found, updating current node definition");
-                    update_node(
-                        client,
-                        &node.metadata.unwrap_or_default().name.unwrap_or_default(),
+                    node_client
+                        .delete(&config.node_name, &DeleteParams::default())
+                        .await
+                        .expect("Unable to recreate node...aborting");
```
rylev, line 50: Should this function return a `Result` instead of panicing on these? This means we'll crash the program on network failure.

thomastaylor312: This should panic here because it wasn't able to re-establish control over the node, leaving things in a completely unknown state

rylev: Hmm.. this is the first write action we perform in the error recovery arm. Shouldn't the caller be able to recover?

thomastaylor312: The proper thing here would be a retry. I'll add an issue to 0.2 for it

thomastaylor312: Created #155

PR #169, opened 2020-04-08
Update dependencies and move from new clap to struct opt

rylev:
This should allow us to publish the crates to crates.io


crates/oci-distribution/src/client.rs
```diff
@@ -261,8 +260,6 @@ impl ClientProtocol {
 #[derive(serde::Deserialize, Default)]
 struct RegistryToken {
     access_token: String,
-    expires_in: Option<u32>,
```
thomastaylor312, line 264: I know these are unused now, but I think we will need them later. That correct @radu-matei?

crates/oci-distribution/Cargo.toml
```diff
@@ -13,12 +13,11 @@ authors = [
...
+reqwest = { version = "0.10", features = ["json", "stream"] }
 anyhow = "1.0"
-tokio = {version  = "0.2.14", features = ["macros", "fs"] }
+tokio = {version  = "0.2", features = ["macros", "fs"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 www-authenticate = "0.3.0"
-hyperx = "0.13"
-chrono = { version = "0.4", features = ["serde"] }
+hyperx = "0.13.0"
```
thomastaylor312, line 22: Why does this one have a fully qualified version (with patch number), but the others don't?

crates/wascc-provider/src/lib.rs
```diff
@@ -29,26 +29,26 @@
 //! }
 //! ```
 
-#![warn(missing_docs)]
+#![deny(missing_docs)]
```
thomastaylor312, line 32: Putting down the hammer? 😆

crates/oci-distribution/src/client.rs
rylev: I originally added `[allow(dead_code)]` because I figured we would need them eventually but then I just removed them because adding them back should be easy.

crates/oci-distribution/Cargo.toml
rylev: Whoops. Will fix

crates/oci-distribution/src/client.rs
thomastaylor312: Ok, I think that is fine then

PR #173, opened 2020-04-09
Retry actions on nodes

### LINKED ISSUE #155: Add retries to node create/update logic
Right now, if we fail to take ownership of a node, we panic and exit. It would be much more graceful and user friendly if we instead did things in a retry loop for a while before exiting.

rylev:
Possibly fixes #155

### DIFF
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -8,7 +8,7 @@ use kube::Error;
 use log::{debug, error, info};
 
 macro_rules! retry {
-    ($action:expr, times: $num_times:expr, $on_err:expr) => {{
+    ($action:expr, times: $num_times:expr, error: $on_err:expr) => {{
         let mut n = 0u8;
         let mut duration = std::time::Duration::from_millis(100);
         loop {
@@ -17,7 +17,9 @@ macro_rules! retry {
             match result {
                 Ok(_) => break result,
                 Err(ref e) => {
-                    $on_err(e);
+                    if $on_err(e, n) {
+                        break result;
+                    };
                     tokio::time::delay_for(duration).await;
                     duration *= (n + 1) as u32;
                     if n == $num_times {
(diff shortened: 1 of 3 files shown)


crates/kubelet/src/node.rs
```diff
@@ -4,8 +4,34 @@ use k8s_openapi::api::coordination::v1::Lease;
 use k8s_openapi::api::core::v1::Node;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::{Api, DeleteParams, PatchParams, PostParams};
+use kube::Error;
 use log::{debug, error, info};
 
+macro_rules! retry {
+    ($action:expr, times: $num_times:expr, $on_err:expr) => {{
+        let mut n = 0u8;
+        let mut duration = std::time::Duration::from_millis(100);
```
thomastaylor312, line 13: For maximum flexibility, this should probably be a configurable value. Not a blocking request though

crates/kubelet/src/node.rs
```diff
@@ -4,8 +4,34 @@ use k8s_openapi::api::coordination::v1::Lease;
 use k8s_openapi::api::core::v1::Node;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::{Api, DeleteParams, PatchParams, PostParams};
+use kube::Error;
 use log::{debug, error, info};
 
+macro_rules! retry {
+    ($action:expr, times: $num_times:expr, $on_err:expr) => {{
+        let mut n = 0u8;
```
thomastaylor312, line 12: For my own edification, why is this specified this way instead of just `1`?

crates/kubelet/src/node.rs
```diff
@@ -4,8 +4,34 @@ use k8s_openapi::api::coordination::v1::Lease;
...
+        let mut duration = std::time::Duration::from_millis(100);
+        loop {
+            n += 1;
+            let result = $action;
+            match result {
+                Ok(_) => break result,
+                Err(ref e) => {
+                    $on_err(e);
+                    tokio::time::delay_for(duration).await;
+                    duration *= (n + 1) as u32;
```
thomastaylor312, line 22: Nit: Maybe we should just stick to a linear retry and then refactor later to use something like an exponential backoff with jitter because this feels like it is weirdly in between

crates/kubelet/src/node.rs
rylev: Because we increment `n` at the beginning of the loop instead of the end. I can refactor it to start at 1 if you want.

crates/kubelet/src/node.rs
rylev: Why is linear retry preferred to exponential without jitter?

crates/kubelet/src/node.rs
thomastaylor312: It doesn't matter too much to me, I was just curious on the reasoning

crates/kubelet/src/node.rs
thomastaylor312: I just realized this was a `*=` and not a `+=` so never mind me over here

PR #175, opened 2020-04-09
Added Krustlet on WSL2 instructions

nunix:
Instructions for adding Krustlet as a node to Docker Desktop for Windows Kubernetes cluster, and using WSL2 as docker daemon backend.

I copied the KinD howto, so parts of the instructions will be identical.


docs/howto/krustlet-on-wsl2.md
```diff
@@ -0,0 +1,148 @@
...
+Last but not least, this will work on Windows 10 Home edition.
+
+## Prerequisites
+
+You will require a WSL2 distro and Docker Desktop for Windows for this how-to. The WSL2 backend and Kubernetes features will need to be also enabled.
+See the [Docker Desktop for Windows > Getting started > Kubernetes](https://docs.docker.com/docker-for-windows/#kubernetes) howto for more information.
+
+This specific tutorial will be running Krustlet on your WSL2 distro and will explain how to access it from Windows.
+
+## Step 1: Create Certificate
```
bketelsen, line 19: This entire step is not necessary.

instead, type:
`kubectl config get-contexts`

then `kubectl config set-context docker-desktop`

bketelsen: this may be dependent on enabling WSL integration in docker> settings > resources > WSL Integration

docs/howto/krustlet-on-wsl2.md
```diff
@@ -0,0 +1,148 @@
...
+        RX errors 0  dropped 0  overruns 0  frame 0
+        TX packets 1093  bytes 115724 (113.0 KiB)
+        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
+```
+
+In this example, I should use `172.26.47.208`.
+
+The hostname being "applied" from Windows, the default hostname will not resolve to this address, therefore you need to pass the `--node-ip` and `--node-name` flag to Krustlet.
+
+### Add a route to 172.17.0.1
```
bketelsen, line 101: adding a route isn't necessary

docs/howto/krustlet-on-wsl2.md
```diff
@@ -0,0 +1,148 @@
...
+**DO NOT** make this route permanent as the WSL2 default gateway is DHCP based and will change upon every reboot.
+
+## Step 3: Install and run Krustlet
+
+First, install the latest release of Krustlet following [the install guide](../intro/install.md).
+
+Once you have done that, run the following commands to run Krustlet's WASI provider:
+
+```shell
+$ krustlet-wasi --node-ip 172.17.0.1 --node-name krustlet --pfx-password password
```
bketelsen, line 125: instead of the gateway IP, use the actual IP address in wsl `172.26.47.208`.
Also the pfx password isn't required anymore if you use the context from above.
so it can be simplified to:
`KRUSTLET_NODE_IP=172.26.47.208 just run-wasi`

docs/howto/krustlet-on-wsl2.md
nunix: oh ok, nice, I kept it form the KinD instructions as it was "approximately" the same.
And if we are only using Docker Desktop (`docker` and `kubectl` will be "mounted" from the docker desktop distro), then the context is by default already docker-desktop

nunix: what I can/will do is set it as optional then

docs/howto/krustlet-on-wsl2.md
nunix: the route is optional indeed, like I stated, it was in case we want to have a "linux feeling".
I still have to run the demo, but I guess this means you rely on Docker Desktop doing all the ports forwarding and using `localhost` only?

docs/howto/krustlet-on-wsl2.md
nunix: thank you, I will remove the password option and will amend the `--node-ip`

PR #188, opened 2020-04-15
fixed: now krustlet can pull wasm image from standard registry

fleeto:
I'm trying to run a pod with an image on my own registry, but when the pod was created, I saw that in my registry's log: `OCI manifest found, but accept header does not support OCI manifests"}]}`

Appended "application/vnd.oci.image.manifest.v1+json" in the `accept` header, it works now.


fleeto:
> That's a great catch, thanks!
> 
> (I am also curious what registry you are using - I assume built on top of Distribution?)

registry:2 :)

PR #189, opened 2020-04-15
Update kube to 0.32

rylev:


crates/kubelet/src/kubelet.rs
```diff
@@ -125,12 +126,9 @@ mod test {
...
 
     fn mock_client() -> kube::Client {
-        kube::config::Configuration {
-            base_path: ".".to_string(),
-            client: reqwest::Client::new(),
-            default_ns: " ".to_string(),
-        }
-        .into()
+        kube::Client::new(kube::Config::new(
+            reqwest::Url::parse("http://example.com").unwrap(),
```
bacongobbler, line 130: slight nitpick: even though this is a mock test, in order to absolutely prevent any chance of reaching that URL, can we change this to `http://127.0.0.1:8080`, or is that not possible?

tests/integration_tests.rs
```diff
@@ -90,12 +84,11 @@ async fn test_wascc_provider() -> Result<(), Box<dyn std::error::Error>> {
 
     assert_eq!(pod.status.unwrap().phase.unwrap(), "Pending");
 
-    let inf: Informer<Pod> = Informer::new(
-        client,
+    let api = Api::namespaced(client, "default");
+    let inf: Informer<Pod> = Informer::new(api).params(
         ListParams::default()
-            .fields("metadata.name=greet-wascc")
+            .fields("metadata.name=hello-wasi")
```
bacongobbler, line 90: was this accidentally changed?

tests/integration_tests.rs
```diff
@@ -118,6 +111,9 @@ async fn test_wascc_provider() -> Result<(), Box<dyn std::error::Error>> {
 
     assert!(went_ready, "pod never went ready");
 
+    // Allow some time for the pod to actually become ready
+    tokio::time::delay_for(std::time::Duration::from_secs(2)).await;
```
thomastaylor312, line 115: This is what the watcher loop is supposed to be doing. It is not actually ready when it hits running?

PR #195, opened 2020-04-16
feat(*): Adds volume support to wasi

### LINKED ISSUE #180: Implement secret volume mounting
A user should be able to specify a Kubernetes `Secret` to mount into a running instance with both the wascc and wasi providers. We don't need to worry about linux file permissions as they don't matter in a non-container context.

### LINKED ISSUE #181: Implement config map volume mounting
A user should be able to specify a Kubernetes `ConfigMap` to mount into a running instance with both the wascc and wasi providers. We don't need to worry about linux file permissions as they don't matter in a non-container context.

thomastaylor312:
Closes #180 
Closes #181

### DIFF
--- a/crates/kubelet/src/handle.rs
+++ b/crates/kubelet/src/handle.rs
@@ -16,6 +16,7 @@ use tokio::task::JoinHandle;
 
 use crate::provider::ProviderError;
 use crate::status::{ContainerStatus, Status};
+use crate::volumes::VolumeRef;
 use crate::Pod;
 
 /// Any provider wanting to use the [`RuntimeHandle`] and
@@ -57,19 +58,19 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
         }
     }
 
+    /// Signal the running instance to stop. Use [`RuntimeHandle::wait`] to wait for the process to exit. This
+    /// uses the underlying [`Stop`] implementation passed to the
+    /// constructor
+    pub async fn stop(&mut self) -> anyhow::Result<()> {
+        self.stopper.stop().await
+    }
+
     /// Write all of the output from the running process into the given buffer.
     /// Returns the number of bytes written to the buffer
     pub(crate) async fn output(&mut self, buf: &mut Vec<u8>) -> anyhow::Result<usize> {
         let bytes_written = self.output.read_to_end(buf).await?;
         self.output.seek(SeekFrom::Start(0)).await?;
         Ok(bytes_written)
-    }
-
-    /// Signal the running instance to stop. Use [`RuntimeHandle::wait`] to wait for the process to exit. This
-    /// uses the underlying [`Stop`] implementation passed to the
-    /// constructor
-    pub async fn stop(&mut self) -> anyhow::Result<()> {
-        self.stopper.stop().await
     }
 
     /// Returns a clone of the status_channel for use in reporting the status to
@@ -93,17 +94,22 @@ pub struct PodHandle<S, R> {
     container_handles: RwLock<HashMap<String, RuntimeHandle<S, R>>>,
     status_handle: JoinHandle<()>,
     pod: Pod,
+    // Storage for the volume references so they don't get dropped until the runtime handle is
+    // dropped
+    _volumes: Option<HashMap<String, VolumeRef>>,
 }
 
 impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> PodHandle<S, R> {
-    /// Creates a new pod handle that manages the given map of container names
-    /// to [`RuntimeHandle`]s. The given pod and client are used to
-    /// maintain a reference to the kubernetes object and to be able to update
-    /// the status of that object
+    /// Creates a new pod handle that manages the given map of container names to
+    /// [`RuntimeHandle`]s. The given pod and client are used to maintain a reference to the
+    /// kubernetes object and to be able to update the status of that object. The optional volumes
+    /// parameter allows a caller to pass a map of volumes to keep reference to (so that they will
+    /// be dropped along with the pod)
     pub fn new(
         container_handles: HashMap<String, RuntimeHandle<S, R>>,
         pod: Pod,
         client: kube::Client,
+        volumes: Option<HashMap<String, VolumeRef>>,
     ) -> anyhow::Result<Self> {
         let mut channel_map = StreamMap::with_capacity(container_handles.len());
         for (name, handle) in container_handles.iter() {
@@ -135,6 +141,7 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> PodHandle<S, R> {
             container_handles: RwLock::new(container_handles),
             status_handle,
             pod,
+            _volumes: volumes,
         })
     }
 
--- a/crates/kubelet/src/handle.rs
+++ b/crates/kubelet/src/handle.rs
@@ -58,9 +58,8 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
         }
     }
 
-    /// Signal the running instance to stop. Use [`RuntimeHandle::wait`] to wait for the process to exit. This
-    /// uses the underlying [`Stop`] implementation passed to the
-    /// constructor
+    /// Signal the running instance to stop. Use [`RuntimeHandle::wait`] to wait for the process to
+    /// exit. This uses the underlying [`Stop`] implementation passed to the constructor
     pub async fn stop(&mut self) -> anyhow::Result<()> {
         self.stopper.stop().await
     }
@@ -96,7 +95,7 @@ pub struct PodHandle<S, R> {
     pod: Pod,
     // Storage for the volume references so they don't get dropped until the runtime handle is
     // dropped
-    _volumes: Option<HashMap<String, VolumeRef>>,
+    _volumes: HashMap<String, VolumeRef>,
 }
 
 impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> PodHandle<S, R> {
@@ -141,7 +140,7 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> PodHandle<S, R> {
             container_handles: RwLock::new(container_handles),
             status_handle,
             pod,
-            _volumes: volumes,
+            _volumes: volumes.unwrap_or_default(),
         })
     }
 
--- a/crates/kubelet/src/volumes.rs
+++ b/crates/kubelet/src/volumes.rs
@@ -6,6 +6,7 @@ use std::path::PathBuf;
 
 use k8s_openapi::api::core::v1::Volume as KubeVolume;
 use k8s_openapi::api::core::v1::{ConfigMap, Secret};
+use k8s_openapi::ByteString;
 use kube::api::Api;
 use log::{debug, error};
 
@@ -21,6 +22,7 @@ enum VolumeType {
 /// A smart wrapper around the location of a volume on the host system. If this is a ConfigMap or
 /// Secret volume, dropping this reference will clean up the temporary volume. [AsRef] and
 /// [std::ops::Deref] are implemented for this type so you can still use it like a normal PathBuf
+#[derive(Debug)]
 pub struct VolumeRef {
     host_path: PathBuf,
     volume_type: VolumeType,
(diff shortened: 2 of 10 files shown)


crates/kubelet/src/handle.rs
```diff
@@ -57,6 +58,13 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
         }
     }
 
+    /// Signal the running instance to stop. Use [`RuntimeHandle::wait`] to wait for the process to exit. This
+    /// uses the underlying [`Stop`] implementation passed to the
+    /// constructor
```
rylev, line 63: nit: move this line to the one above it.

crates/kubelet/src/handle.rs
```diff
@@ -93,17 +94,22 @@ pub struct PodHandle<S, R> {
     container_handles: RwLock<HashMap<String, RuntimeHandle<S, R>>>,
     status_handle: JoinHandle<()>,
     pod: Pod,
+    // Storage for the volume references so they don't get dropped until the runtime handle is
+    // dropped
+    _volumes: Option<HashMap<String, VolumeRef>>,
```
rylev, line 99: Is there a logical difference between `volumes == None` and `volumes == Some(HashMap::default())`?

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+enum VolumeType {
+    ConfigMap,
+    Secret,
+    HostPath,
+}
+
+/// A smart wrapper around the location of a volume on the host system. If this is a ConfigMap or
+/// Secret volume, dropping this reference will clean up the temporary volume. [AsRef] and
+/// [std::ops::Deref] are implemented for this type so you can still use it like a normal PathBuf
+pub struct VolumeRef {
```
rylev, line 24: Can we impl Debug for this?

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// This is a gnarly function to check all of the supported data members of the Volume struct.
+/// Because it isn't a HashMap, we need to check all fields individually
+async fn configure(
+    vol: &KubeVolume,
+    namespace: &str,
+    client: &kube::Client,
+    path: &PathBuf,
+) -> anyhow::Result<VolumeType> {
+    if let Some(cm) = &vol.config_map {
```
rylev, line 105: nit I think this would read better as match instead of a long `if else` chain

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+async fn populate_from_secret(
+    name: &str,
+    namespace: &str,
+    client: &kube::Client,
+    path: &PathBuf,
+) -> anyhow::Result<VolumeType> {
+    tokio::fs::create_dir(path).await?;
+    let secret_client: Api<Secret> = Api::namespaced(client.clone(), namespace);
+    let secret = secret_client.get(name).await?;
+    let data = secret.data.unwrap_or_default();
```
rylev, line 141: Do we really want the default if the data's not there?

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+    let data = data.iter().map(|(key, data)| async move {
```
rylev, line 142: Can you deconstruct the `data` arg to the closure like this `ByteString(data)` so that you don't have to call `data.0`.

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+
+async fn populate_from_config_map(
+    name: &str,
+    namespace: &str,
+    client: &kube::Client,
+    path: &PathBuf,
+) -> anyhow::Result<VolumeType> {
+    tokio::fs::create_dir(path).await?;
+    let cm_client: Api<ConfigMap> = Api::namespaced(client.clone(), namespace);
+    let cm = cm_client.get(name).await?;
```
rylev, line 162: nit: not a fan of the two letter variable name here. Unless the usage is super local (within a line or two of the declaration) I try to avoid naming things so tersely.

crates/wasi-provider/src/lib.rs
```diff
@@ -107,18 +113,48 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
 
         let mut modules = self.store.fetch_pod_modules(&pod).await?;
         let client = kube::Client::from(self.kubeconfig.clone());
+        // TODO: Make volume refs part of the handle so they don't get dropped
```
rylev, line 116: Are they not a part of the handle?

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+    let binary_data = cm.binary_data.unwrap_or_default();
```
rylev, line 163: Same as above

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+    let binary_data = binary_data.iter().map(|(key, data)| async move {
```
rylev, line 164: Same with `data` as above

crates/wasi-provider/src/lib.rs
```diff
@@ -107,18 +113,48 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
...
             let module_data = modules
                 .remove(&container.name)
                 .expect("FATAL ERROR: module map not properly populated");
+            let container_volumes: HashMap<String, Option<String>> =
+                if let Some(volume_mounts) = container.volume_mounts.as_ref() {
+                    volume_mounts
+                        .iter()
+                        .map(|vm| -> anyhow::Result<(String, Option<String>)> {
+                            // Check the volume exists first
+                            let vol = volumes.get(&vm.name).ok_or_else(|| {
```
rylev, line 130: Is it possible to move this closure to another function. This is slightly gnarly.

tests/integration_tests.rs
```diff
@@ -193,6 +193,40 @@ async fn test_wasi_provider() -> Result<(), Box<dyn std::error::Error>> {
         }
     );
 
+    let secrets: Api<Secret> = Api::namespaced(client.clone(), "default");
+    secrets
+        .create(
+            &PostParams::default(),
+            &serde_json::from_value(json!({
```
rylev, line 200: nit: we should probably move these json values somewhere else so that the test function isn't so verbose. It's become a bit hard to read.

crates/wasi-provider/src/lib.rs
```diff
@@ -107,18 +113,48 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
...
         let client = kube::Client::from(self.kubeconfig.clone());
+        // TODO: Make volume refs part of the handle so they don't get dropped
+        let volumes = VolumeRef::volumes_from_pod(&self.volume_path, &pod, &client).await?;
         info!("Starting containers for pod {:?}", pod_name);
         for container in pod.containers() {
             let env = Self::env_vars(&container, &pod, &client).await;
             let module_data = modules
                 .remove(&container.name)
                 .expect("FATAL ERROR: module map not properly populated");
+            let container_volumes: HashMap<String, Option<String>> =
```
rylev, line 124: Is there a reason we decided to make these Strings instead of `PathBufs`. It seems like a good idea to keep these `PathBuf`s for as long as we can

crates/kubelet/src/volumes.rs
bacongobbler: I'm also wondering if there's a way the `secret_client` or `configmap_client` could be part of the KubeVolume, kind of like a storage driver works for Helm. That way a caller cannot place a KubeVolume in a location other than where the creator intended the volume to go. That would clean up `populate` to just take a `path: &PathBuf`. That being said, this is third-party code so it might not be possible without going through a few hoops, so I'll let you make the call on that.

That cleanup could potentially make it worthwhile implementing `.populate` as a trait for a KubeVolume. That way we can just call `vol.populate(path)` and leave the KubeVolume to error out if it does not support `populate`. Then again, I might be misunderstanding how traits work in Rust so 🤷‍♂️ 

What are your thoughts?

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+                .as_ref()
+                .ok_or_else(|| anyhow::anyhow!("no configmap name was given"))?,
+            namespace,
+            client,
+            path,
+        )
+        .await
+    } else {
+        Err(anyhow::anyhow!(
+            "Unsupported volume type. Currently supported types: ConfigMap, Secret, and HostPath"
```
bacongobbler, line 127: the error mentions that it supports the HostPath volume type, yet it isn't checked for and I think a KubeVolume of type hostpath would error out here if I'm not mistaken. Either we should remove HostPath from the supported types or implement `populate_from_hostpath` (unless that isn't necessary, in which case we should noop)

crates/kubelet/src/volumes.rs
```diff
@@ -0,0 +1,186 @@
...
+impl VolumeRef {
+    /// Resolves the volumes for a pod, including preparing temporary directories containing the
+    /// contents of secrets and configmaps. Returns a HashMap of volume names to a PathBuf for the
+    /// directory where the volume is mounted
+    pub async fn volumes_from_pod(
+        volume_dir: &PathBuf,
+        pod: &Pod,
+        client: &kube::Client,
+    ) -> anyhow::Result<HashMap<String, Self>> {
+        let base_path = volume_dir.join(pod_dir_name(pod));
```
bacongobbler, line 38: this appears to be the only location where `pod_dir_name` is called, and it just looks to be a `format!` call. Is there a particular reason this was broken out as a separate function?

Would it make sense to make the string format (`"{}-{}"`) a constant instead, or is this being used as a constant reference later down the road for something?

crates/kubelet/src/handle.rs
thomastaylor312: That is a good point. I should just unwrap_or_default the option that is passed in

crates/kubelet/src/volumes.rs
thomastaylor312: @bacongobbler let me look at it again

@rylev I tried to do a match but I couldn't get it to work as these are all individual data members of a struct. Is there another way to do this?

crates/kubelet/src/volumes.rs
thomastaylor312: I can't think of a case when `data` shouldn't be set for a `Secret`, so I figured this was better than panicking in the off case that it happens

crates/kubelet/src/volumes.rs
thomastaylor312: Oooh right, I forgot about that handy thing

crates/kubelet/src/volumes.rs
thomastaylor312: `config_map` and `config_map_client` felt really verbose. Generally I agree, but I felt like there was enough context here. I can change it for clarity though

crates/wasi-provider/src/lib.rs
thomastaylor312: Oops, forgot to remove this after I implemented

crates/kubelet/src/volumes.rs
thomastaylor312: So this one we do need because you can have both `binaryData` and `data` or just one or the other in a `ConfigMap` so just unwrapping to a default is fine.

crates/wasi-provider/src/lib.rs
thomastaylor312: Yeah, I can try to pop it out

tests/integration_tests.rs
thomastaylor312: What are your thoughts on maybe just defining them as normal pod YAML manifests and then reading them from a file?

crates/wasi-provider/src/lib.rs
thomastaylor312: So I thought they had to be strings when passing them down the line to the actual wasmtime setup and it turns out I remembered wrong. I'll refactor that

crates/kubelet/src/volumes.rs
thomastaylor312: So I purposefully left that in here because it is the next piece of work to do and I am probably going to do it in the next day or so. Can remove it if you want

crates/kubelet/src/volumes.rs
thomastaylor312: I'll chalk it up to premature optimization. I figured it could possibly be used more down the road. Also, I like that it gives a human readable name to the functionality instead of a bare format call

crates/kubelet/src/volumes.rs
thomastaylor312: @rylev and I talked offline and decided it would be better this way

crates/wasi-provider/src/lib.rs
thomastaylor312: If we did this, we'd have to import `k8s_openapi` and add that direct dependency to our Cargo.toml, so I think I am ok without it. What do you think?

Issue #184, opened 2020-04-12
Feature request: add --node-labels command line option


peterhuene:
Kubelet's `--node-labels` option is used by tools such as [eksctl](https://eksctl.io/) to manage groups of nodes in a K8S cluster.

Here's the help text for the option:

```
Labels to add when registering the node in the cluster.
Labels must be key=value pairs separated by ','.
Labels in the 'kubernetes.io' namespace must begin with an allowed prefix
(kubelet.kubernetes.io, node.kubernetes.io) or be in the specifically allowed set
(beta.kubernetes.io/arch, beta.kubernetes.io/instance-type, beta.kubernetes.io/os,
failure-domain.beta.kubernetes.io/region, failure-domain.beta.kubernetes.io/zone,
failure-domain.kubernetes.io/region, failure-domain.kubernetes.io/zone,
kubernetes.io/arch, kubernetes.io/hostname, kubernetes.io/instance-type,
kubernetes.io/os)
```

It would be nice if krustlet also supported this option so that external node management with such tools would also work.

thomastaylor312:
Labeling this as a good first issue if anyone wanted to get started with krustlet. Shouldn't be too hard to add on to things and the hardest part will likely just be the parsing/validation for the CLI flag (the `Config` object can just accept a `HashMap`)

DerekStrickland:
I've forked and branched.  Trying to figure out how to test it now.  Hopefully, I'll have something for you today or tomorrow.

PR #197, opened 2020-04-20
Allow specifying labels

jingweno:
Add `--node-labels` to the command line option which appends extra
labels when registering a Krustlet node.

Disclaimer: My Rust skill is rusty and my code may not be idiomatic Rust code :-).

This fixes https://github.com/deislabs/krustlet/issues/184

### DIFF
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -11,6 +11,8 @@ use rpassword;
 #[cfg(feature = "cli")]
 use structopt::StructOpt;
 
+use std::collections::HashMap;
+
 const DEFAULT_PORT: u16 = 3000;
 
 /// The configuration needed for a kubelet to run properly.
@@ -33,6 +35,8 @@ pub struct Config {
     pub server_config: ServerConfig,
     /// The directory where the Kubelet will store data
     pub data_dir: PathBuf,
+    /// Labels to add when registering the node in the cluster
+    pub node_labels: HashMap<String, String>,
 }
 /// The configuration for the Kubelet server.
 #[derive(Clone, Debug)]
@@ -58,6 +62,7 @@ impl Config {
         Ok(Config {
             node_ip: default_node_ip(&mut hostname.clone(), preferred_ip_family)?,
             node_name: sanitize_hostname(&hostname),
+            node_labels: HashMap::new(),
             hostname,
             data_dir: default_data_dir()?,
             server_config: ServerConfig {
@@ -97,6 +102,15 @@ impl Config {
         let node_name = opts
             .node_name
             .unwrap_or_else(|| sanitize_hostname(&hostname));
+
+        let node_labels = opts
+            .node_labels
+            .unwrap_or_default()
+            .split(',')
+            .map(|i| split_one_label(i))
+            .filter_map(Result::ok)
+            .collect();
+
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
 
@@ -110,6 +124,7 @@ impl Config {
         Config {
             node_ip,
             node_name,
+            node_labels,
             hostname,
             data_dir,
             server_config: ServerConfig {
@@ -183,6 +198,13 @@ pub struct Opts {
         help = "The IP address of the node registered with the Kubernetes master. Defaults to the IP address of the node name in DNS as a best effort try at a default"
     )]
     node_ip: Option<IpAddr>,
+
+    #[structopt(
+        long = "node-labels",
+        env = "KRUSTLET_NODE_LABELS",
+        help = "Labels to add when registering the node in the cluster. Labels must be key-value pairs separated by ','"
+    )]
+    node_labels: Option<String>,
 
     #[structopt(
         long = "hostname",
@@ -270,4 +292,13 @@ fn is_same_ip_family(first: &IpAddr, second: &IpAddr) -> bool {
 fn read_password_from_tty() -> String {
     let password = rpassword::read_password_from_tty(Some("PFX file password: ")).unwrap();
     return password;
+}
+
+fn split_one_label(in_string: &str) -> Result<(String, String), String> {
+    let mut splitter = in_string.splitn(2, '=');
+    let key = splitter.next().unwrap();
+    match splitter.next() {
+        Some(val) => Ok((key.to_string(), val.to_string())),
+        None => Err("empty label".to_string()),
+    }
 }
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -297,8 +297,12 @@ fn read_password_from_tty() -> String {
 fn split_one_label(in_string: &str) -> Result<(String, String), String> {
     let mut splitter = in_string.splitn(2, '=');
     let key = splitter.next().unwrap();
-    match splitter.next() {
-        Some(val) => Ok((key.to_string(), val.to_string())),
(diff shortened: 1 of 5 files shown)


crates/kubelet/src/config.rs
```diff
@@ -260,3 +282,12 @@ fn is_same_ip_family(first: &IpAddr, second: &IpAddr) -> bool {
         IpAddr::V6(_) => second.is_ipv6(),
     }
 }
+
+fn split_one_label(in_string: &str) -> Result<(String, String), String> {
+    let mut splitter = in_string.splitn(2, '=');
+    let key = splitter.next().unwrap();
+    match splitter.next() {
+        Some(val) => Ok((key.to_string(), val.to_string())),
+        None => Err("empty label".to_string()),
```
peterhuene, line 291: Label values in other resources are allowed to be empty.  Is that not true of nodes?

docs/howto/assets/eks/files/krustlet.service
```diff
@@ -6,6 +6,7 @@ Environment=KUBECONFIG=/etc/eksctl/kubeconfig.yaml
 Environment=PFX_PATH=/etc/krustlet/cert.pfx
 Environment=PFX_PASSWORD=krustlet
 Environment=KRUSTLET_DATA_DIR=/var/cache/krustlet
+Environment=KRUSTLET_NODE_LABELS=alpha.eksctl.io/cluster-name=krustlet-demo,alpha.eksctl.io/nodegroup-name=krustlet
```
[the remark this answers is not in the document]
jingweno: Good call. I write `NODE_LABELS` in a file `/etc/eksctl/kubelet.local.env` and refer to it in the systems file with `NODE_LABELS=...`

crates/kubelet/src/config.rs
jingweno: Yup, labels can be empty. I adjusted the code to take it into account.

crates/kubelet/src/config.rs
```diff
@@ -178,6 +193,13 @@ pub struct Opts {
     )]
     node_ip: Option<IpAddr>,
 
+    #[structopt(
+        long = "node-labels",
+        env = "KRUSTLET_NODE_LABELS",
```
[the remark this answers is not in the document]
jingweno: I agree it's more concise without the `KRUSTLET_` prefix and this will be more consistent how `kubelet` works. I renamed this to `NODE_LABELS`. I will let the rest of the `KRUSTLET_` prefixed flags untouched for now.

docs/howto/assets/eks/files/bootstrap.sh
```diff
@@ -7,6 +7,35 @@ set -o pipefail
...
+}
+
+while [[ $# -gt 0 ]]; do
+    key="$1"
+    case $key in
+        -h|--help)
+            print_help
+            exit 1
+            ;;
+        --krustlet-node-labels)
```
bacongobbler, line 25: isn't the flag name `--node-labels`?

docs/howto/assets/eks/files/bootstrap.sh
```diff
@@ -7,6 +7,35 @@ set -o pipefail
 set -o nounset
 set -o errexit
 
+function print_help {
+    echo "usage: $0 [options] <cluster-name>"
+    echo "Bootstraps a Krustlet instance into an EKS cluster"
+    echo ""
+    echo "-h,--help print this help"
+    echo "--krustlet-node-labels Add extra labels to Krustlet."
```
bacongobbler, line 15: isn't the flag name `--node-labels`?

docs/howto/krustlet-on-eks.md
```diff
@@ -67,7 +67,7 @@ nodeGroups:
     desiredCapacity: 2
     ssh:
       allow: true
-    overrideBootstrapCommand: /etc/eks/bootstrap.sh
+    overrideBootstrapCommand: /etc/eks/bootstrap.sh --krustlet-node-labels "alpha.eksctl.io/cluster-name=krustlet-demo,alpha.eksctl.io/nodegroup-name=krustlet"
```
bacongobbler, line 70: isn't the flag name `--node-labels`?

crates/kubelet/src/config.rs
```diff
@@ -178,6 +193,13 @@ pub struct Opts {
     )]
     node_ip: Option<IpAddr>,
 
+    #[structopt(
+        long = "node-labels",
+        env = "NODE_LABELS",
+        help = "Labels to add when registering the node in the cluster. Labels must be key-value pairs separated by ','"
+    )]
+    node_labels: Option<String>,
```
bacongobbler, line 201: It might be valuable to use a repeated flag instead. That way users can specify node labels like `--node-labels foo=bar --node-labels car=star` rather than one looooong string. It also removes all of the comma splitting logic.

StructOpt also has the option to provide a delimiter, so your example would work the same with

```
#[structopt(long = "node-labels", raw(use_delimiter = "true"))]
node_labels: Vec<String>
```

source: https://github.com/TeXitoi/structopt/issues/94

docs/howto/assets/eks/files/bootstrap.sh
jingweno: This is the bootstrap script that sets up certs, env vars etc. Do we want this flag to name as `--node-labels` too? The `--krustlet` prefix here is to make it clear that this is krustlet's node labels. This script may set up things other than krustlet in the future. I don't have strong opinions on removing the `--krustlet` prefix for the bootstrap script as well.

jingweno: You can see the EKS bootstrap script as an example which also has the kubelet prefix: https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L22

peterhuene: Personally, I think `--krustlet-node-labels` is fine here.

crates/kubelet/src/config.rs
peterhuene: There's incentive to keep it the same way that `kubelet` works, primarily that tools like `eksctl` generate an environment file for the service with `NODE_LABELS` set to a comma-delimited value.

It would be nice to simply pick these values up directly from the tooling without having to special-case it for krustlet (i.e. read the value that got generated by `eksctl` and transform it in bootstrapping to something `krustlet` supports).

peterhuene: Oh, I missed the point that StructOpt would still support the delimiter. :+1: then!

docs/howto/assets/eks/files/bootstrap.sh
bacongobbler: ah thank you! I just noticed this was translated over to NODE_LABELS afterwards.

One question with that behaviour though: wouldn't that override what I set locally for `NODE_LABELS`?

e.g. if I do

```
$ export NODE_LABELS="foo=bar"
./bootstrap.sh
```

As far as I can tell, no node labels will be applied as the script will read from `--krustlet-node-labels` and override the existing environment variable.

crates/kubelet/src/config.rs
jingweno: Good call. I changed it to `use_delimiter = true` in https://github.com/deislabs/krustlet/pull/197/commits/f71f6e1f039710716127d1c38216c8b352b4e0c7. FWIW, `raw` was removed and it becomes https://github.com/TeXitoi/structopt/issues/367#issuecomment-606619645 now.

PR #201, opened 2020-04-20
feat(config): prompt for --pfx-password

### LINKED ISSUE #140: prompt for --pfx-password
Most CLI tools do not allow users to enter the password on the command line. We should instead open a prompt window to ask for the password.

### LINKED ISSUE #141: detect certificate from --pfx-path is password-protected
we should prompt and ask for a password on startup when we find a valid PFX bundle

bacongobbler:
If no password is provided as a flag, prompt for one. Passewords are mandatory for PFX bundles, so there's no need to wrap this in an `Option`.

closes #141 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1741,6 +1741,7 @@ dependencies = [
  "native-tls",
  "oci-distribution",
  "reqwest",
+ "rpassword",
  "serde",
  "serde_json",
  "structopt",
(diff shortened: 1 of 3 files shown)


thomastaylor312:
Test failure was a flake due to the registry pull. I re-kicked them

bacongobbler:
After doing some more thinking, we don't need a feature flag to prompt for the password. If all of the previous functions don't supply one, krustlet will now automatically open a prompt in your terminal:

```
><> ./target/debug/krustlet-wasi
PFX file password:
```
A feature flag like `--pfx-password-stdin` implies the password is optional, which it isn't. I opened #202 to follow up on the discussion raised in #140 around `--pfx-password/$PFX_PASSWORD`. I don't think we should remove them until we have the proper solution in place.

Issue #205, opened 2020-04-20
Create configurable testing module for e2e tests


thomastaylor312:
We need a module that we can configure to read and write from specific directories and to output anything else needed for our e2e tests. 

One outstanding question is whether we should put it in this repo or have it be a separate thing

bacongobbler:
If it makes sense as its own standalone repository, then perhaps it should be broken out of kruslet. On that same note though, have we done any research on other crates out there that may already provide this functionality for us?

For example: https://github.com/mindsbackyard/galvanic-test

itowlson:
I'll take a crack at this as it would be useful for the init containers work.

PR #207, opened 2020-04-20
feat(kubelet): Adds host path volume support

### LINKED ISSUE #179: Implement host path volume mounting
A user should be able to specify a host path to mount into a running instance with both the wascc and wasi providers. We don't need to worry about linux file permissions as they don't matter in a non-container context

thomastaylor312:
Also fixes a few clippy issues we missed earlier.

Closes #179

### DIFF
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -100,9 +100,7 @@ impl Config {
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
 
-        let pfx_password = opts
-            .pfx_password
-            .unwrap_or_else(|| read_password_from_tty());
+        let pfx_password = opts.pfx_password.unwrap_or_else(read_password_from_tty);
 
         let data_dir = opts
             .data_dir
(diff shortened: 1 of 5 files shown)


tests/integration_tests.rs
```diff
@@ -235,6 +236,10 @@ async fn test_wasi_provider() -> Result<(), Box<dyn std::error::Error>> {
                         {
                             "mountPath": "/bar",
                             "name": "configmap-test"
+                        },
+                        {
+                            "mountPath": "/baz",
```
thomastaylor312, line 241: Right now this is just checking that we can open the dir and give it to the running wasmtime module. I opened #205  to capture the need for making a separate module

crates/kubelet/src/config.rs
```diff
@@ -100,9 +100,7 @@ impl Config {
         let port = opts.port;
         let pfx_path = opts.pfx_path.unwrap_or_else(default_pfx_path);
 
-        let pfx_password = opts
-            .pfx_password
-            .unwrap_or_else(|| read_password_from_tty());
+        let pfx_password = opts.pfx_password.unwrap_or_else(read_password_from_tty);
```
bacongobbler, line 103: nice! I wasn't aware of this syntax. TIL

crates/kubelet/src/config.rs
```diff
@@ -268,6 +266,5 @@ fn is_same_ip_family(first: &IpAddr, second: &IpAddr) -> bool {
 }
 
 fn read_password_from_tty() -> String {
-    let password = rpassword::read_password_from_tty(Some("PFX file password: ")).unwrap();
-    return password;
+    rpassword::read_password_from_tty(Some("PFX file password: ")).unwrap()
```
bacongobbler, line 269: 👍

PR #210, opened 2020-04-21
add node label constraints per man page

DerekStrickland:
Update node-labels help to match original [issue](https://github.com/deislabs/krustlet/issues/184) .  Added method that aggregates static, dynamic, and user defined node labels.  User defined node labels are constrained per the man page rules.

### DIFF
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -6,7 +6,8 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::{Api, DeleteParams, PatchParams, PostParams};
 use kube::error::ErrorResponse;
 use kube::Error;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
+use std::collections::HashMap;
 
 macro_rules! retry {
     ($action:expr, times: $num_times:expr, error: $on_err:expr) => {{
@@ -249,15 +250,7 @@ fn node_definition(config: &Config, arch: &str) -> serde_json::Value {
         "kind": "Node",
         "metadata": {
             "name": config.node_name,
-            "labels": {
-                "beta.kubernetes.io/arch": arch,
-                "beta.kubernetes.io/os": "linux",
-                "kubernetes.io/arch": arch,
-                "kubernetes.io/os": "linux",
-                "kubernetes.io/hostname": config.hostname,
-                "kubernetes.io/role":     "agent",
-                "type": "krustlet"
-            },
+            "labels": {},
             "annotations": {
                 "node.alpha.kubernetes.io/ttl": "0",
                 "volumes.kubernetes.io/controller-managed-attach-detach": "true"
@@ -338,8 +331,9 @@ fn node_definition(config: &Config, arch: &str) -> serde_json::Value {
         }
     });
 
+    let node_labels = node_labels_definition(arch, &config);
     // extra labels from config
-    for (key, val) in &config.node_labels {
+    for (key, val) in node_labels {
         json["metadata"]["labels"][key] = serde_json::json!(val);
     }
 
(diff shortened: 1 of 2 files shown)


DerekStrickland:
Didn't have this issue building locally.  I'll take a look in the morning and see if I can resolve and resubmit.

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,66 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
     )
 }
+
+/// Defines the labels that will be applied to this node
+///
+/// Default values and passed node-labels arguments are injected by config.
+fn node_labels_definition(arch: &str, config: &Config) -> HashMap<String, String> {
+    // Add mandatory static labels
+    let mut labels = HashMap::new();
+    labels.insert("beta.kubernetes.io/os".to_string(), "linux".to_string());
```
bacongobbler, line 393: can you switch these to a `to_owned`? While both achieve the same result, one is slightly less computationally expensive since we don't have to do any type casting from an `&str` to a string (at least, that's my understanding).

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,66 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+            warn!(
+                "User provided node label {} omitted. Namespace label managed by runtime.",
+                key
+            );
+        } else if key.contains(k8s_namespace)
+            && (key.starts_with("kubelet.kubernetes.io")
+                || key.starts_with("node.kubernetes.io")
+                || allowed_k8s_namespace_labels.contains(&&key[..]))
+        {
+            labels.insert(key.to_owned(), value.to_owned());
```
bacongobbler, line 442: this call can be moved after the `if` block, and the `else` clause can be removed. Net result will be the same thing, but it will remove the `clippy` error

https://github.com/rust-lang/rust-clippy

crates/kubelet/src/node.rs
bacongobbler: http://www.lowlevelmanager.com/2016/02/rust-tostring-vs-toowned-for-string.html

crates/kubelet/src/node.rs
```diff
@@ -249,15 +250,7 @@ fn node_definition(config: &Config, arch: &str) -> serde_json::Value {
...
-            "labels": {
-                "beta.kubernetes.io/arch": arch,
-                "beta.kubernetes.io/os": "linux",
-                "kubernetes.io/arch": arch,
-                "kubernetes.io/os": "linux",
-                "kubernetes.io/hostname": config.hostname,
-                "kubernetes.io/role":     "agent",
-                "type": "krustlet"
-            },
+            "labels": {},
```
jingweno, line 253: I personally think it's cleaner to specify the default labels using JSON literal as it did before, instead of all the `to_string` or `to_owned` in https://github.com/deislabs/krustlet/pull/210/files#diff-3c33c7d659ceab610971cced03231c9bR392-R403.

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,66 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
         }
     )
 }
+
+/// Defines the labels that will be applied to this node
+///
+/// Default values and passed node-labels arguments are injected by config.
+fn node_labels_definition(arch: &str, config: &Config) -> HashMap<String, String> {
```
jingweno, line 390: Could we add a unit test for this method?

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,66 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+    for (key, value) in user_labels.iter() {
+        if managed_namespace_labels.contains(&&key[..]) {
+            warn!(
+                "User provided node label {} omitted. Namespace label managed by runtime.",
+                key
+            );
+        } else if key.contains(k8s_namespace)
+            && (key.starts_with("kubelet.kubernetes.io")
+                || key.starts_with("node.kubernetes.io")
+                || allowed_k8s_namespace_labels.contains(&&key[..]))
```
jingweno, line 440: This doesn't look correct to me. I think we want to warn about invalid keys that don't follow the spec. This is what I'm thinking:

```diff
diff --git a/crates/kubelet/src/node.rs b/crates/kubelet/src/node.rs
index 9c69bd9..ae652a7 100644
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -6,7 +6,7 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::{Api, DeleteParams, PatchParams, PostParams};
 use kube::error::ErrorResponse;
 use kube::Error;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 
 macro_rules! retry {
     ($action:expr, times: $num_times:expr, error: $on_err:expr) => {{
@@ -339,13 +339,48 @@ fn node_definition(config: &Config, arch: &str) -> serde_json::Value {
     });
 
     // extra labels from config
-    for (key, val) in &config.node_labels {
+    for (key, val) in config.node_labels.iter().filter(|(k, _)| is_invalid_node_label_key(k)) {
         json["metadata"]["labels"][key] = serde_json::json!(val);
     }
 
     json
 }
 
+fn is_invalid_node_label_key(key: &str) -> bool {
+    let managed_namespace_labels = vec![
+        "beta.kubernetes.io/arch",
+        "beta.kubernetes.io/os",
+        "kubernetes.io/arch",
+        "kubernetes.io/hostname",
+        "kubernetes.io/os",
+        "kubernetes.io/role",
+        "type",
+    ];
+    let is_invalid = managed_namespace_labels.contains(&key) || is_invalid_k8s_node_label_key(key);
+
+    if is_invalid {
+        warn!("invalid label {}", key);
+    }
+
+    is_invalid
+}
+
+fn is_invalid_k8s_node_label_key(key: &str) -> bool {
+    let allowed_k8s_namespace_labels = vec![
+        "beta.kubernetes.io/instance-type",
+        "failure-domain.beta.kubernetes.io/region",
+        "failure-domain.beta.kubernetes.io/zone",
+        "failure-domain.kubernetes.io/region",
+        "failure-domain.kubernetes.io/zone",
+        "kubernetes.io/instance-type",
+    ];
+
+    key.contains("kubernetes.io")
+        && !(key.starts_with("kubelet.kubernetes.io")
+            || key.starts_with("node.kubernetes.io")
+            || allowed_k8s_namespace_labels.contains(&key))
+}
+
 /// Define a new coordination.Lease object for Kubernetes
 ///
 /// The lease tells Kubernetes that we want to claim the node for a while

```

DerekStrickland: This is interesting.  Filter instead of builder....

crates/kubelet/src/node.rs
DerekStrickland: I think clippy actually saved me from myself here.  I added a warn! for this second error condition.

crates/kubelet/src/node.rs
DerekStrickland: I struggled with this decision. On the one hand it's more readable, on the other, you have two different methods managing the same thing.  I don't feel strongly about it one way or the other, so I defaulted to one method for one thing.  That said, if we followed the suggestion you make later on about filtering instead of building, this problem kind of goes away.

crates/kubelet/src/node.rs
DerekStrickland: I tried, but struggled as it isn't a public method, and none of the other foo_definition methods in this class are public, and I didn't want to A) break what appears to be an intentional design or B) start introducing DI which is how I would do it in other languages.  I'm new to rust.  I'd be open to some help or advice here if you have any.

crates/kubelet/src/node.rs
DerekStrickland: Done.

crates/kubelet/src/node.rs
jingweno: For unit tests, they are usually in the same file, e.g. https://doc.rust-lang.org/book/ch11-01-writing-tests.html. A real-world example: https://github.com/bytecodealliance/wasmtime/blob/7eea5d8d438bf0d65710410c0acc58e5b27018d7/crates/runtime/src/vmcontext.rs#L53-L62.

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,69 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+    for (key, value) in user_labels.iter() {
+        if managed_namespace_labels.contains(&&key[..]) {
+            warn!(
+                "User provided node label {} omitted. Namespace label managed by runtime.",
+                key
+            );
+        } else if key.contains(k8s_namespace)
+            && !key.starts_with("kubelet.kubernetes.io")
+            && !key.starts_with("node.kubernetes.io")
+            && !allowed_k8s_namespace_labels.contains(&&key[..])
```
jingweno, line 440: Why would you need to copy the key here? `!allowed_k8s_namespace_labels.contains(&key)` would be more efficient.

crates/kubelet/src/node.rs
DerekStrickland: Oh wow!  I had no idea you could just super:: out of the test mod like that.  Thx.  I'm giving it a whirl.

crates/kubelet/src/node.rs
DerekStrickland: I tried that originally but the borrow checker complains like so 

&& !allowed_k8s_namespace_labels.contains(&key)
    |                                                       ^^^^ expected `str`, found struct `std::string::String`

I felt icky writing &&key[..] but I'm not sure of a better way to handle it.  Totally open to input.

crates/kubelet/src/node.rs
DerekStrickland: @bacongobbler kk.  I think it's ready for a review.

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,120 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+    let managed_namespace_labels = vec![
+        "beta.kubernetes.io/arch",
+        "beta.kubernetes.io/os",
+        "kubernetes.io/arch",
+        "kubernetes.io/hostname",
+        "kubernetes.io/os",
+        "kubernetes.io/role",
+        "type",
+    ];
+    let allowed_k8s_namespace_labels = vec![
```
rylev, line 416: Since this list doesn't need to grow any bigger, you can just make it a fixed sized array by dropping the `vec!` part of the declaration.

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,120 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+        "failure-domain.kubernetes.io/region",
+        "failure-domain.kubernetes.io/zone",
+        "kubernetes.io/instance-type",
+    ];
+
+    // Attempt to append node labels from passed arguments.
+    // First, check for managed namespace and log exclusion
+    // Next, check if label contains k8s namespace and ensure it's allowable
+    // Else, if not k8s namspace, insert
+    let user_labels = config.node_labels.clone();
```
rylev, line 429: The clone is not neccesarry. You can borrow instead:
```rust
let user_labels = &config.node_labels;
```

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,120 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
         }
     )
 }
+
+/// Defines the labels that will be applied to this node
+///
+/// Default values and passed node-labels arguments are injected by config.
+fn node_labels_definition(arch: &str, config: &Config) -> HashMap<String, String> {
```
rylev, line 390: Are the node labels always known at compile time? Is there any reason why the hashmap can't be `HasMap<&'static str, &'static str>`?

crates/kubelet/src/node.rs
```diff
@@ -389,3 +383,120 @@ fn lease_spec_definition(node_name: &str) -> serde_json::Value {
...
+    ];
+
+    // Attempt to append node labels from passed arguments.
+    // First, check for managed namespace and log exclusion
+    // Next, check if label contains k8s namespace and ensure it's allowable
+    // Else, if not k8s namspace, insert
+    let user_labels = config.node_labels.clone();
+
+    for (key, value) in user_labels.iter() {
+        if managed_namespace_labels.contains(&&key[..]) {
```
rylev, line 432: Slightly less awkward syntax would be:

```rust
managed_namespace_labels.contains(&key.as_str())
```

crates/kubelet/src/node.rs
DerekStrickland: They aren't always known. There are three types of labels as I understand it.  Some A)are static and know at compile type, some have a static key, but a variable value (depend on the arch parameter), and some are supplied by the user. User supplied values are loaded into Config.node_labels which is a HashMap<String, String>. I think that implies that no, it can't be `HashMap<&'static str, &'static str>`.  Sound right?

crates/kubelet/src/node.rs
DerekStrickland: applied and pushed

crates/kubelet/src/node.rs
DerekStrickland: applied and push.  Thanks for the coaching!

crates/kubelet/src/node.rs
DerekStrickland: applied and pushed

DerekStrickland:
@rylev ready for review I think.  Not sure who to inform

crates/kubelet/src/node.rs
bacongobbler: was this resolved?

DerekStrickland: yes.  Unit test has been added

bacongobbler:
thanks @DerekStrickland!

crates/kubelet/src/node.rs
rylev: Yes that's correct.

PR #211, opened 2020-04-21
fix(kubelet): Adds per pod queuing

thomastaylor312:
Our old event handling method locked the provider for any pod until event
handling was complete, essentially limiting us to one action at a time.
This introduces a new queuing method that creates a per pod queue that
only handles the latest sent event.

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -2,14 +2,20 @@
 ///! Kubelet with a specific handler (called a `Provider`)
 use crate::config::Config;
 use crate::node::{create_node, update_node};
+use crate::queue::PodQueue;
 use crate::server::start_webserver;
+use crate::status::Phase;
 use crate::Provider;
 
 use futures::{StreamExt, TryStreamExt};
 use k8s_openapi::api::core::v1::Pod as KubePod;
-use kube::{api::ListParams, runtime::Informer, Api};
-use log::{debug, warn};
-use tokio::sync::Mutex;
+use kube::{
+    api::{ListParams, Meta, PatchParams, Resource},
+    runtime::Informer,
+    Api,
+};
+use log::{debug, error, warn};
+use tokio::sync::mpsc;
 
 use std::sync::Arc;
 
@@ -26,7 +32,7 @@ use std::sync::Arc;
 /// run one (instance of a) Provider. So a provider may be passed around from
 /// thread to thread during the course of the Kubelet's lifetime.
 pub struct Kubelet<P> {
-    provider: Arc<Mutex<P>>,
+    provider: Arc<P>,
     kube_config: kube::Config,
     config: Config,
 }
@@ -36,7 +42,7 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
     /// and a kubelet configuration
     pub fn new(provider: T, kube_config: kube::Config, config: Config) -> Self {
         Self {
-            provider: Arc::new(Mutex::new(provider)),
+            provider: Arc::new(provider),
             kube_config,
             config,
         }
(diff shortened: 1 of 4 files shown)


crates/kubelet/src/kubelet.rs
```diff
@@ -63,8 +69,53 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
...
-        // This informer listens for pod events.
-        let provider = self.provider.clone();
+        // TODO: How should we configure this value? We should eventually have a max pods setting
+        // just like a normal kubelet, so maybe that?
+        let (error_sender, mut error_receiver) = mpsc::channel::<(KubePod, anyhow::Error)>(200);
+        let client_clone = client.clone();
+        let error_handler = tokio::task::spawn(async move {
+            let client = client_clone;
+            while let Some(e) = error_receiver.recv().await {
+                let (pod, err) = e;
```
rylev, line 79: nit: can you just inline this in the `Some` pattern?

crates/kubelet/src/kubelet.rs
```diff
@@ -63,8 +69,53 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
...
+                let json_status = serde_json::json!(
+                    {
+                        "metadata": {
+                            "resourceVersion": "",
+                        },
+                        "status": {
+                            "phase": Phase::Failed,
+                            "message": format!("{}", err),
```
rylev, line 87: Is the `format!` necessary?

crates/kubelet/src/kubelet.rs
```diff
@@ -63,8 +69,53 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
...
+
+                debug!(
+                    "Setting pod status for {} using {:?}",
+                    pod.name(),
+                    json_status
+                );
+
+                let data = serde_json::to_vec(&json_status).expect("Should always serialize");
+                let pod_resource =
+                    Resource::namespaced::<KubePod>(&pod.namespace().unwrap_or_default());
```
rylev, line 100: Is there simpler way to do this using the `Api` api from kube?

crates/kubelet/src/kubelet.rs
```diff
@@ -63,8 +69,53 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
...
+                let pod_name = pod.name();
+                match pod_resource.patch_status(&pod_name, &PatchParams::default(), data) {
+                    Ok(req) => {
+                        if let Err(e) = client.request::<KubePod>(req).await {
+                            error!(
```
rylev, line 105: Is just logging an error the right thing to do here?

crates/kubelet/src/queue.rs
```diff
@@ -0,0 +1,114 @@
...
+use k8s_openapi::api::core::v1::Pod as KubePod;
+use kube::api::{Meta, WatchEvent};
+use log::{debug, error};
+use tokio::sync::{mpsc::Sender, watch, Mutex};
+use tokio::task::JoinHandle;
+
+use crate::handle::pod_key;
+use crate::Provider;
+
+/// A per-pod queue that takes incoming Kubernetes events and broadcasts them to the correct queue
```
rylev, line 13: Can you break this documentation up so that the first sentence is on its own line. That makes generated documentation easier to read

crates/kubelet/src/queue.rs
```diff
@@ -0,0 +1,114 @@
...
+        match &event {
+            WatchEvent::Added(pod)
+            | WatchEvent::Bookmark(pod)
+            | WatchEvent::Deleted(pod)
+            | WatchEvent::Modified(pod) => {
+                let mut handlers = self.handlers.lock().await;
+                let pod_name = pod.name();
+                let pod_namespace = pod.namespace().unwrap_or_default();
+                let entry = handlers
+                    .entry(pod_key(&pod_namespace, &pod_name))
```
rylev, line 78: I'm pretty sure if you use pattern matching instead of the entry API you can eliminate the mutex. 

Even with the entry API the only thing we're using the mutex for is to be able to borrow self in the `or_insert_with` closure. A mutex is a _way_ more expensive then a two clones of `Arc`s

crates/kubelet/src/queue.rs
```diff
@@ -0,0 +1,114 @@
...
+                    .or_insert_with(|| {
+                        Worker::create(
+                            event.clone(),
+                            self.provider.clone(),
+                            self.error_sender.clone(),
+                        )
+                    });
+                // TODO: There is a possible condition where we get two modifies right in a row and
```
rylev, line 86: This isn't true. Enqueues will be done in the order that we get them from the informer. This functions `&mut self` ensures that only one thread can run this code at a time. 

This means we're reliant on whatever ordering guarantees the `kube` Informer implementation has.

crates/kubelet/src/server.rs
```diff
@@ -8,7 +8,6 @@ use log::{debug, error, info};
 use native_tls::{Identity, TlsAcceptor};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::stream::StreamExt;
-use tokio::sync::Mutex;
```
rylev, line 11: This makes me happy. 😀

rylev:
If we get an update before we've had the chance to handle the create, then the create will not be processed. That's fine?

crates/kubelet/src/kubelet.rs
thomastaylor312: Oh, I guess it will just get serialized, so no.

crates/kubelet/src/kubelet.rs
thomastaylor312: We don't want to crash things just because of a single error. We could try wrapping this in the retry stuff you wrote though

crates/kubelet/src/kubelet.rs
thomastaylor312: Actually yes. I think I was planning on doing something different here and then never did it

thomastaylor312:
> If we get an update before we’ve had the chance to handle the create, then the create will not be processed. That’s fine?
I think we are ok here. As long as things are functioning, we should always get a create first. If there is some sort of hiccup, then we could get a modified event afterwards. We could add some detection around this and call a reset on the informer if needed. Otherwise, the informer will likely detect that we missed something anyway (it'll get a `410 Gone`) and then reset itself. On reset we get added events

crates/kubelet/src/kubelet.rs
thomastaylor312: Looks like anyhow doesn't implement serialize in a way that serde likes

crates/kubelet/src/queue.rs
thomastaylor312: You were correct. It is better now

PR #218, opened 2020-04-24
posix-compatible lowercasing

bacongobbler:
The bash substitution is not portable and will not work on macOS runners (old version of bash?). Using `tr` is portable and should work across both POSIX systems.


bacongobbler:
confirmed working in https://github.com/deislabs/krustlet/pull/218/checks?check_run_id=616242004.

PR #222, opened 2020-04-29
feat(wascc): add volume support via fs-provider

### LINKED ISSUE #192: Implement hostpath, Secret, and ConfigMap volumes for Wascc
Implement volume "mounting" for the wascc provider. You cannot configure directories directly for an actor, but you can using the fs-provider: https://github.com/wascc/fs-provider.

bacongobbler:
closes #192


bacongobbler:
e2e failure to follow up on:

```
message: 'Failed to run pod: Error configuring capabilities for module: WebAssembly
    module authorization failure: Unauthorized binding: actor MDXEE6PA62JOB2L3GC7UVOGL7A746PLGCKGSBI2G4SV5H36CGAYLRQZC
    is not authorized to use capability wascc:blobstore.'
```

This is because Kubernetes wants to inject a service account into the webassembly module, so the `wascc:blobstore` capability is being added to the actor. A check may need to be put in place so that these capabilities are not bound to the actor when it is not authorized for those bindings.

crates/wascc-provider/src/lib.rs
```diff
@@ -422,12 +432,73 @@ fn wascc_run(
...
+    // [`NativeCapability::from_instance`].
+    //
+    // Portable capabilities are WASM modules.  Portable capabilities
+    // don't fully work, and won't until the WASI spec has matured.
+    //
+    // Here we are using the native capabilties as statically linked libraries that will
+    // be compiled into the wascc-provider binary.
+    info!("Loading HTTP capability");
+    let http_provider = HttpServerProvider::new();
+    let data = NativeCapability::from_instance(http_provider, None)
```
thomastaylor312, line 486: My understanding was that only one http provider should be run at a time. Did that only apply to the shared object files?

bacongobbler:
fixed and ready for review :)

crates/wascc-provider/src/lib.rs
```diff
@@ -422,12 +432,73 @@ fn wascc_run(
...
+    if load.capabilities().contains(&FS_CAPABILITY.to_owned()) {
+        for vol in &volumes {
+            info!(
+                "Loading File System capability for volume name: '{}' host_path: '{}'",
+                vol.name,
+                vol.host_path.display()
+            );
+            let mut fsenv: HashMap<String, String> = HashMap::new();
+            fsenv.insert(
+                "ROOT".to_owned(),
```
thomastaylor312, line 451: Nit: This should probably be a constant

crates/wascc-provider/src/lib.rs
```diff
@@ -422,12 +432,73 @@ fn wascc_run(
...
     capabilities.push(Capability {
         name: LOG_CAPABILITY,
+        binding: None,
         env: logenv,
     });
 
     let load = Actor::from_bytes(data).map_err(|e| anyhow::anyhow!("Error loading WASM: {}", e))?;
     let pk = load.public_key();
 
+    if load.capabilities().contains(&FS_CAPABILITY.to_owned()) {
```
thomastaylor312, line 442: So what is happening here? Does the module have to specify that it needs an FS_CAPABILITY? Also, I think you can just do `FS_CAPABILITY` here instead of doing `to_owned` because it is already an `&str`

demos/wascc/fileserver/Makefile
```diff
@@ -0,0 +1,37 @@
+COLOR ?= always # Valid COLOR options: {always, auto, never}
```
thomastaylor312, line 1: Any reason you used `make` over `just` here?

demos/wascc/fileserver/Makefile
```diff
@@ -0,0 +1,37 @@
...
+
+keys: keys-account
+keys: keys-module
+
+keys-account:
+	@mkdir -p $(KEYDIR)
+	nk gen account > $(KEYDIR)/account.txt
+	awk '/Seed/{ print $$2 }' $(KEYDIR)/account.txt > $(KEYDIR)/account.nk
+
+keys-module:
```
thomastaylor312, line 34: Super Nit (especially because this is a demo): you could make this depend on keys-account and then make `build` depend on this. You could also have a `$(KEYDIR).account.nk` target with `keys-account` being an alias so it doesn't rebuild if you already have it.

demos/wascc/fileserver/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Running the example
+
+This example has already been pre-built, so you only need to install it into your Kubernetes
+cluster.
+
+Create the pod and configmap with `kubectl`:
+
+```shell
+$ kubectl create -f k8s.yaml
+```
```
thomastaylor312, line 22: With all the different `make` targets and other goodies you added in here, I'd really love to see some building instructions with the different toolsets I need.

demos/wascc/fileserver/k8s.yaml
```diff
@@ -0,0 +1,36 @@
...
+      imagePullPolicy: Always
+      name: fileserver-wascc
+      env:
+        - name: PORT
+          value: "8080"
+      ports:
+        - containerPort: 8080
+      volumeMounts:
+        - name: storage
+          mountPath: /tmp # this is ignored by waSCC, but is still necessary from the k8s API
```
thomastaylor312, line 19: I am not clear what this means. Does that mean I can't mount `/tmp` from the host to an arbitrary directory within the container?

demos/wascc/fileserver/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    codec::core::OP_HEALTH_REQUEST => health
+}
+
+fn fetch(r: codec::http::Request) -> CallResult {
+    // k8s volumes are mounted into the waSCC runtime using the same volume mount name
+    let store = objectstore::host("storage");
+    let mut path = String::from(r.path);
+
+    // strip the leading slash from the path
+    path = path.replacen("/", "", 1);
```
thomastaylor312, line 22: This should probably use `trim_start_matches` in case for some reason the path doesn't start with a `/`

crates/wascc-provider/src/lib.rs
bacongobbler: Kubernetes attempts to inject a service account into each application runtime, so it is guranteed that each application will have at least one volume. For waSCC, if the application does not include a particular capability, then it will fail to run.

That's why we have to check if the actor has been signed with the ability to use the filesystem capability. If it has not signed off on the `wascc::blobstore` capability (like with the `greet`, `hello-world-assemblyscript` and `uppercase` demos), it will fail to load.

demos/wascc/fileserver/Makefile
bacongobbler: copypasta from the `uppercase` demo.

demos/wascc/fileserver/README.md
bacongobbler: this might be a good question for @bketelsen as he worked on the makefile for the uppercase demo. I'm just copying code here :)

demos/wascc/fileserver/k8s.yaml
bacongobbler: No, because waSCC does not mount or pre-open directories like WASI does. It instead streams files to and from the actor using the filesystem capability.

From [the docs](https://github.com/wascc/fs-provider/):

> Because of the way WebAssembly and the waSCC host work, all `wascap:blobstore` capability providers must stream files to and from the actor. This allows actors to unblock long enough to allow other messages from other providers to be processed and keeps the WebAssembly module from allocating too much memory.

In other words, a directory is not mounted into the runtime. It's accessed through a handle to fetch and push files to the filesystem, effectively nullifying the `mountPath`.

crates/wascc-provider/src/lib.rs
bacongobbler: Good point. I was trying some things out earlier. We can put this back where it was now

demos/wascc/fileserver/k8s.yaml
bacongobbler: https://github.com/deislabs/krustlet/blob/9914e7b4f0ed611eb96b965d71d2142c6bee1406/demos/wascc/fileserver/src/lib.rs#L17-L18

demos/wascc/fileserver/Makefile
thomastaylor312: Nah, no hard requirement here

demos/wascc/fileserver/k8s.yaml
thomastaylor312: Ah ok that makes more sense

demos/wascc/fileserver/Makefile
bacongobbler: Good call. I actually ran into that exact issue.

crates/wascc-provider/src/lib.rs
```diff
@@ -133,33 +148,28 @@ impl<S: ModuleStore + Send + Sync> WasccProvider<S> {
...
-        tokio::task::spawn_blocking(move || {
-            info!("Loading HTTP Capability");
-            let http_provider = HttpServerProvider::new();
-            let data = NativeCapability::from_instance(http_provider, None)
-                .map_err(|e| anyhow::anyhow!("Failed to instantiate HTTP capability: {}", e))?;
+        info!("Loading HTTP capability");
+        let http_provider = HttpServerProvider::new();
+        let data = NativeCapability::from_instance(http_provider, None)
+            .map_err(|e| anyhow::anyhow!("Failed to instantiate HTTP capability: {}", e))?;
+        host.lock()
```
thomastaylor312, line 155: The host lock should use a `tokio::sync::Mutex` if we are going to take this out of the `spawn_blocking`

demos/wascc/fileserver/Makefile
```diff
@@ -0,0 +1,36 @@
+COLOR ?= always # Valid COLOR options: {always, auto, never}
+CARGO = cargo --color $(COLOR)
+TARGET = target/wasm32-unknown-unknown
+DEBUG = $(TARGET)/debug
+RELEASE = $(TARGET)/release
+KEYDIR ?= .keys
+VERSION = 0.1.0
+NAME = fileserver
+
+.PHONY: all bench build check clean doc test update keys keys-account keys-module
```
thomastaylor312, line 10: You don't need `keys-account` or `keys-module` any more

crates/wascc-provider/src/lib.rs
bacongobbler: once again, my blind copy/paste.... The old code cloned the host lock and threw it in a spawn_blocking. Let me fix that up.

PR #223, opened 2020-04-30
feat(wasi): Adds interrupt of long running processes

### LINKED ISSUE #212: Add stopping of long running WASM processes
Now that wasmtime has [support for stopping a running instance](https://github.com/bytecodealliance/wasmtime/pull/1490), we should be able to add this in to our `Stop` implementation for the wasi provider.

This will need to wait until a new version of the wasmtime crate is released

thomastaylor312:
There isn't an easy way to assert on this for a test, but you can see
it working by replacing the image in any of the examples with
webassembly.azurecr.io/waiter:v0.1.0

Closes #212


crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -242,13 +255,18 @@ impl WasiRuntime {
...
                 })
                 .expect("status should be able to send");
-            match instance
-                .get_export("_start")
-                .expect("_start import should exist in wasm module")
-                .func()
-                .unwrap()
-                .call(&[])
-            {
+            let export = instance.get_export("_start").ok_or(anyhow::anyhow!(
```
peterhuene, line 258: Perhaps `ok_or_else` rather than `ok_or`?  Depending on the features used, `anyhow!` can do some expensive things such as capture a backtrace or format error messages.

thomastaylor312: Good point. I'll switch that out

thomastaylor312:
@peterhuene We really appreciate the work there. The API was nice, clean, and easy to use

PR #224, opened 2020-04-30
initial add of simple server

technosophos:
This is just a simple server based on @bacongobbler 's tutorial. It is nice to have a demo that runs for more than a second or two, so I thought we should check this in.


demos/wasi/simpleserver/simpleserver.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: simpleserver
+  labels:
+    app: simpleserver
+spec:
+  containers:
+    - image: webassembly.azurecr.io/simpleserver:v1.0.0
+      imagePullPolicy: Always
+      name: simpleserver
+      ports:
+        - containerPort: 80
```
thomastaylor312, line 13: I don't think you need this container port

demos/wasi/simpleserver/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "simpleserver"
+version = "0.1.0"
+authors = ["Matt Butcher <matt.butcher@microsoft.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
thomastaylor312, line 9: Can you add an empty `[workspace]` block here? It silences a warning from the top level cargo that it has some unknown thing in its workspace

technosophos: Oh, _that's_ how you do it? Google had utterly failed me

demos/wasi/simpleserver/simpleserver.yaml
technosophos: I copied the YAML straight from teh tutorial, so we should probably fix it there in a follow-up PR.

technosophos:
Fixed

PR #225, opened 2020-05-04
Improve Kubelet Log API

kflansburg:
* Add support for tailing logs (`tailLines` query param).
* Add support for following logs (`follow` query param).
* Async stream log lines to client using `hyper::Body::channel`.
* Allows for multiple log clients.

Closes: #49

Issue #49 was already fixed on master.


crates/kubelet/src/handle.rs
```diff
@@ -66,10 +163,16 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
...
-        Ok(bytes_written)
+    pub(crate) async fn output(
+        &mut self,
+        sender: hyper::body::Sender,
+        tail: Option<usize>,
+        follow: bool,
+    ) -> anyhow::Result<()> {
+        let mut output = self.handle.output();
+        output.seek(SeekFrom::Start(0)).await?;
+        tokio::spawn(stream_logs(output, sender, tail, follow));
```
thomastaylor312, line 174: For my own edification, this is just going to return immediately and then any errors are passed to the user through `hyper::body::Sender`?

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
     async fn wait(&mut self) -> anyhow::Result<()>;
 }
 
+/// Trait to describe necessary behavior for creating multiple log readers.
+/// TODO: Both providers make a handle containing a tempfile. If this is a common pattern,
+/// it might make sense to provide that implementation here. This would add `tempfile` as a
+/// dependency of `kubelet`.
+pub trait LogHandle<R>: Sync + Send {
```
thomastaylor312, line 41: So after looking at the code, do we need to use this trait? We already have an `R` that is constrained with the proper traits, so couldn't we just reuse that, avoid the `Box` altogether and then just have `stream_logs` accept `Lines<BufReader>`? No matter what the implementation, it should always be a reader of some type, so we should be able to get a BufReader. 

@rylev any thoughts here?

crates/kubelet/src/provider.rs
```diff
@@ -69,7 +69,10 @@ pub trait Provider {
         namespace: String,
         pod: String,
         container: String,
-    ) -> anyhow::Result<Vec<u8>>;
+        sender: hyper::body::Sender,
```
thomastaylor312, line 72: I am not sure if there is a way around this, but it feels a little...dirty? It just feels weird to be passing a hyper specific sender all the way down the stack and making it part of a public trait. Is there any way we can get around this or make it more generic? 

Once again, @rylev your thoughts would be appreciated here

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
     async fn wait(&mut self) -> anyhow::Result<()>;
 }
 
+/// Trait to describe necessary behavior for creating multiple log readers.
+/// TODO: Both providers make a handle containing a tempfile. If this is a common pattern,
```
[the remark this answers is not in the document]
kflansburg: Yep, I figured as much, do you think it is safe to say that all logging will be done to a file? The generic handle could just accept a path. Regardless, providers would still be free to implement the trait for whatever they are using.

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
...
 /// Pod
 pub struct RuntimeHandle<S, R> {
     stopper: S,
-    output: R,
+    handle: Box<dyn LogHandle<R>>,
     status_channel: Receiver<ContainerStatus>,
 }
 
-impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
+impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin + Send + 'static> RuntimeHandle<S, R> {
```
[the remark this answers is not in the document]
kflansburg: Agreed, will do.

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
...
+}
+
 /// Represents a handle to a running "container" (whatever that might be). This
 /// can be used on its own, however, it is generally better to use it as a part
 /// of a [`PodHandle`], which manages a group of containers in a Kubernetes
 /// Pod
 pub struct RuntimeHandle<S, R> {
     stopper: S,
-    output: R,
+    handle: Box<dyn LogHandle<R>>,
```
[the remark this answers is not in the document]
kflansburg: Hmmm, I will see if that works. May have to add a third generic type.

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
...
+                            debug!("channel closed.");
+                            return Ok(());
+                        } else {
+                            error!("channel error: {}", e);
+                            bail!(e);
+                        }
+                    }
+                }
+            }
+            tokio::time::delay_for(std::time::Duration::from_millis(500)).await;
```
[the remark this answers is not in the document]
kflansburg: It doesn't block at EOF I believe (notice that I don't have to detect EOF when streaming the file at the start of the function. So I think this will busy wait on the outside loop without this delay.

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,123 @@ pub trait Stop {
...
+) -> anyhow::Result<()> {
+    let buf = tokio::io::BufReader::new(output);
+    let mut lines = buf.lines();
+
+    if let Some(n) = tail {
+        // Stream last n lines.
+        // TODO: this uses a lot of memory for large n and scans the entire file.
+        let mut line_buf = std::collections::VecDeque::with_capacity(n);
+
+        while let Some(line) = lines.next_line().await? {
```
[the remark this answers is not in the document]
kflansburg: I gave this a shot, but lines does not actually implement DoubleEndedIterator and therefore not rev(), so the only way would be to read the whole file into memory in a Vec and create a new Iterator from that. 

https://users.rust-lang.org/t/idiomatic-way-of-reading-a-text-file-line-by-line-in-reverse/8547

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev

crates/kubelet/src/handle.rs
kflansburg: Yes, it does. It doesn't handle errors very well (aside from logging them). Right now, the channel just gets dropped and then hyper closes the request. I could modify it to at least send the error to the client before that. I don't think that there is much else to be done since the response code / headers have already been sent at that point.

crates/kubelet/src/handle.rs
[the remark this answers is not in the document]
bacongobbler: I'm not too sure on that assumption. One could write a provider that streams logs elsewhere. Having to avoid the middle-ground of writing to a file, then streaming that file somewhere else would be nice to avoid

crates/kubelet/src/handle.rs
kflansburg: I ran into a couple of issues here.

1. This does not allow you to service two log requests at once because you would have to borrow / transfer ownership of the reader to the scope servicing the first request. 
2. Enforcing that R is clone is actually not very easy, for instance tokio::fs::File is not clone. 
3. Even if you can have two tasks reading from the same async reader, a new connection would cause it to seek and other clients would receive different sections of the reader than they were expecting. 

So at the very least we need to know how to create new AsyncReaders on demand, and they need to be Sync + Send to get spawned into the streaming function.

crates/kubelet/src/provider.rs
kflansburg: I agree, it is not ideal, initially I tried to at least stick to using something from tokio instead. I can explore adding a new type so that `kubelet` can change the underlying implementation in the future without breaking provider implementations.

kflansburg:
@thomastaylor312 I can give some of the changes that you brought up a crack tonight if you want to wait before merging.

crates/kubelet/src/handle.rs
kflansburg: Was able to handle errors when reading from AsyncRead. The only other error is when sending to channel, which probably cannot be relayed to the user at that point.

crates/kubelet/src/handle.rs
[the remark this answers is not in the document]
kflansburg: Made handle the generic. Traits on R and LogHandle<R> are only enforced on the `output` method now.

crates/kubelet/src/provider.rs
kflansburg: Created a new type to wrap this. Should present a cleaner, more stable API to users.

crates/kubelet/src/server.rs
```diff
@@ -83,12 +84,40 @@ where
...
+                        Ok(n) => tail = Some(n),
+                        Err(e) => {
+                            warn!(
+                                "Unable to parse tailLines query parameter ({}): {:?}",
+                                value, e
+                            );
+                        }
+                    },
+                    "follow" if value == "true" => follow = true,
+                    s => warn!("Unknown query parameter: {}={}", s, value),
```
rylev, line 110: Do we actually want to warn when follow is explicitly set to "false"?

crates/wasi-provider/Cargo.toml
```diff
@@ -27,6 +27,7 @@ tokio = { version = "0.2", features = ["fs", "stream", "macros", "io-util", "syn
 chrono = { version = "0.4", features = ["serde"] }
 futures = "0.3"
 k8s-openapi = { version = "0.7", features = ["v1_17"] }
+hyper = "0.13"
```
rylev, line 30: This shouldn't be needed anymore

crates/kubelet/src/handle.rs
rylev: I don't think issue 1 and 2 are necessarily impossible to overcome, but issue 3 is definitely an issue. I think the solution we have here isn't a bad one although our RuntimeHandle is becoming _very_ generic which makes it hard to reason about.

crates/kubelet/src/handle.rs
```diff
@@ -33,27 +34,36 @@ pub trait Stop {
     async fn wait(&mut self) -> anyhow::Result<()>;
 }
 
+/// Trait to describe necessary behavior for creating multiple log readers.
+/// TODO: Both providers make a handle containing a tempfile. If this is a common pattern,
+/// it might make sense to provide that implementation here. This would add `tempfile` as a
+/// dependency of `kubelet`.
+pub trait LogHandle<R>: Sync + Send {
+    /// Create new log reader.
+    fn output(&self) -> R;
```
rylev, line 43: Not sure `output` is the best name. In a way `R` here is really a handle and the type we're calling `LogHandle` is more like a handle dispenser, but I'm not sure how to make that more clear.

crates/kubelet/src/handle.rs
```diff
@@ -64,12 +74,22 @@ impl<S: Stop, R: AsyncRead + AsyncSeek + Unpin> RuntimeHandle<S, R> {
...
-        let bytes_written = self.output.read_to_end(buf).await?;
-        self.output.seek(SeekFrom::Start(0)).await?;
-        Ok(bytes_written)
+    /// Streams output from the running process into the given sender.
+    /// Optionally tails the output and/or continues to watch the file and stream changes.
+    pub(crate) async fn output<R>(
+        &mut self,
+        sender: LogSender,
+        tail: Option<usize>,
+        follow: bool,
```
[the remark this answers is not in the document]
kflansburg: I will bundle it with LogSender, rather than creating a new type.

crates/kubelet/src/logs.rs
```diff
@@ -0,0 +1,173 @@
...
+impl LogSender {
+    /// Create new `LogSender` from `hyper::body::Sender`.
+    pub fn new(sender: hyper::body::Sender) -> Self {
+        LogSender {
+            sender: Some(sender),
+        }
+    }
+
+    /// Async send some data to a client.
+    pub async fn send(&mut self, data: String) -> Result<(), LogSendError> {
```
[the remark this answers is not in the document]
kflansburg: I think ideally this method would take ownership and avoid the copy. Bytes actually implements From<String>, so I can use that.

crates/kubelet/src/logs.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+/// Future that streams logs from provided `AsyncRead` to provided `hyper::body::Sender`.
+pub async fn stream_logs<R: AsyncRead + std::marker::Unpin>(
+    output: R,
+    mut sender: LogSender,
+    tail: Option<usize>,
+    follow: bool,
+) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
kflansburg: Will do.

crates/kubelet/src/server.rs
kflansburg: Will fix.

crates/kubelet/src/logs.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Gracefully close the channel.
+    pub fn close(&mut self) {
+        match self.sender.take() {
+            Some(sender) => sender.abort(),
+            None => (),
+        }
+    }
+}
+
+/// Future that streams logs from provided `AsyncRead` to provided `hyper::body::Sender`.
```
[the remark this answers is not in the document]
kflansburg: Will do.

crates/kubelet/src/handle.rs
kflansburg: Could rename to `build` and LogHandle could be LogHandleBuilder.

crates/wasi-provider/Cargo.toml
kflansburg: Yep.

crates/kubelet/src/handle.rs
rylev: While I'm not sure it's the best name, I do think it's an improvement 😀

kflansburg: Went with Factory, builder didn't seem quite right since we aren't implementing the real builder pattern here.

crates/kubelet/src/handle.rs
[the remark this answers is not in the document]
thomastaylor312: Yeah I wasn't sure if it did or not. Docs seemed to indicate that it maybe did, but apparently it didn't

crates/kubelet/src/handle.rs
thomastaylor312: Ok, so let's keep this as is. We can revisit it later as needed

bacongobbler:
Hey @kflansburg! Let us know when this is ready for another round of reviews. :)

crates/kubelet/src/logs.rs
```diff
@@ -0,0 +1,179 @@
...
+                    error!("channel error: {}", e);
+                    LogSendError::Abnormal(anyhow::Error::new(e))
+                }
+            }),
+            None => Err(LogSendError::ChannelClosed),
+        }
+    }
+
+    /// Gracefully close the channel.
+    pub fn close(&mut self) {
```
thomastaylor312, line 83: Nit: Should this just be implemented as `Drop`?

thomastaylor312:
@kflansburg This should be good to merge once it is rebased (had another PR get merged right as I finished reviewing). No need to address my nit unless you feel like it

kflansburg:
Fixed the merge conflict. 
@bacongobbler I'm ready for further review / merge.
@thomastaylor312 Fixed your nit.

crates/kubelet/src/logs.rs
rylev: This should be the default implementation of drop. Do we need this function at all?

kflansburg: Done.

thomastaylor312:
Thanks again @kflansburg!

PR #232, opened 2020-05-05
docs(*): Various doc updates for 0.2

thomastaylor312:
This contains various and sundry doc updates and clarifications for
0.2. I fixed and/or added a few things while I went through the docs
and updated the GH action to push each crate separately.

This SHOULD NOT be merged until right before we cut 0.2

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,11 +53,12 @@ jobs:
     if: github.event_name == 'release'
     steps:
       - uses: actions/checkout@v2
-      - name: publish to crates.io
-        uses: actions-rs/cargo@v1
-        with:
-          command: publish
-          args: --token ${{ secrets.CargoToken }}
+      - name: publish kubelet to crates.io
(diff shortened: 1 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -53,11 +53,12 @@ jobs:
     if: github.event_name == 'release'
     steps:
       - uses: actions/checkout@v2
-      - name: publish to crates.io
-        uses: actions-rs/cargo@v1
```
thomastaylor312, line 57: You can't use `working-directory` with a `uses`, so I just went back to using the built in `cargo`

bacongobbler:
To prevent us from accidentally publishing the wasi-provider, wascc-logging and wascc-provider crates, would you mind adding the `publish = false` flag to their respective Cargo.tomls?

https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

thomastaylor312:
Oooh I was wondering if there was something like that. Done

PR #234, opened 2020-05-05
Don't watch files under ./target in VS Code

itowlson:
In the default Ubuntu configuration, VS Code displays a "too many files to watch" warning when opening the Krustlet directory.  Rather than updating a global setting that could consume a lot of memory, we can tell VS Code not to watch the `target` directory.


autodidaddict:
This has plagued me for months and I never knew how to fix it. Going to use this config in my projects! An absolute life saver.

Issue #50, opened 2020-03-09
Better path parsing and error handling for the kubelet server


thomastaylor312:
Right now we just use a basic match to route the requests. We should return things like "Method Not Allowed" if the wrong HTTP method is used and do some more checks for invalid paths

PR #235, opened 2020-05-05
Improve path parsing and error handling for kubelet API.

kflansburg:
Closes: #50

### DIFF
--- a/crates/kubelet/src/logs.rs
+++ b/crates/kubelet/src/logs.rs
@@ -40,13 +40,13 @@ impl std::error::Error for LogSendError {
 #[serde(rename_all = "camelCase")]
 /// Client options for fetching logs.
 pub struct LogOptions {
-    tail: Option<usize>,
+    tail_lines: Option<usize>,
     follow: Option<bool>,
 }
 
 impl LogOptions {
     pub fn tail(&self) -> Option<usize> {
-        self.tail
+        self.tail_lines
     }
 
     pub fn follow(&self) -> bool {
--- a/crates/kubelet/src/logs.rs
+++ b/crates/kubelet/src/logs.rs
@@ -1,5 +1,6 @@
 use anyhow::bail;
 use log::{debug, error};
+use serde::Deserialize;
 use tokio::io::{AsyncBufReadExt, AsyncRead};
 
 /// Possible errors sending log data.
@@ -35,31 +36,44 @@ impl std::error::Error for LogSendError {
     }
 }
 
-/// Sender for streaming logs to client.
-pub struct LogSender {
-    sender: hyper::body::Sender,
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Client options for fetching logs.
+pub struct LogOptions {
     tail: Option<usize>,
-    follow: bool,
+    follow: Option<bool>,
 }
 
-impl LogSender {
-    /// Create new `LogSender` from `hyper::body::Sender`.
-    pub fn new(sender: hyper::body::Sender, tail: Option<usize>, follow: bool) -> Self {
-        LogSender {
-            sender,
-            tail,
-            follow,
-        }
-    }
-
-    /// The tail flag indicated by the request if present.
+impl LogOptions {
     pub fn tail(&self) -> Option<usize> {
         self.tail
     }
 
+    pub fn follow(&self) -> bool {
+        self.follow.unwrap_or(false)
+    }
+}
+
+/// Sender for streaming logs to client.
+pub struct LogSender {
+    sender: hyper::body::Sender,
+    opts: LogOptions,
+}
+
+impl LogSender {
+    /// Create new `LogSender` from `hyper::body::Sender`.
+    pub fn new(sender: hyper::body::Sender, opts: LogOptions) -> Self {
+        LogSender { sender, opts }
+    }
+
+    /// The tail flag indicated by the request if present.
+    pub fn tail(&self) -> Option<usize> {
+        self.opts.tail()
+    }
+
     /// The follow flag indicated by the request, or `false` if absent.
     pub fn follow(&self) -> bool {
-        self.follow
+        self.opts.follow()
     }
 
     /// Async send some data to a client.
--- a/crates/kubelet/src/logs.rs
+++ b/crates/kubelet/src/logs.rs
@@ -37,21 +37,12 @@ impl std::error::Error for LogSendError {
 }
 
 #[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
 /// Client options for fetching logs.
 pub struct LogOptions {
-    tail_lines: Option<usize>,
-    follow: Option<bool>,
-}
-
-impl LogOptions {
-    pub fn tail(&self) -> Option<usize> {
-        self.tail_lines
-    }
-
-    pub fn follow(&self) -> bool {
-        self.follow.unwrap_or(false)
-    }
+    #[serde(rename = "tailLines")]
+    pub tail: Option<usize>,
+    #[serde(default)]
+    pub follow: bool,
 }
 
 /// Sender for streaming logs to client.
@@ -68,12 +59,12 @@ impl LogSender {
 
     /// The tail flag indicated by the request if present.
     pub fn tail(&self) -> Option<usize> {
-        self.opts.tail()
+        self.opts.tail
     }
 
     /// The follow flag indicated by the request, or `false` if absent.
     pub fn follow(&self) -> bool {
-        self.opts.follow()
+        self.opts.follow
     }
 
     /// Async send some data to a client.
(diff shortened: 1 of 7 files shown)


kflansburg:
A couple of notes:
* I wasn't sure which health checks / ping endpoints we want to support. 
* There could be value in going with one of the lightweight frameworks to handle routing and param parsing. 
* Does anyone know where the Kubelet API is documented?

thomastaylor312:
Apparently there is no documentation for Kubelet because "it is internal" (see https://github.com/kubernetes/kubernetes/issues/13470).

As for frameworks, is there one you had in mind? I don't think we are opposed to it if it will make the code significantly cleaner

crates/kubelet/src/server.rs
```diff
@@ -111,79 +118,143 @@ where
...
+    Ok(Response::builder().status(code).body(Body::from(body))?)
 }
 
 /// Get the logs from the running WASM module
 ///
 /// Implements the kubelet path /containerLogs/{namespace}/{pod}/{container}
 async fn get_container_logs<T: Provider + Sync>(
     provider: &T,
-    req: &Request<Body>,
+    _req: &Request<Body>,
```
thomastaylor312, line 212: Looks like we don't need this any more?

kflansburg:
I believe that [actix-web](https://github.com/actix/actix-web) would be pretty easy to drop in here and the routing macros it provides would clean up the code a lot here.

crates/kubelet/src/server.rs
kflansburg: Probably, I want sure if we wanted API consistency with `post_exec`, `get_container_logs` and any future service functions.

thomastaylor312: I think we are ok to remove, especially because this is a private function

thomastaylor312:
@kflansburg So I did some digging and there are some concerns around its complexity and its use of `unsafe`. @rylev suggested we look at [warp](https://github.com/seanmonstar/warp), which is built on hyper and Tokio, which are existing dependencies in the project.

Otherwise, if you'd rather not implement this, we can merge as is. Just let me know!

kflansburg:
Yeah, there was some drama a few months ago about that. Warp is an interesting project but I'm not sold on the ergonomics. It may be the best option though.

thomastaylor312:
Yeah, I think the primary thing is that warp builds on what we already have right now, so I think it would be the best option here. Does implementing that interest you or would you rather we just merge now?

kflansburg:
Yes, I will take a crack at it tonight. I think most async frameworks will build on tokio and hyper. Warp does seem to be the most actively maintained one though. Another option is `tower-web` but it seems to have less activity. I will give Warp a shot.

kflansburg:
@rylev @thomastaylor312 I'm having a little trouble with the logs response handler. Warp expects the return type to `impl warp::Reply`. The channel produces a `hyper::Body`. Warp implements `Reply` for `http::Response<T> where T: From<Body>`, but not `where T: Body` which is talked about a little here seanmonstar/warp#38 . So the obvious `http::Response<hyper::Body>` does not appear to work. I can implement a wrapper struct that implements `Reply` but that seems clunky. Any ideas?

kflansburg:
Ok, @thomastaylor312, I've completed the Warp implementation. A couple of notes:

* This moved to using the PEM TLS files rather than the output from `openssl`. It moves to using command line arguments to specifying the path to those file, and these match the arguments used by Kubernetes' Kubelet. Hopefully this is ok. I think some Docs may need to be updated to remove the openssl step (and dev dependency?). 
* Anything that doesn't match a path right now appears to return method not allowed. If we want more friendly responses ("please specify a container"), then those will probably have to be additional handlers. 

Let me know what you think.

kflansburg:
Looks like the e2e tests don't like the new API. Will look into it.

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,113 @@
...
+    _container: String,
+) -> Result<Response<Body>, Infallible> {
+    return_with_code(
+        StatusCode::NOT_IMPLEMENTED,
+        format!("Exec not implemented."),
+    )
+}
+
+fn return_with_code(code: StatusCode, body: String) -> Result<Response<Body>, Infallible> {
+    let mut response = Response::new(body.into());
```
rylev, line 105: Can we use a ResponseBuilder instead?

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,113 @@
...
+    *response.status_mut() = code;
+    Ok(response)
+}
+
+/// Return a simple status message
+fn get_ping() -> &'static str {
+    "this is the Krustlet HTTP server"
```
rylev, line 112: Let's use a `const` instead of a function

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,113 @@
...
+    _pod: String,
+    _container: String,
+) -> Result<Response<Body>, Infallible> {
+    return_with_code(
+        StatusCode::NOT_IMPLEMENTED,
+        format!("Exec not implemented."),
+    )
+}
+
+fn return_with_code(code: StatusCode, body: String) -> Result<Response<Body>, Infallible> {
```
rylev, line 104: Why return a result? So as to not have to wrap it in an `Ok` at the call site?

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,113 @@
...
+        .and(warp::query::<LogOptions>())
+        .and_then(move |namespace, pod, container, opts| {
+            let provider = logs_provider.clone();
+            get_container_logs(provider, namespace, pod, container, opts)
         });
-    }
 
-    Ok(())
-}
+    let exec_provider = provider.clone();
```
rylev, line 34: I don't believe we need this extra clone. Do we?

crates/kubelet/src/logs.rs
```diff
@@ -35,31 +36,44 @@ impl std::error::Error for LogSendError {
     }
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Client options for fetching logs.
+pub struct LogOptions {
+    tail_lines: Option<usize>,
+    follow: Option<bool>,
```
rylev, line 44: Can we annotate this as being false if not present on deserializing it? I'd rather just leave this a struct with public fields rather than have the getting methods

crates/kubelet/src/server.rs
kflansburg: Response builder returns a result. Warp doesn't seem to like failable functions, so this avoids that result.

crates/kubelet/src/server.rs
kflansburg: Will do.

crates/kubelet/src/server.rs
kflansburg: Yes, I wanted to match the return type of the calling functions. This is the type Warp is expecting back from the handler.

crates/kubelet/src/server.rs
kflansburg: I'm not sure if there is a cleaner way to do this, but both closures need to move a provider.

crates/kubelet/src/logs.rs
kflansburg: Ok, will do.

kflansburg:
The end to end tests pass on my local machine. I'm not sure what the issue is.

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,110 @@
...
-
+use log::{debug, error};
+use std::convert::Infallible;
 use std::sync::Arc;
+use warp::Filter;
 
-use crate::config::ServerConfig;
-use crate::logs::LogSender;
-use crate::provider::{NotImplementedError, Provider};
+const PING: &'static str = "this is the Krustlet HTTP server";
```
rylev, line 15: nit: `'static` isn't needed here

crates/kubelet/src/server.rs
```diff
@@ -1,189 +1,110 @@
...
         Err(e) => {
             error!("Error fetching logs: {}", e);
-            let mut res = Response::new(Body::from(format!("Server error: {}", e)));
-            *res.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
             if e.is::<NotImplementedError>() {
-                res = Response::new(Body::from("Not Implemented"));
-                *res.status_mut() = StatusCode::NOT_IMPLEMENTED;
+                return_with_code(
+                    StatusCode::NOT_IMPLEMENTED,
+                    format!("Logs not implemented in provider."),
```
rylev, line 79: No need to use the `format!` macro here. `String::from`, `.to_owned()` or `.to_string()`  would be preferable

kflansburg:
Fixed those comments. I'm ready to merge unless you guys want to add some handlers to improve the error codes returned (rather than METHOD_NOT_FOUND for everything which seems to be Warp's default).

thomastaylor312:
@kflansburg This looks amazing! Because we are going to cut 0.2 soon, we are going to hold off on merging until we do so. We'll also create some issues to capture doc updates when we do

thomastaylor312:
@kflansburg Now that we have 0.2 out we can merge this! Mind doing a rebase?

kflansburg:
Will do.

thomastaylor312:
Thanks @kflansburg!

PR #240, opened 2020-05-10
Allow Provider to Edit Node Definition.

kflansburg:
Not sure if this is something you all are interested in, in particular it allows for non-wasm providers .
 
* (unrelated) Make `stream_logs` available to providers that don't use Pod/RuntimeHandle.
* Create builder pattern for constructing node definitions.
* Implement basic defaults for node definition.
* Populate builder with calculated values and values from config.
* Give Provider an opportunity to edit the node definition.
* Move wasm-specific taint/architecture to example providers.

### DIFF
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -111,7 +111,7 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
 
     builder.set_port(config.server_config.port as i32);
 
-    provider.node(&mut builder);
+    provider.node(&mut builder).await;
     let node = builder.build().into_inner();
 
     match retry!(node_client.create(&PostParams::default(), &node).await, times: 4, break_on: &Error::Api(ErrorResponse { code: 409, .. }))
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -11,6 +11,8 @@ use log::{debug, error, info, warn};
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
+const KUBELET_VERSION: &'static str = env!("CARGO_PKG_VERSION");
+
 macro_rules! retry {
     ($action:expr, times: $num_times:expr, error: $on_err:expr) => {{
         let mut n = 0u8;
@@ -85,14 +87,14 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
     builder.add_capacity("hugepages-1Gi", "0");
     builder.add_capacity("hugepages-2Mi", "0");
     builder.add_capacity("memory", "4032800Ki");
-    builder.add_capacity("pods", "30");
+    builder.add_capacity("pods", &config.max_pods.to_string());
 
     builder.add_allocatable("cpu", "4");
     builder.add_allocatable("ephemeral-storage", "61255492Ki");
     builder.add_allocatable("hugepages-1Gi", "0");
     builder.add_allocatable("hugepages-2Mi", "0");
     builder.add_allocatable("memory", "4032800Ki");
-    builder.add_allocatable("pods", "30");
+    builder.add_allocatable("pods", &config.max_pods.to_string());
 
     let ts = Utc::now();
     builder.add_condition("Ready", "True", &ts, "KubeletReady", "kubelet is ready");
@@ -401,14 +403,16 @@ impl Node {
         Default::default()
     }
 
-    /// Create node definition from `k8s_openapi::api::core::v1::Node` object.
-    pub fn from_k8s_openapi(node: k8s_openapi::api::core::v1::Node) -> Node {
-        Node(node)
-    }
-
     /// Extract inner `k8s_openapi::api::core::v1::Node` object from node definition.
     pub fn into_inner(self) -> KubeNode {
         self.0
+    }
+}
+
+impl From<KubeNode> for Node {
+    /// Create node definition from `k8s_openapi::api::core::v1::Node` object.
+    fn from(node: KubeNode) -> Self {
+        Node(node)
     }
 }
 
@@ -590,8 +594,7 @@ impl Default for NodeBuilder {
             taints: vec![],
             architecture: "".to_string(),
             kube_proxy_version: "v1.17.0".to_string(),
-            // TODO Should this be the crate version?
-            kubelet_version: "v1.17.0".to_string(),
+            kubelet_version: KUBELET_VERSION.to_string(),
             container_runtime_version: "mvp".to_string(),
             operating_system: "linux".to_string(),
             capacity: BTreeMap::new(),
@@ -647,6 +650,7 @@ mod test {
             },
             data_dir: PathBuf::new(),
             node_labels,
+            max_pods: 110,
         };
 
         let mut builder = Node::builder();
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -111,7 +111,11 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
 
     builder.set_port(config.server_config.port as i32);
 
-    provider.node(&mut builder).await;
+    match provider.node(&mut builder).await {
+        Ok(()) => (),
+        Err(e) => warn!("Provider node annotation error: {:?}", e),
+    }
+
     let node = builder.build().into_inner();
 
     match retry!(node_client.create(&PostParams::default(), &node).await, times: 4, break_on: &Error::Api(ErrorResponse { code: 409, .. }))
--- a/crates/kubelet/src/node.rs
+++ b/crates/kubelet/src/node.rs
@@ -1,13 +1,15 @@
 use crate::config::Config;
+use crate::provider::Provider;
 use chrono::prelude::*;
 use k8s_openapi::api::coordination::v1::Lease;
-use k8s_openapi::api::core::v1::Node;
+use k8s_openapi::api::core::v1::Node as KubeNode;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::{Api, DeleteParams, PatchParams, PostParams};
 use kube::error::ErrorResponse;
 use kube::Error;
 use log::{debug, error, info, warn};
-use std::collections::HashMap;
+use std::collections::BTreeMap;
+use std::sync::Arc;
 
 macro_rules! retry {
     ($action:expr, times: $num_times:expr, error: $on_err:expr) => {{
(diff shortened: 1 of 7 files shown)


kflansburg:
Another alternative is to pass a mutable reference to the k8s_openapi Node instead if the builder pattern is too clunky here.

kflansburg:
I would also be in favor of adding more types to the kubelet::node module to increase the type safety of the builder methods (vs all Strings). I didn't want to overcomplicate the API though.

squillace:
Hi @kflansburg, can you elaborate on the purpose of this PR? Krustlet is not intended as a generic node; it is specifically for Wasm providers. What kinds of scenarios are you thinking about here? That will help us expand our thinking. :-)

kflansburg:
Hey @squillace, I don't see any reason to place that limitation on the API, but regardless it doesn't hurt to give providers the option of modifying these values, and customizing these values / moving away from the JSON definition to use `k8s_openapi::Node` has already been discussed in #45 .

squillace:
I actually like all the features! But my attention is drawn to `in particular it allows for non-wasm providers .` I'm fine with this, in a general sense. But I do not want anyone to be distracted from Wasm work because we now have the ability to use non-wasm providers. 

Now, all that said in a nice firm tone -- in truth, I'm not **opposed** to it, but I want to hear about specific ideas and use cases first. If we can't come up with one, then we can't be thinking about this. Make sense? If we got some, then it sounds like a great idea. :-) Lemme hear the ideas! I'm good on everything else here, and love the contribution! (I betchya @thomastaylor312 merges you no problem. :-) )

kflansburg:
Sure. I think there is value in the community being able to extend this work to other novel workloads, or traditional containers. This project does not have to implement these providers (I've been working on my own here https://github.com/kflansburg/krustlet-cri which implements the Kubernetes CRI to talk to containerd et. al). I understand if developers here want to stay focused on WASM, and I've tried to limit my changes to things that improve this API and not just address my downstream needs.

squillace:
Oh, great! You know that we did investigate implementing a CRI version for quite a while before falling back to a kubelet due to gRPC and container-laden overhead which we felt was an obstruction to the experiment, right? 

BUT: just to be clear, we do not take any particular position on what I'll call _the eventual community implementation_ of "wasi in k8s". For example, several contributors to containerd have chatted about doing that work (https://twitter.com/tiborvass/status/1248294294508756992?s=20) and we had started by having a look at https://github.com/dmcgowan/containerd-wasm. I've also chatted with Chris Crone of docker about where this functionality will end up, and he too was looking at it from that pov. If you're not already wired up with them, I'd be happy to put you in contact. Our main take away from that work was: "CRI is way too container-centric, and go+gRPC is not really the most productive language ecosystem for this work, despite it's obvious strengths."

I think it's possible the runtime functionality will end up being hosted elsewhere. :-) Don't let ANY of this convo here distract you from pursuing your project!!!!

kflansburg:
Interesting, I can see how CRI is too container centric. I've had a pretty good experience using Rust + gRPC using [tonic](https://docs.rs/tonic). My project is not WASM related, however, it is a Krustlet Provider implementation which talks to existing container runtimes over CRI. So far I've only implemented basic pod creation / deletion and logs.

squillace:
That sounds great. Feel free to ask the peeps here about CRI work and issues; I'm sure they'd be happy to help you along where they can. :-)

crates/kubelet/src/node.rs
```diff
@@ -58,11 +60,57 @@ macro_rules! retry {
...
 /// versions of Kubernetes prior to 1.14.
-pub async fn create_node(client: &kube::Client, config: &Config, arch: &str) {
-    let node_client: Api<Node> = Api::all(client.clone());
-    let node = node_definition(config, arch);
-    let node =
-        serde_json::from_value(node).expect("failed to deserialize node from node definition JSON");
+pub async fn create_node<P: 'static + Provider + Sync + Send>(
+    client: &kube::Client,
+    config: &Config,
+    provider: Arc<P>,
```
bacongobbler, line 66: this part is a litttle concerning to me. This puts a hard dependency on Providers in the `kubelet` crate. I believe @thomastaylor312 designed this crate to be non-krustlet specific. That way other libraries could consume this to create their own kubelets without the provider system.

Is there a way we can somehow split this out? I like the builder pattern used here. I'm just not sure if this is the right way to allow providers to inject their own information into the node definition.

What if there were two functions:

- `new` returns a NodeBuilder with all the defaults set
- `create` which makes the request to create the node (lines 115+)

What do you think about that approach?

thomastaylor312: I don't think I necessarily have a problem tying this to a `Provider`, but I think @bacongobbler has a point here. It would be better to just pass in a kube client and create the node from there instead of tying the implementation to a `Provider`. As far as I understand, creating a node should be the same for any `Provider` and it should be handled purely by logic external to any specific `Provider`

kflansburg: Hmm, perhaps I don't understand the distinction between `krustlet` and the `kubelet` crate. It seems to me that the Provider trait isn't specific to krustlet (I'm assuming you mean the concrete implementations in this repo), but is part of the API for implementing a kubelet, and is defined in the `kubelet` crate.  Provider is already a dependency of the method that calls `create_node`, is it important that `create_node` specifically can be used without a Provider implementation? 

I'm not sure I follow your last sentence, where would providers call `new` and `create`? The goal of this change is to give providers a hook before node registration which does not currently exist. I'm not sure how to add that hook without a dependency on Provider. 

I would support splitting this function into two parts for ergonomics though.

kflansburg: @thomastaylor312 I agree that logic for contacting the kube-apiserver is likely the same for each provider but I think construction of the node definition (especially taints) is going to be provider specific. I'm happy to split the function up.

crates/kubelet/src/node.rs
```diff
@@ -444,16 +387,233 @@ fn node_labels_definition(arch: &str, config: &Config) -> HashMap<String, String
...
+pub struct Node(k8s_openapi::api::core::v1::Node);
+
+impl Node {
+    /// Create builder for node definition.
+    pub fn builder() -> NodeBuilder {
+        Default::default()
+    }
+
+    /// Create node definition from `k8s_openapi::api::core::v1::Node` object.
+    pub fn from_k8s_openapi(node: k8s_openapi::api::core::v1::Node) -> Node {
```
thomastaylor312, line 405: This should probably be done with `impl From<KubeNode> for Node`

crates/kubelet/src/node.rs
```diff
@@ -444,16 +387,233 @@ fn node_labels_definition(arch: &str, config: &Config) -> HashMap<String, String
...
+    fn default() -> NodeBuilder {
+        NodeBuilder {
+            name: "krustlet".to_string(),
+            annotations: BTreeMap::new(),
+            labels: BTreeMap::new(),
+            pod_cidr: "10.244.0.0/24".to_string(),
+            taints: vec![],
+            architecture: "".to_string(),
+            kube_proxy_version: "v1.17.0".to_string(),
+            // TODO Should this be the crate version?
```
thomastaylor312, line 593: Yeah, this should probably be the crate version

crates/kubelet/src/node.rs
```diff
@@ -58,11 +60,57 @@ macro_rules! retry {
...
+
+    node_labels_definition(P::ARCH, &config, &mut builder);
+
+    // TODO Do we want to detect this?
+    builder.add_capacity("cpu", "4");
+    builder.add_capacity("ephemeral-storage", "61255492Ki");
+    builder.add_capacity("hugepages-1Gi", "0");
+    builder.add_capacity("hugepages-2Mi", "0");
+    builder.add_capacity("memory", "4032800Ki");
+    builder.add_capacity("pods", "30");
```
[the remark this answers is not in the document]
kflansburg: To be honest, I don't know how nodes determine this number without knowing what workloads are running.

crates/kubelet/src/node.rs
kflansburg: Agreed.

crates/kubelet/src/node.rs
kflansburg: Cool

crates/kubelet/src/node.rs
thomastaylor312: I think the main concern for me is it is a little weird to have the `Provider` with a `node` function that configures the builder as part of the trait definition. I think it would be better to use `T::ARCH` and pass that in as an argument to the builder itself. It just seems like an odd step to pass your builder to a provider rather than just getting the information you need from the provider

kflansburg: Hmmm, I see. I think the advantage of this is that Providers can completely introspect and even replace the Node object, using the common k8s_openapi type, or not implement the method at all. Using associated constants has some issues I think. It does not allow any dynamic content to be loaded into the node config from the provider (for instance querying the runtime for its version). It also requires `kubelet` to explicitly define what can be customized and limits providers to those fields, and possible results in a very fat trait.

Would it make more sense to have Node return a struct defined in `kubelet` which has fields for a limited set of overrides? That would at least avoid passing in the mutable reference.

crates/kubelet/src/node.rs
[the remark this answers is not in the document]
thomastaylor312: I think the normal kubelet allows you to set this limit with a command line flag

crates/kubelet/src/node.rs
thomastaylor312: Let's give this current design a whirl. If we don't like it, we don't have any API stability guarantees yet, so we can remove/refactor down the road. 

@bacongobbler you agree?

crates/kubelet/src/node.rs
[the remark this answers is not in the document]
kflansburg: Ah, yes. "--max-pods", default 110. Will do that.

crates/kubelet/src/node.rs
bacongobbler: I'm okay with the current approach for now. If we hear design concerns/feedback from others we can circle back on this.

crates/kubelet/src/provider.rs
```diff
@@ -50,7 +50,7 @@ pub trait Provider {
     const ARCH: &'static str;
 
     /// Allows provider to populate node information.
-    fn node(&self, _builder: &mut NodeBuilder) {}
+    async fn node(&self, _builder: &mut NodeBuilder) {}
```
rylev, line 53: By providing a default  implementation we're saying that usually people won't need to override this yet we currently override it in both providers. Are we sure we want to default to this implementation?

thomastaylor312: I actually think it is a good default for making this more generic for others. I think that other providers shouldn't necessarily need extra configuration. We can always remove it down the road

thomastaylor312:
Although there is a lot of code here, the feature is very narrow in scope, so I am ok with pulling this into 0.2

PR #246, opened 2020-05-12
Make Krustlet VM preemptible

DazWilkin:
Add a flag, save money.


bacongobbler:
This might not be worth documenting as the default. The goal of these guides are to provide best practices towards production-readiness (once krustlet is stable, of course).

Preemptible instances can be terminated at any time. If the user has only one Krustlet node (like as what's described in the documentation here), this would result in all of their WebAssembly workloads shutting down with no recourse. I'm not sure that's something we'd want to recommend to first-time users.

I could see this as a separate topic guide, however, on what pre-emptible VM instances are and how it affects your workload.

Let us know your thoughts.

bacongobbler:
https://cloud.google.com/compute/docs/instances/preemptible

PR #250, opened 2020-05-12
feat(*): Update dependencies to use rustls instead of openssl

### LINKED ISSUE #247: Use native TLS Rust implementation
The https://github.com/ctz/rustls crate is fairly stable this point and is a feature we can enable in reqwest, so we should be able to override our dependency on openssl, which would make windows builds much easier

thomastaylor312:
This makes it so we don't have to compile against specific versions
of openssl and might make Windows easier.

This also updates our kube and openapi dependencies

Closes #247

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 anyhow = "1.0"
 tokio = { version = "0.2", features = ["macros"] }
-kube = "0.33"
+kube = "0.34"
 env_logger = "0.7"
 kubelet = { path = "./crates/kubelet", version = "0.2", features = ["cli"] }
 wascc-provider = { path = "./crates/wascc-provider", version = "0.2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,15 +33,20 @@ exclude = [
 [badges]
 maintenance = { status = "actively-developed" }
 
+[features]
+default = ["native-tls"]
+native-tls = ["kube/native-tls", "kubelet/kube-native-tls", "wasi-provider/native-tls", "wascc-provider/native-tls", "oci-distribution/native-tls"]
+rustls-tls = ["kube/rustls-tls", "kubelet/rustls-tls", "wasi-provider/rustls-tls", "wascc-provider/rustls-tls", "oci-distribution/rustls-tls"]
+
 [dependencies]
 anyhow = "1.0"
 tokio = { version = "0.2", features = ["macros"] }
-kube = { version= "0.34", default-features = false, features = ["rustls-tls"] }
+kube = { version= "0.34", default-features = false }
 env_logger = "0.7"
-kubelet = { path = "./crates/kubelet", version = "0.2", features = ["cli"] }
-wascc-provider = { path = "./crates/wascc-provider", version = "0.2" }
-wasi-provider = { path = "./crates/wasi-provider", version = "0.2" }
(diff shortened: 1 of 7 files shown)


thomastaylor312:
Realized I still need to test a few things locally so moving to draft

bacongobbler:
How do we want to address the e2e failure?

```
Error: ReqwestError(reqwest::Error { kind: Request, url: "https://127.0.0.1:32768/api/v1/nodes/krustlet-wascc", source: hyper::Error(Connect, Custom { kind: Other, error: "invalid dnsname" }) })
thread 'test_wascc_provider' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6
```

thomastaylor312:
@bacongobbler That is the feature work I mentioned to you offline. Still working on it

thomastaylor312:
A couple of notes on this:

- This makes it really easy to build on windows once [this patch](https://github.com/clux/kube-rs/pull/240) gets merged upstream
- This _does not_ entire remove openssl from the equation. We are still using `native-tls` for the kubelet API. This means on linux it will still use the local openssl. I am going to create another issue to track the work for that. It will require that we switch to PEM only instead of pkcs12

thomastaylor312:
Currently holding off on this as TLS bootstrapping is going to require the openssl dependency any way to generate CSRs

crates/kubelet/src/server.rs
```diff
@@ -99,7 +99,7 @@ async fn post_exec<T: 'static + Provider + Send + Sync>(
 ) -> Result<Response<Body>, Infallible> {
     return_with_code(
         StatusCode::NOT_IMPLEMENTED,
-        format!("Exec not implemented."),
+        "Exec not implemented.".to_string(),
```
thomastaylor312, line 102: This was a clippy error I found while rebasing

thomastaylor312:
Ok, so I found a way using the `rcgen` crate to generate a CSR and private key without openssl, so once TLS bootstrapping hits, we can refactor to use that and have a completely openssl-less build. We will still want the features this PR adds so that people can still develop locally with kind/minikube (rustls doesn't currently support certs with IP SANs)

Cargo.toml
```diff
@@ -33,23 +33,28 @@ exclude = [
 [badges]
 maintenance = { status = "actively-developed" }
 
+[features]
+default = ["native-tls"]
```
bacongobbler, line 37: so to clarify here, `native-tls` refers to compiling krustlet with native OpenSSL. `rustls-tls` refers to the "native" (<-- where I get a little confused) Rust implementation of OpenSSL. Is that correct?

If we plan to support both features, should we document the two features in the developer docs? That might be helpful for other authors writing external providers.

thomastaylor312: Native-tls matches the naming of the rust crate a features from other dependencies. So you are correct in that it uses native TLS implementations (whether that is openssl or something else).

Right now, this doesn't change anything for people working on the project. It still builds exactly as it did before (the default being native-tls). I was going to hold off on documentation until we got the TLS bootstrapping in and write it along with the documentation and pipeline for Windows builds. This is mainly to avoid writing a bunch of "when we do windows builds" kind of comments. If you'd think we really need the docs right now, I can definitely add those in a follow up

bacongobbler: Thanks for the clarification. Let's hold off until TLS boostrapping is in ans we can re-assess how we want to address this in the documentation (as most may require changes after TLS bootstrapping lands, anyways).

PR #251, opened 2020-05-12
Clean up WASI test resources even if test fails

### LINKED ISSUE #206: Create automatic cleanup for tests
Right now if a test fails, cleanup of testing resources does not happen. We should either implement a "Guard" style resource that wraps the k8s resource with a `Drop` implementation or using a crate that allows for cleanup steps to be run

itowlson:
Fixes #206.


thomastaylor312:
Looks like @rylev might be busy. So going to merge this and we can refactor later

PR #253, opened 2020-05-13
introduce the wasm3 provider

bacongobbler:
This is still a work-in-progress, and much work is still required. Posting this for other's interest.


bacongobbler:
After discussing with @thomastaylor312 about this PR, we agreed that it would be easier to demonstrate this as a separate project. This will give others a reference point on how they can write their own providers.

PR #254, opened 2020-05-14
Graceful shutdown. 

kflansburg:
I wanted to invite comments on what I have been working on:

* Catch sigterm and certain errors and trigger graceful shutdown.
* Graceful shutdown cordons node and evicts pods, before deregistering it with the apiserver.
* Refactor kubelet futures for readability.
* Rework the combinators used for kubelet futures to enable graceful error handling.

This seems to do a good job of allowing Kubernetes to reschedule pods, rather than leaving zombie pods, and it gives the provider an opportunity to properly clean up pod resources as they are deleted. I'm not totally sure if all of this is desired behavior so I wanted to get some feedback before continuing.

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -104,8 +104,13 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
 
         // Create a queue that locks on events per pod
         let queue = PodQueue::new(self.provider.clone(), error_sender);
-        let pod_informer =
-            start_pod_informer::<T>(client.clone(), self.config.node_name.clone(), queue).fuse();
+        let pod_informer = start_pod_informer::<T>(
+            client.clone(),
+            self.config.node_name.clone(),
+            queue,
+            Arc::clone(&signal),
+        )
+        .fuse();
 
         // These must all be running for graceful shutdown. An error here exits ungracefully.
         let core = async {
@@ -156,6 +161,7 @@ async fn start_pod_informer<P: 'static + Provider + Sync + Send>(
     client: kube::Client,
     node_name: String,
     mut queue: PodQueue<P>,
+    signal: Arc<AtomicBool>,
 ) -> anyhow::Result<()> {
     let node_selector = format!("spec.nodeName={}", node_name);
     let params = ListParams {
@@ -177,6 +183,14 @@ async fn start_pod_informer<P: 'static + Provider + Sync + Send>(
             match stream.try_next().await {
                 Ok(Some(event)) => {
                     debug!("Handling Kubernetes pod event: {:?}", event);
+                    if let kube::api::WatchEvent::Added(_) = event {
+                        if signal.load(Ordering::Relaxed) {
+                            warn!(
+                                "Node is shutting down and unschedulable. Dropping Add Pod event."
+                            );
+                            continue;
+                        }
+                    }
                     match queue.enqueue(event).await {
                         Ok(()) => debug!("Enqueued event for processing"),
                         Err(e) => warn!("Error enqueuing pod event: {}", e),
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -1,7 +1,7 @@
 ///! This library contains code for running a kubelet. Use this to create a new
 ///! Kubelet with a specific handler (called a `Provider`)
 use crate::config::Config;
-use crate::node::{create_node, delete_node, drain_node, update_node};
+use crate::node::{create_node, drain_node, update_node};
 use crate::queue::PodQueue;
 use crate::server::start_webserver;
 use crate::status::{update_pod_status, Phase};
@@ -17,6 +17,7 @@ use kube::{
 use log::{debug, error, warn};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use tokio::signal::ctrl_c;
 use tokio::sync::mpsc;
 
 /// A Kubelet server backed by a given `Provider`.
@@ -60,7 +61,7 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
 
         // Flag to indicate graceful shutdown has started.
         let signal = Arc::new(AtomicBool::new(false));
-        signal_hook::flag::register(signal_hook::SIGINT, Arc::clone(&signal))?;
+        let signal_task = start_signal_task(Arc::clone(&signal)).fuse();
 
         // Start the webserver
         let webserver = start_webserver(self.provider.clone(), &self.config.server_config).fuse();
@@ -70,19 +71,19 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
         let error_handler = start_error_handler(error_receiver, client.clone()).fuse();
 
         // Start updating the node lease periodically
-        let node_updater = start_node_updater(
-            Arc::clone(&signal),
-            client.clone(),
-            self.config.node_name.clone(),
-        )
-        .fuse();
+        let node_updater = start_node_updater(client.clone(), self.config.node_name.clone()).fuse();
 
         // If any of these tasks fail, we can initiate graceful shutdown.
         let services = async {
-            futures::pin_mut!(webserver, error_handler, node_updater);
+            futures::pin_mut!(webserver, error_handler, node_updater, signal_task);
 
             futures::select! {
-                res = error_handler => error!("Error handler task completed with result: {:?}", &res),
+                res = error_handler => if let Err(e) = res {
+                    error!("Error handler task completed with error: {:?}", &e);
+                },
+                res = signal_task => if let Err(e) = res {
+                    error!("Signal task completed with error {:?}", &e);
+                },
                 res = webserver => error!("Webserver task completed with result {:?}", &res),
                 res = node_updater => if let Err(e) = res {
                     error!("Node updater task completed with error {:?}", &e);
@@ -122,8 +123,8 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
         };
 
         // Services will not return an error, so this will wait for both to return, or core to
-        // return an error. Services will return if signal is set because node_updater checks this
-        // and will exit.
+        // return an error. Services will return if signal is set because pod_informer will drop
+        // error_sender and error_handler will exit.
         futures::try_join!(core, services)?;
         Ok(())
     }
@@ -141,6 +142,14 @@ impl<P> Clone for Kubelet<P> {
     }
 }
 
+/// Awaits SIGINT and sets graceful shutdown flag if detected.
+async fn start_signal_task(signal: Arc<AtomicBool>) -> anyhow::Result<()> {
+    ctrl_c().await?;
+    warn!("Caught keyboard interrupt.");
+    signal.store(true, Ordering::Relaxed);
+    Ok(())
+}
+
 /// Listens for updates to pods on this node and forwards them to queue.
 async fn start_pod_informer<P: 'static + Provider + Sync + Send>(
     client: kube::Client,
@@ -155,31 +164,37 @@ async fn start_pod_informer<P: 'static + Provider + Sync + Send>(
     let api = Api::<KubePod>::all(client);
     let informer = Informer::new(api).params(params);
     loop {
-        // TODO Should these results be ignored if Err? It seems reasonable to just keep trying,
-        // especially since this function is critical for a graceful shutdown.
-        let mut stream = informer.poll().await?.boxed();
-        while let Some(event) = stream.try_next().await? {
-            debug!("Handling Kubernetes pod event: {:?}", event);
-            match queue.enqueue(event).await {
-                Ok(()) => debug!("Enqueued event for processing"),
-                Err(e) => warn!("Error enqueuing pod event: {}", e),
-            };
+        let mut stream = match informer.poll().await {
+            Ok(stream) => stream.boxed(),
+            Err(e) => {
+                warn!("Error polling pod informer: {:?}", e);
+                tokio::time::delay_for(std::time::Duration::from_secs(5)).await;
+                continue;
+            }
+        };
+        loop {
+            match stream.try_next().await {
+                Ok(Some(event)) => {
+                    debug!("Handling Kubernetes pod event: {:?}", event);
+                    match queue.enqueue(event).await {
+                        Ok(()) => debug!("Enqueued event for processing"),
+                        Err(e) => warn!("Error enqueuing pod event: {}", e),
+                    };
+                }
+                Ok(None) => break,
+                Err(e) => warn!("Error streaming pod events: {:?}", e),
+            }
         }
     }
 }
 
 /// Periodically renew node lease. Exits if signal is caught.
-async fn start_node_updater(
-    signal: Arc<AtomicBool>,
-    client: kube::Client,
-    node_name: String,
-) -> anyhow::Result<()> {
+async fn start_node_updater(client: kube::Client, node_name: String) -> anyhow::Result<()> {
     let sleep_interval = std::time::Duration::from_secs(10);
-    while !signal.load(Ordering::Relaxed) {
+    loop {
         update_node(&client, &node_name).await;
         tokio::time::delay_for(sleep_interval).await;
     }
-    Ok(())
 }
 
 /// Checks for shutdown signal and cleans up resources gracefully.
@@ -193,7 +208,6 @@ async fn start_signal_handler(
         if signal.load(Ordering::Relaxed) {
             warn!("Signal caught.");
             drain_node(&client, &node_name).await?;
-            delete_node(&client, &node_name).await?;
             break Ok(());
         }
         tokio::time::delay_for(duration).await;
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -14,7 +14,7 @@ use kube::{
     runtime::Informer,
     Api,
 };
-use log::{debug, error, warn};
+use log::{debug, error, info, warn};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use tokio::signal::ctrl_c;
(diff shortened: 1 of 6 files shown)


thomastaylor312:
I think we are completely open to this @kflansburg. The reason we hadn't focused on it was that the WASM providers don't leave zombie pods when they restart as they are all running within the krustlet process. I'll go leave comments in the code (if I have any) on the approaches you are taking.

Also, going to mark this as a draft for now just so we know when it is ready for a final review by pulling it out of a draft state

crates/kubelet/src/kubelet.rs
```diff
@@ -54,98 +54,77 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
...
         let client = kube::Client::new(self.kube_config.clone());
+
         // Create the node. If it already exists, "adopt" the node definition
         create_node(&client, &self.config, self.provider.clone()).await;
 
-        // Get the node name for use in the update loop
-        let node_name = self.config.node_name.clone();
+        // Flag to indicate graceful shutdown has started.
+        let signal = Arc::new(AtomicBool::new(false));
+        signal_hook::flag::register(signal_hook::SIGINT, Arc::clone(&signal))?;
```
thomastaylor312, line 63: Since we are alerting only on SIGINT, we should probably just leverage `tokio::signal::ctrl_c`

crates/kubelet/src/kubelet.rs
```diff
@@ -162,6 +141,107 @@ impl<P> Clone for Kubelet<P> {
...
+) -> anyhow::Result<()> {
+    let node_selector = format!("spec.nodeName={}", node_name);
+    let params = ListParams {
+        field_selector: Some(node_selector),
+        ..Default::default()
+    };
+    let api = Api::<KubePod>::all(client);
+    let informer = Informer::new(api).params(params);
+    loop {
+        // TODO Should these results be ignored if Err? It seems reasonable to just keep trying,
```
thomastaylor312, line 158: I agree here. Just log the error and let it keep trying so this can handle the graceful shutdown

kflansburg:
Sounds good, while there may be nothing running after the WASM provider shuts down, the node remains registered with kube-apiserver, and pods remain with a "Running" state. Maybe zombie isn't the best term for this.

crates/kubelet/src/node.rs
```diff
@@ -166,6 +167,124 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+        field_selector: Some(node_selector),
+        ..Default::default()
+    };
+    let kube::api::ObjectList { items: pods, .. } = pod_client.list(&params).await?;
+
+    warn!("Evicting {} pods.", pods.len());
+
+    // TODO: Filter mirror pods,daemonsets, kube-system?
+    for pod in &pods {
+        if let Some(name) = pod.metadata.as_ref().and_then(|meta| meta.name.as_ref()) {
```
thomastaylor312, line 211: You can bring the `Meta` trait into scope to avoid all of this unwrapping. Also, it may be worth it to convert all of these to our internal `Pod` wrapper to get convenient accessors on this data

crates/kubelet/src/node.rs
```diff
@@ -166,6 +167,124 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+
+async fn evict_pod(client: &kube::Client, name: &str, namespace: &str) -> anyhow::Result<()> {
+    let ns_client: Api<Pod> = Api::namespaced(client.clone(), namespace);
+    let lp = ListParams::default()
+        .fields(&format!("metadata.name={}", name))
+        .fields(&format!("metadata.namespace={}", namespace));
+
+    // The delete call may return a "pending" response, we must watch for the actual delete event.
+    // TODO Timeout?
+    let mut stream = ns_client.watch(&lp, "0").await?.boxed();
```
thomastaylor312, line 239: I'm not sure if it matters much in this PR, but I thought I'd bring it up: We might want to check if there is any performance overhead for spinning up a separate watch for each pod. If there are 100s of pods on the node, I don't know what the API Server does with the list params and what the overhead is for opening a stream for each individual pod. If it is a problem, we might want to see if there is a good way to share the same watch

crates/kubelet/src/kubelet.rs
```diff
@@ -162,6 +141,107 @@ impl<P> Clone for Kubelet<P> {
...
+    signal: Arc<AtomicBool>,
+    client: kube::Client,
+    node_name: String,
+) -> anyhow::Result<()> {
+    let duration = std::time::Duration::from_millis(100);
+    loop {
+        if signal.load(Ordering::Relaxed) {
+            warn!("Signal caught.");
+            drain_node(&client, &node_name).await?;
+            delete_node(&client, &node_name).await?;
```
thomastaylor312, line 196: If I remember correctly, the normal kubelet doesn't deregister itself from the API. I think cordoning would be ok, but deleting probably isn't needed

crates/kubelet/src/kubelet.rs
kflansburg: I've used ctrl_c in the past, but we may wish to support other signals such as SIGTERM, if, for instance, we wish to run Krustlet inside of a container.

crates/kubelet/src/kubelet.rs
kflansburg: Will do.

crates/kubelet/src/node.rs
kflansburg: Yeah, I was on the fence about using the `Pod` wrapper, I will try it out.

crates/kubelet/src/node.rs
kflansburg: True, I'm not sure how `kubectl` [does it](https://kubernetes.io/images/docs/kubectl_drain.svg), I don't think its very sophisticated. I think it would be possible to do with a single `Stream` so I can give that a try.

crates/kubelet/src/kubelet.rs
kflansburg: Hmmm, ok. I'm fine with this but I think then I would like to look at the re-registration functionality because I have been having issues with that.

crates/kubelet/src/kubelet.rs
thomastaylor312: Yeah, I figure we could, although windows support for that is a bit hairy.

Also, we may want to use the `tokio-support` feature to convert this into a `Stream`

crates/kubelet/src/node.rs
thomastaylor312: If it becomes too complex, don't sweat it, we can just leave a comment and revisit later

crates/kubelet/src/kubelet.rs
thomastaylor312: So right now the re-registration is a hack because we don't have TLS bootstrapping and can't update the node status directly. Another community member is almost done with that (we are waiting to cut 0.2 for it), so this should be greatly improved soon-ish

crates/kubelet/src/kubelet.rs
kflansburg: Ah, I see.

crates/kubelet/src/kubelet.rs
kflansburg: Ok, sounds good.

crates/kubelet/src/node.rs
```diff
@@ -166,6 +168,120 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+    let params = ListParams {
+        field_selector: Some(node_selector),
+        ..Default::default()
+    };
+    let kube::api::ObjectList { items: pods, .. } = pod_client.list(&params).await?;
+
+    let lp = ListParams::default().fields(&format!("spec.nodeName={}", node_name));
+
+    // The delete call may return a "pending" response, we must watch for the actual delete event.
+    // TODO I dont know what the version number does here.
```
thomastaylor312, line 211: This is the resource version of the object in Kubernetes. From the docs:
> When specified with a watch call, shows changes that occur after that particular version of a resource. Defaults to changes from the beginning of history. When specified for list: - if unset, then the result is returned from remote storage based on quorum-read flag; - if it's 0, then we simply return what we currently have in cache, no guarantee; - if set to non zero, then the result is at least as fresh as given rv.

crates/kubelet/src/node.rs
```diff
@@ -166,6 +168,120 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+    let mut stream = pod_client.watch(&lp, "0").await?.boxed();
+
+    warn!("Evicting {} pods.", pods.len());
+
+    // TODO: Filter mirror pods,daemonsets, kube-system?
```
thomastaylor312, line 216: For feature parity with what `kubectl drain` does, it should _technically_ error if it can't remove these kinds of pods (except for mirror, those are ignored). However, we probably want the `--ignore-daemonset` equivalent and don't care if we error here anyway.

tl;dr: Ignore mirror and daemonsets. Kube-system should be evicted if they are not daemonsets

thomastaylor312:
@kflansburg This is looking close! Feel free to mark as ready for review when you'd like me to review

crates/kubelet/src/node.rs
kflansburg: Ok, I will give this a shot. In my testing, kube-proxy gets rescheduled onto the node and can be left as a zombie. Do you think it would be reasonable to have `pod_informer` drop add pod events after shutdown has started to prevent this scheduling?

kflansburg:
Ok, I have added filtering for static pods and daemon sets. For static pods, I still patch the status of the pod as terminated. I think this is most informative to consumers of the kube-api and it causes the pod to be rescheduled on the node when it is restarted.

crates/kubelet/src/kubelet.rs
```diff
@@ -162,6 +142,120 @@ impl<P> Clone for Kubelet<P> {
     }
 }
 
+/// Awaits SIGINT and sets graceful shutdown flag if detected.
+async fn start_signal_task(signal: Arc<AtomicBool>) -> anyhow::Result<()> {
+    ctrl_c().await?;
+    warn!("Caught keyboard interrupt.");
+    signal.store(true, Ordering::Relaxed);
```
rylev, line 149: nit: it'd be good to have a comment here on ordering and why `Relaxed` is the right one.

crates/kubelet/src/node.rs
```diff
@@ -166,6 +169,142 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
     info!("Successfully created node '{}'", &config.node_name);
 }
 
+/// Fetch the uid of a node by name.
+pub async fn node_uid(client: &kube::Client, node_name: &str) -> anyhow::Result<String> {
+    let node_client: Api<KubeNode> = Api::all(client.clone());
+    match retry!(node_client.get(node_name).await, times: 4, log_error: |e| error!("Failed to get node to cordon: {:?}", e))
+    {
+        Ok(node) => match node.metadata.and_then(|meta| meta.uid) {
```
rylev, line 177: nit can the `node` be matched on inline inside of the `Ok`? 
`Ok(KubeNode { metadata: Some(Metadata { uid: Some(uid) }`

crates/kubelet/src/node.rs
```diff
@@ -166,6 +169,142 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+    // The delete call may return a "pending" response, we must watch for the actual delete event.
+    let mut stream = pod_client.watch(&lp, "0").await?.boxed();
+
+    warn!("Evicting {} pods.", pods.len());
+
+    for pod in pods {
+        let pod = Pod::new(pod);
+        if pod.is_daemonset() {
+            warn!("Skipping eviction of DaemonSet '{}'", pod.name());
+            continue;
```
rylev, line 220: nit: why `if` with `continue` instead of `if { } else if {}`?

crates/kubelet/src/node.rs
```diff
@@ -166,6 +169,142 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+                    if (name == pod.name()) & (namespace == pod.namespace()) {
+                        warn!("Pod '{}' evicted.", name);
+                        break;
+                    }
+                }
+                _ => (),
+            }
+        }
+    } else {
+        warn!("Pod '{}' evicted.", name);
```
rylev, line 287: Should all these logs be `debug` not `warn`?

crates/kubelet/src/node.rs
```diff
@@ -166,6 +169,142 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+    let response = ns_client.delete(name, &params).await?;
+
+    if response.is_left() {
+        // TODO Timeout?
+        warn!("Waiting for pod '{}' eviction.", name);
+        while let Some(event) = stream.try_next().await? {
+            match event {
+                kube::api::WatchEvent::Deleted(s) => {
+                    let pod = Pod::new(s);
+                    if (name == pod.name()) & (namespace == pod.namespace()) {
```
rylev, line 278: Did you mean to do a bitwise `and` here?

crates/kubelet/src/node.rs
```diff
@@ -166,6 +169,142 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+        }
+    } else {
+        warn!("Pod '{}' evicted.", name);
+    }
+    Ok(())
+}
+
+/// Mark this node as unschedulable.
+pub async fn cordon_node(client: &kube::Client, node_name: &str) -> anyhow::Result<()> {
+    warn!("Cordoning node.");
```
rylev, line 294: nit: log name of node

crates/kubelet/src/pod.rs
```diff
@@ -84,6 +84,26 @@ impl Pod {
...
+    /// dont know how reliable this is.
+    pub fn is_static(&self) -> bool {
+        self.0.meta().owner_references.is_none()
+    }
+
+    /// Indicate if this pod is part of a Daemonset
+    pub fn is_daemonset(&self) -> bool {
+        if let Some(owners) = &self.0.meta().owner_references {
+            for owner in owners {
+                match owner.kind.as_ref() {
```
rylev, line 98: nit: use `if let` instead of match here

kflansburg: Just curious, what makes one better than the other?

crates/kubelet/src/node.rs
kflansburg: I don't follow. Isn't that a logical and of two `bool`s?

crates/kubelet/src/node.rs
kflansburg: It started out as just a one line check and was more readable that way. I will change it.

crates/kubelet/src/kubelet.rs
kflansburg: Well, my reasoning isn't very specific. We aren't trying to synchronize anything here. Once the flag is set, other tasks discover this with best effort and we don't really need any hard requirements on when this happens.

kflansburg: I think this could also be implemented with a Mutex<bool> or something and ordering would be a non-issue.

crates/kubelet/src/node.rs
thomastaylor312: If you can find a nice way to let the pod informer know, I think that is a good idea

crates/kubelet/src/node.rs
thomastaylor312: I think `info` is the right level here

crates/kubelet/src/node.rs
thomastaylor312: This is a bitwise operation rather than a boolean operation `&` vs `&&`

kflansburg: I don't believe this distinction applies to `bool`. If you want to remove the parens and compare the expressions logically, I can do that.

crates/kubelet/src/node.rs
kflansburg: ok, will do

crates/kubelet/src/node.rs
thomastaylor312: I think the end result is the same, but the bitwise operator might have other performance/side effects I don't know about. I'll leave it to @rylev to clarify

crates/kubelet/src/pod.rs
rylev: It’s just a bit more terse when only one pattern is being handled. It’s the more idiomatic thing to do.

rylev: It’s just a bit more terse when only one pattern is being handled. It’s the more idiomatic thing to do.

rylev: It’s just a bit more terse when only one pattern is being handled. It’s the more idiomatic thing to do.

crates/kubelet/src/node.rs
rylev: I don’t believe it makes a difference runtime wise. It’s just very surprising to see bite wise and. Seems clever when it doesn’t have to be

thomastaylor312:
@kflansburg You ready for final review on this?

kflansburg:
@thomastaylor312 Yep, ready for review.

crates/kubelet/src/node.rs
```diff
@@ -166,6 +170,140 @@ pub async fn create_node<P: 'static + Provider + Sync + Send>(
...
+    namespace: &str,
+    stream: &mut PodStream,
+) -> anyhow::Result<()> {
+    let ns_client: Api<KubePod> = Api::namespaced(client.clone(), namespace);
+    info!("Evicting namespace '{}' pod '{}'", namespace, name);
+    let params = Default::default();
+    let response = ns_client.delete(name, &params).await?;
+
+    if response.is_left() {
+        // TODO Timeout?
```
thomastaylor312, line 274: If you want to, this can be done with the `retry!` macro in this file. Not going to block on it though

bacongobbler:
Thanks @kflansburg! Let's get this in and follow up with future PRs if you'd like to address @thomastaylor312's comment. Great work here.

PR #255, opened 2020-05-14
Use `token` for image registry auth. 

peterhuene:
According to the [Docker registry documentation](https://docs.docker.com/registry/spec/auth/token/#requesting-a-token),
`token` is the preferred field and `access_token` is used for OAuth 2
compatibility.

This commit changes the expected auth response to be `token` with an alias of
`access_token` to support both.

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -326,12 +326,13 @@ impl ClientProtocol {
 /// A token granted during the OAuth2-like workflow for OCI registries.
 #[derive(serde::Deserialize, Default)]
 struct RegistryToken {
-    access_token: String,
+    #[serde(alias = "access_token")]
+    token: String,
 }
 
 impl RegistryToken {
     fn bearer_token(&self) -> String {
-        format!("Bearer {}", self.access_token)
+        format!("Bearer {}", self.token)
     }
 }
 
@@ -412,7 +413,7 @@ mod test {
 
         let tok = c.token.expect("token is available");
         // We test that the token is longer than a minimal hash.
-        assert!(tok.access_token.len() > 64);
+        assert!(tok.token.len() > 64);
     }
 
     #[tokio::test]
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -13,6 +13,7 @@ use futures_util::stream::StreamExt;
 use hyperx::header::Header;
 use log::debug;
 use reqwest::header::HeaderMap;
+use std::collections::HashMap;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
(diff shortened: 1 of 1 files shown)


peterhuene:
GCP container registries appear to use `token`, so without this the pull failed trying to deserialize the JSON response for authing with the registry.

peterhuene:
I'll look into the e2e failures tomorrow.

thomastaylor312:
The fix looks nice and simple. I am guessing that the e2e failures are due to issues pulling the module

bacongobbler:
Hey @peterhuene. Thanks for looking into this.

Just to be on the safe side, have you checked and confirmed if all other cloud providers follow the same behaviour? I want to make sure that things will continue to work for non-GKE users.

peterhuene:
I haven't yet tested it with ACR since I don't have access to the public repo feature yet (my cluster is not currently on Azure).

This will need a bit more testing before being ready, so I'm going to transition this to WIP for now.

DazWilkin:
Aha! Presumably this explains #257 ?

thomastaylor312:
@peterhuene any update on this? Need any help from us?

peterhuene:
This fell off my radar.  I'll see if I can get some time soon to move this forward.

thomastaylor312:
No rush, just wanted to make sure you weren't waiting on us for something

peterhuene:
Sorry for the continued delay to this PR.  I have a spare afternoon today to get this green and tested.

peterhuene:
Ok I've finally figured out why this PR is failing the e2e tests 😓.  It turns out the fix revealed both a bug and a design limitation in the OCI support.

The bug is that Krustlet attempts to pull the image for `kube-proxy`, which is part of a daemon set and thus scheduled on the Krustlet node.  The image for `kube-proxy` is `k8s.gcr.io/kube-proxy:<version>`.  Without this PR, the deserialization of the authentication token fails (Google Container Registry returns `token` and not `access_token`) and we fail the `kube-proxy` pod with a message saying we failed to deserialize the auth request.  With the fix, we successfully auth, but then fail the pod because the image manifest isn't what we expect for a Wasm image.  We need to figure out a reliable way to ignore daemon set pods that aren't Wasm workloads and thus entirely skip pulling their images.

This then brings us to the design limitation of `oci_distribution::Client`: it only supports auth against a single registry.  With these changes, we successfully auth with GCR as a result of attempting to pull the image for `kube-proxy`.  When the E2E tests run, the providers are asked to pull an image from Azure Container Registry.  The client uses the GCR token to auth with ACR, which then fails.  Thus, the test pods are immediately set into a failure state and the E2E tests fail.

To fix this limitation, we should be keeping a mapping between registry and registry token.

peterhuene:
I'll see about implementing multiple registry support in the OCI client to get this PR green.

thomastaylor312:
Great investigation work on this @peterhuene! I'll try circle back to the system pods problem again this next week. Is that causing the PR to fail though or still just causing noise?

peterhuene:
Adding support for authing with multiple registries would be all that's needed to get this green, so we can address the issue with pulling kube-proxy's image at a later time.

Issue #257, opened 2020-05-14
add imagePullSecrets support


DazWilkin:
I'm receiving this error from Krustlet trying to pull images from a public (!) GCR registry.

As shown below, `wasm-to-oci` can push|pull images to this registry and I've configured it to support public (non-authenticated) access. I'm able to run regular containers from it on the cluster without problem.

```
kubectl describe pod/hello-world-wasi-golang
Name:         hello-world-wasi-golang
Namespace:    default
Priority:     0
Node:         krustlet/
Labels:       <none>
Annotations:  Status:  Failed
Message:      Failed to decode registry token from auth request
IP:           
IPs:          <none>
Containers:
  hello-world-wasi-golang:
    Image:      gcr.io/${PROJECT}/hello-world-wasi-golang:v1
    Port:       <none>
    Host Port:  <none>
    Environment:
      FOO:             bar
      POD_NAME:        hello-world-wasi-golang (v1:metadata.name)
      CONFIG_MAP_VAL:  <set to the key 'myval' of config map 'hello-world-wasi-golang'>  Optional: false
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-x75t7 (ro)
Conditions:
  Type           Status
  PodScheduled   True 
Volumes:
  default-token-x75t7:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-x75t7
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     krustlet/arch=wasm32-wasi:NoExecute
                 node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type    Reason     Age        From               Message
  ----    ------     ----       ----               -------
  Normal  Scheduled  <unknown>  default-scheduler  Successfully assigned default/hello-world-wasi-golang to krustlet
```
I'm able to pull regular containers from the same registry to the same cluster.

```bash
PROJECT=....

docker pull hello-world
docker tag hello-world gcr.io/${PROJECT}/hello-world
docker push gcr.io/${PROJECT}/hello-world

kubectl run hello-world --image=gcr.io/${PROJECT}/hello-world

kubectl logs pod/hello-world

Hello from Docker!
...
```

I'm able to `wasm-to-oci [push|pull]` images to the registry:

```bash
go run ./cmd push \
  hello-world-wasi-golang \
[...]

PR #259, opened 2020-05-14
Golang WASI "hello world" example

DazWilkin:
A Golang equivalent of the Rust WASI "hello world" example.

Blocked by #257


DazWilkin:
On second thoughts, I think this won't work; Golang does not yet support WASI. Sorry, will withdraw.

PR #261, opened 2020-05-18
fix(*): Makes crate publish happen on tags

thomastaylor312:
Also removes an errant field in the kubelet Cargo.toml and fixes some ordering
issues

### DIFF
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -14,7 +14,6 @@ edition = "2018"
 license-file = "../../LICENSE"
 description = "A Kubernetes kubelet implementation in Rust"
 repository = "https://github.com/deislabs/krustlet"
-readme = "README.md"
 keywords = [
     "wasm",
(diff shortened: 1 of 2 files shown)


crates/kubelet/Cargo.toml
```diff
@@ -14,7 +14,6 @@ edition = "2018"
 license-file = "../../LICENSE"
 description = "A Kubernetes kubelet implementation in Rust"
 repository = "https://github.com/deislabs/krustlet"
-readme = "README.md"
```
radu-matei, line 17: #232 introduced this field in other crates as well - should it be removed from there too?

thomastaylor312: We aren't publishing them, so it isn't a big deal, but I'll go remove them

thomastaylor312: Actually, just checked. It does need to be there in oci-distribution and none of the rest have them

PR #264, opened 2020-05-19
add docs.rs links for kubelet and oci-distribution

bacongobbler:


bacongobbler:
added and merging after CI goes green

PR #265, opened 2020-05-20
fix(*): Updates justfile and config to work with new cert setup

thomastaylor312:
A recent change switched things back to a separate cert and key. However
the `just bootstrap` target was not updated to reflect this. I also noticed
some inconsistencies with how we were setting the default file path for
the cert and key and corrected those.

This also updates all of the docs to reflect the change

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build and Test
 on:
   pull_request:
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
 jobs:
   build:
     runs-on: ${{ matrix.config.os }}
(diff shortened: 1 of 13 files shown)


thomastaylor312:
@peterhuene if you have some time, I'd appreciate it if you could check that my EKS changes look correct. Things seem simple enough, but EKS does things differently than the other instructions, so I wanted to double check

thomastaylor312:
I'm not entirely sure how that would work. Would we want to test every single flag and then add a new test for any other flag? That seems more of a burden than a help

bacongobbler:
I'll mess around with the code and see if I can come up with a PR to test it. Shouldn't block this from going through however

PR #267, opened 2020-05-20
add cross-compile target config

bacongobbler:
This is necessary for compiling krustlet-wasi and krustlet-wascc for 64-bit ARM devices (and eventually 32-bit ARM as well, though that is currently not supported).

TODO:

- [ ] add documentation for cross-compiling


bacongobbler:
relies on #266.

bacongobbler:
will add docs as a follow-up.

PR #270, opened 2020-05-25
Read settings from config file

### LINKED ISSUE #202: support feature flag input from a config file
Similar to `~/.kube/config`, we should allow users to write krustlet config to a file and read it upon startup.

Once complete, we should remove `--pfx-password` and `$PFX_PASSWORD` as acceptable flag inputs and rely only on the TTY prompt or from the config file as input.

itowlson:
Fixes #202.

With this PR, the kubelet configuration can be read from a config file as well as from the CLI.  If an option is specified in both the CLI and the config file, the CLI takes precedence.

The initial version of this PR is rather prototype-y and the following things remain to be refined:

* ~Agree better names for the config file keys~ I have now agreed these. With myself.
* ~Agree default config file path~ Ditto
* How to override default config file path e.g. environment variable and/or command line flag
* Default config file should probably be YAML rather than JSON though we hates them all precious
* ~Improve error messages for invalid config file entries~ They're not going to win any prizes but they're less worse than they were I think
* Agree on-error behaviour - currently panics on bad config value unless overridden on CLI
* Should `config.rs` be broken up?  It has grown quite large...
* ~Documentation (not so much needing to be refined as, er, created)~ I maded some

Something to consider:

* Should we, instead of dealing with this specific case, instead build a standalone crate that bridges `config-rs` and `structopt` and if so can I call it `frankenconfig` please?

### DIFF
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -56,6 +56,7 @@ oci-distribution = { path = "../oci-distribution", version = "0.1", default-feat
 url = "2.1"
 warp = { version = "0.2", features = ['tls'] }
 http = "0.2"
+# TODO: select specific features
 config_file = { package = "config", version = "0.10.0" }
 
 [dev-dependencies]
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -56,6 +56,7 @@ oci-distribution = { path = "../oci-distribution", version = "0.1", default-feat
 url = "2.1"
 warp = { version = "0.2", features = ['tls'] }
 http = "0.2"
+config_file = { package = "config", version = "0.10.0" }
 
 [dev-dependencies]
 reqwest = {version = "0.10", default-features = false }
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -56,8 +56,6 @@ oci-distribution = { path = "../oci-distribution", version = "0.1", default-feat
 url = "2.1"
 warp = { version = "0.2", features = ['tls'] }
 http = "0.2"
-# TODO: select specific features
-config_file = { package = "config", version = "0.10.0" }
 
 [dev-dependencies]
 reqwest = {version = "0.10", default-features = false }
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -3,6 +3,7 @@
 //! The best way to configure the kubelet is by using [`Config::default_config`]
 //! or by turning on the "cli" feature and using [`Config::new_from_flags`].
 
+use std::iter::FromIterator;
 use std::net::IpAddr;
 use std::net::ToSocketAddrs;
 use std::path::PathBuf;
@@ -53,6 +54,32 @@ pub struct ServerConfig {
     pub tls_private_key_file: PathBuf,
 }
 
+#[derive(Clone, Debug, Default)]
+struct ConfigBuilder {
+    // TODO: consider if these should all be plain strings so we don't have to
+    // parse them until we are ready to build (relevant for overrides in the
+    // case where a value is invalid). Otherwise perhaps they should be Results
+    // rather than Options.
+    pub node_ip: Option<IpAddr>,
+    pub hostname: Option<String>,
+    pub node_name: Option<String>,
+    pub data_dir: Option<PathBuf>,
+    pub node_labels: Option<HashMap<String, String>>,
+    pub max_pods: Option<u16>,
+    pub server_addr: Option<IpAddr>,
+    pub server_port: Option<u16>,
+    pub server_tls_cert_file: Option<PathBuf>,
+    pub server_tls_private_key_file: Option<PathBuf>,
+}
+
+struct ConfigBuilderFallbacks {
+    hostname: fn() -> String,
+    data_dir: fn() -> PathBuf,
+    cert_path: fn(data_dir: &PathBuf) -> PathBuf,
+    key_path: fn(data_dir: &PathBuf) -> PathBuf,
+    node_ip: fn(hostname: &mut String, preferred_ip_family: &IpAddr) -> IpAddr,
+}
+
 impl Config {
     /// Returns a Config object set with all of the defaults.
     ///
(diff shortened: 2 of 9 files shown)


crates/kubelet/Cargo.toml
```diff
@@ -49,6 +49,7 @@ oci-distribution = { path = "../oci-distribution", version = "0.1.0" }
 url = "2.1"
 warp = { version = "0.2", features = ['tls'] }
 http = "0.2"
+config_file = { package = "config", version = "0.10.0" }
```
thomastaylor312, line 52: We should probably turn off the default features and pull in only the formats we want to support so as to avoid any extra dependencies

crates/kubelet/src/config.rs
```diff
@@ -85,77 +122,240 @@ impl Config {
...
     /// Parses all command line flags and sets the proper defaults. The version
     /// of your application should be passed to set the proper version for the CLI
     #[cfg(any(feature = "cli", feature = "docs"))]
     #[cfg_attr(feature = "docs", doc(cfg(feature = "cli")))]
-    pub fn new_from_flags(version: &str) -> Self {
-        // TODO: Support config files too. config-rs and clap don't just work
-        // together, so there is no easy way to merge together everything right
-        // now. This function is here so we can do that data massaging and
-        // merging down the road
+    pub fn new_from_flags_only(version: &str) -> Self {
```
thomastaylor312, line 149: Nit: I don't think the `only` is needed in the function name here or in `new_from_file`

crates/kubelet/src/config.rs
```diff
@@ -85,77 +122,240 @@ impl Config {
...
+        ConfigBuilder::from_config_source(config_source)
+    }
 
-        let data_dir = opts
-            .data_dir
-            .unwrap_or_else(|| default_data_dir().expect("unable to get default directory"));
+    fn from_config_source<T>(source: T) -> anyhow::Result<ConfigBuilder>
+    where
+        T: 'static,
+        T: config_file::Source + Send + Sync,
```
thomastaylor312, line 249: Can this line not be combined with the one above?

crates/kubelet/src/config.rs
```diff
@@ -85,77 +122,240 @@ impl Config {
...
+    {
+        let mut settings = config_file::Config::default();
+        match settings.merge(source) {
+            Ok(s) => Ok(ConfigBuilder::from_config_settings(s.clone())),
+            Err(e) => Err(anyhow::Error::new(e)),
+        }
+    }
+
+    fn from_config_settings(settings: config_file::Config) -> ConfigBuilder {
+        let port = settings
```
thomastaylor312, line 259: Did you by chance look at using [`try_into`](https://docs.rs/config/0.10.1/config/struct.Config.html#method.try_into) and deserializing it to a concrete data structure instead of having to do all these lookups?

crates/kubelet/src/config.rs
```diff
@@ -53,6 +54,42 @@ pub struct ServerConfig {
...
+const NODE_NAME_CONFIG_KEY: &str = "node_name";
+const DATA_DIR_CONFIG_KEY: &str = "data_dir";
+const NODE_LABELS_CONFIG_KEY: &str = "node_labels";
+const MAX_PODS_CONFIG_KEY: &str = "max_pods";
+const SERVER_ADDR_CONFIG_KEY: &str = "addr";
+const SERVER_PORT_CONFIG_KEY: &str = "port";
+const SERVER_TLS_CERT_FILE_CONFIG_KEY: &str = "tls_cert_file";
+const SERVER_TLS_PRIVATE_KEY_FILE_CONFIG_KEY: &str = "tls_private_key_file";
+
+struct ConfigBuilderFallbacks {
```
thomastaylor312, line 85: This struct feels janky...trying to think of a way to do it differently.

thomastaylor312:
Also, just realized I forgot to answer this question:

> Should we, instead of dealing with this specific case, instead build a standalone crate that bridges config-rs and structopt and if so can I call it frankenconfig please?

If you feel like doing the work for this, I think it would be really valuable not only for us, but probably for others. We could push it up to crates.io and let everyone use it

crates/kubelet/src/config.rs
itowlson: I hadn't - thanks!  This seems like it should be the way to go, but `try_into` requires the target type to implement `Deserialize`.  And I can't derive `Deserialize` on ConfigBuilder because serde doesn't know how to handle the `Option<Result<T>>` traits, and if I go back to straight `Option<T>` then I will get a different behaviour, namely that _any_ invalid value in the config file renders the entire config file unusable, even if that value is overridden on the command line.  Maybe that's an acceptable trade-off (maybe it's even desirable).

Of course I could implement `Deserialize` manually but then I'd be back to doing all the lookups so there's probably not much value in that!

itowlson: Or maybe there are attributes I can apply to the `<Option<Result<T>>` fields that would get the desired behaviour...

crates/kubelet/Cargo.toml
itowlson: Cool; let's do that once we agree on the desired format(s).

crates/kubelet/src/config.rs
thomastaylor312: Let’s give it a try and see what it looks like

crates/kubelet/src/config.rs
thomastaylor312: There are some attributes I think. An option result is a bit weird but custom deserialization isn’t too bad if I remember correctly from doing it myself

crates/kubelet/src/config.rs
itowlson: Right now I feel the completer trait looks messy.  Introducing a completer object increases duplication, and requires either an additional type or an unexpressed invariant.  I'm finding myself getting lost in the code, and I'm not seeing any clarify or consolidation benefit.

I think I have a sense for in what way this struct feels 'janky' and I do sympathise but right now it feels more pragmatic than the completer trait.  Do you want me to continue exploring alternatives, or could we put this aside for now?

thomastaylor312: Let's put it aside for now. Pinging @rylev on this one for any additional thoughts

crates/kubelet/src/config.rs
itowlson: BY THE HOARY HOSTS OF HOGGOTH @thomastaylor312 YOU ARE RIGHT.  Works out significantly nicer this way.

crates/kubelet/Cargo.toml
itowlson: Removed in favour of plain ol' serde anyway

PR #273, opened 2020-06-03
feat(kubelet): TLS bootstrapping support

### LINKED ISSUE #150: Add TLS bootstrapping to kubelet crate
Currently there is no way to update node status so it can show as being `Ready` after being in a state of `NotReady`. This turns out to be due to improper permissions for our kubeconfigs with use of the `NodeRestriction` admission plugin (which is enabled by default in at least Minikube, but probably others as well). We need to bind the [`system:node` role to a service account](https://kubernetes.io/docs/reference/access-authn-authz/node/) and possibly do [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/) (though that can be broken out into a separate task if it isn't required to get things working)

thomastaylor312:
Adds automatic TLS bootstrapping support for Kubelet. This includes bootstrapping for the serving certificates and for authorization with the Kubelet. Due to the nature
of how we set up providers, this has to be done outside of the main kubelet
start logic. Hopefully in future refactors, this can be streamlined so users
don't have call `bootstrap` themselves.

As part of this I also removed the cordoning step from the graceful shutdown.
Cordoning does not appear to happen on a normal kubelet shutdown and it makes
us have to implement the logic about uncordoning it on startup (which could
be an undesired behavior if the user has cordoned it for a reason).

Huge thanks to @DerekStrickland for the original effort in getting this most of the way there. His commits are included along with this PR

Closes #150


thomastaylor312:
NOTE: The code is pretty much ready for review (minus any fixes for tests). Marked this as a draft because I haven't updated docs. However, there will be _a bunch_ of doc changes, so if we want those in a separate PR, I can do that

crates/kubelet/src/config.rs
```diff
@@ -253,6 +260,14 @@ pub struct Opts {
         help = "The data path (logs, container images, etc) for krustlet storage. Defaults to $HOME/.krustlet"
     )]
     data_dir: Option<PathBuf>,
+
+    #[structopt(
+        long = "bootstrap-file",
+        env = "KRUSTLET_BOOTSTRAP_FILE",
+        help = "The path to the bootstrap config",
+        default_value = "/etc/kubernetes/bootstrap-kubelet.conf"
```
bacongobbler, line 268: Should this reference BOOTSTRAP_FILE? That way we're not maintaining the same string in two locations

bacongobbler:
Okay. I've gone through the code and it looks good. Let me know when this is ready to take for a spin and I'll test this.

crates/kubelet/src/config.rs
thomastaylor312: Let me see if that works. I've had problems using variables in these before

thomastaylor312:
@bacongobbler This is ready to go. I have explicitly left out documentation updates for the different k8s distributions because I didn't want to confuse users of 0.2. If you are ok with that, I'll create an issue to make sure we update those before cutting 0.3

bacongobbler:
LGTM! I've gone ahead and rebased this over at #275. Once that passes CI, I'll go ahead and merge.

Thank you @DerekStrickland and @thomastaylor312 for all your work!

PR #279, opened 2020-06-09
Pass container arguments to WASI module

### LINKED ISSUE #258: add container arguments support
```bash
echo `
apiVersion: v1
kind: ConfigMap
metadata:
  name: hello-world-wasi-rust
data:
  myval: "cool stuff"
---
apiVersion: v1
kind: Pod
metadata:
  name: hello-world-wasi-rust
spec:
  containers:
    - name: hello-world-wasi-rust
      image: webassembly.azurecr.io/hello-world-wasi-rust:v0.1.0
      args:
        - arg1
        - arg2
      env:
        - name: FOO
          value: bar
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: CONFIG_MAP_VAL
          valueFrom:
            configMapKeyRef:
              key: myval
              name: hello-world-wasi-rust
  tolerations:
    - key: "krustlet/arch"
      operator: "Equal"
      value: "wasm32-wasi"
      effect: "NoExecute"
' | kubectl apply --filename=-
```
Then 

```bash
kubectl logs pod/hello-world-wasi-rust
hello from stdout!
hello from stderr!
CONFIG_MAP_VAL=cool stuff
FOO=bar
POD_NAME=hello-world-wasi-rust
Args are: []
```

If I run the sample locally, it works as expected:

```
cargo run arg1 arg2
   Compiling hello-world-rust v0.1.0 (/tmp/krustlet/demos/wasi/hello-world-rust)
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/hello-world-rust arg1 arg2`
hello from stdout!
hello from stderr!
CARGO=...
CARGO_HOME=...
CARGO_MANIFEST_DIR=/tmp/krustlet/demos/wasi/hello-world-rust
CARGO_PKG_AUTHORS=Taylor Thomas <taylor.thomas@microsoft.com>
...
Args are: ["target/debug/hello-world-rust", "arg1", "arg2"]

```

itowlson:
Fixes #258.

### DIFF
--- a/crates/wasi-provider/src/lib.rs
+++ b/crates/wasi-provider/src/lib.rs
@@ -110,6 +110,7 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
         info!("Starting containers for pod {:?}", pod_name);
         for container in pod.containers() {
             let env = Self::env_vars(&container, &pod, &client).await;
+            let args = container.args.clone().unwrap_or_default();
             let module_data = modules
                 .remove(&container.name)
                 .expect("FATAL ERROR: module map not properly populated");
@@ -142,7 +143,7 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
             let runtime = WasiRuntime::new(
                 module_data,
                 env,
-                Vec::default(),
+                args,
                 container_volumes,
                 self.log_path.clone(),
             )
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -174,19 +174,23 @@ async fn verify_wasi_node(node: Node) -> () {
     );
 }
 
+const SIMPLE_WASI_POD: &str = "hello-wasi";
+const VERBOSE_WASI_POD: &str = "hello-world-verbose";
+
 async fn create_wasi_pod(client: kube::Client, pods: &Api<Pod>) -> anyhow::Result<()> {
+    let pod_name = SIMPLE_WASI_POD;
     // Create a temp directory to use for the host path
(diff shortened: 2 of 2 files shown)


crates/wasi-provider/src/lib.rs
```diff
@@ -110,6 +110,7 @@ impl<S: ModuleStore + Send + Sync> Provider for WasiProvider<S> {
         info!("Starting containers for pod {:?}", pod_name);
         for container in pod.containers() {
             let env = Self::env_vars(&container, &pod, &client).await;
+            let args = container.args.as_ref().unwrap_or(&Vec::default()).to_vec();
```
thomastaylor312, line 113: I would recommend just cloning `container.args` and then doing `unwrap_or_default` since you eventually end up cloning anyway with `to_vec`. So `container.args.clone().unwrap_or_default()`

tests/integration_tests.rs
```diff
@@ -242,10 +246,51 @@ async fn create_wasi_pod(client: kube::Client, pods: &Api<Pod>) -> anyhow::Resul
 
     assert_eq!(pod.status.unwrap().phase.unwrap(), "Pending");
 
+    wait_for_pod(client, pod_name).await
+}
+
+async fn create_fancy_schmancy_wasi_pod(
```
thomastaylor312, line 252: 👏

crates/wasi-provider/src/lib.rs
itowlson: I knew there must be a better way.  Thank you!

PR #280, opened 2020-06-09
fix(docs): use /dev/urandom for generating token secrets

### LINKED ISSUE #276: hack/bootstrap.sh blocks on WSL2
Oddly, calling `just run-wasi` on WSL2 hangs indefinitely. After doing a bit of digging, I found that it was because `hack/bootstrap.sh` blocks on WSL2.

Changing this line:

```
token_secret=$(< /dev/random tr -dc a-z0-9 | head -c${1:-16};echo;)
```

To use `/dev/urandom` instead appears to work:

```
token_secret=$(< /dev/urandom tr -dc a-z0-9 | head -c${1:-16};echo;)
```

It looks as though reading from `/dev/random` blocks indefinitely.

bacongobbler:
As these are hack scripts to help spin up a development environment, we should lean towards usability over security.

WSL does not have enough entropy available to generate a random seed with /dev/random.
Using /dev/urandom is less cryptographically secure, but it enables users on WSL to utilize the hack script.

closes #276

### DIFF
--- a/docs/howto/assets/bootstrap.sh
+++ b/docs/howto/assets/bootstrap.sh
@@ -4,8 +4,7 @@ set -euo pipefail
 export LC_CTYPE=C
 
 token_id="$(</dev/urandom tr -dc a-z0-9 | head -c "${1:-6}";echo;)"
-# This needs to use /dev/random so it is cryptographically safe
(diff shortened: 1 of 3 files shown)


bacongobbler:
Today I also learned https://www.2uo.de/myths-about-urandom/

Issue #263, opened 2020-05-19
Fix cargo publishing in release pipeline


thomastaylor312:
Right now there are two issues with the cargo publish part of the timeline:

1. It isn't blocked on the asset build/upload succeeding
2. If we don't bump the version of the `oci-distribution` crate but do bump the `kubelet` crate, the `cargo publish` command fails. We should check to see if it is already published and ignore the error

bacongobbler:
the first one has been addressed, but not the second. Removing from 0.3 as we don't need to address this release

PR #285, opened 2020-06-10
fix(ci): add dependency on publish action

bacongobbler:
This way the `crates` job won't start until krustlet is built and uploaded to Azure.

addresses the first point in #263


itowlson:
Should the documentation be updated to refer to the `v` tag prefix for creating a release?

thomastaylor312:
@itowlson Do we even have documentation on releasing yet? I think our developer documentation is the weak spot right now

bacongobbler:
We don't, but that is a great suggestion. We should have that documented. I'll open a new ticket to document the release process and we can add that there.

PR #292, opened 2020-06-16
refactor kubelet module

bacongobbler:
This PR shifts things around the kubelet package to make it easier to work with as a Provider.

The TL;DR of this PR:

In order to reduce redundancy, structures like `LogSender` or `ContainerStatus` were re-named to `log::Sender` or `container::Status`.

The `RuntimeHandle` and `PodHandle` structs were renamed as `container::Handle` and `pod::Handle`, respectively. `Stopper` was renamed to `StopHandler` to clarify the difference between the trait (`handle::StopHandler`) and the implementation (`pod::Handle`).

Some packages like the `PodQueue` now have reduced package visibilty, as no provider currently requires a PodQueue. This is an internal package for the `Kubelet`, and as such it was switched to `pub(crate)` visibility.

`module_store` was moved to module `store::oci`. The trait was moved to the `store` module. This way alternative stores for other protocols (like [WAPM](https://medium.com/wasmer/announcing-wapm-the-webassembly-package-manager-18d52fae0eea) or npm) can be implemented with relative ease

Note that I noticed a few changes could be made to the `node` and `volume` module (specifically `node::create()` and `Ref::volumes_from_pod`) but refrained from doing any further changes in this PR other than shifting things around. This is just to shift things around into more composable packages.


crates/kubelet/src/container/status.rs
```diff
@@ -89,50 +75,10 @@ impl ContainerStatus {
             // handle probes, this should be updated accordingly
             started: Some(true),
             // The rest of the items in status (see docs here:
-            // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#containerstatus-v1-core)
+            // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#status-v1-core)
```
thomastaylor312, line 78: I think this still needs to be `containerstatus`

crates/kubelet/src/handle/mod.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here re: removing

crates/kubelet/src/image/mod.rs
```diff
@@ -0,0 +1,3 @@
+#![allow(missing_docs)]
+
+pub mod client;
```
thomastaylor312, line 3: This is probably too nested. I think bringing up the client into here is probably sufficient

crates/kubelet/src/image/mod.rs
```diff
@@ -0,0 +1,3 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here

crates/kubelet/src/kubeconfig.rs
```diff
@@ -0,0 +1,25 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Likewise

crates/kubelet/src/lib.rs
```diff
@@ -47,25 +47,20 @@
...
+pub mod store;
+pub mod volume;
 
 pub use self::kubelet::Kubelet;
 pub use bootstrapping::bootstrap;
-pub use handle::{LogHandleFactory, PodHandle, RuntimeHandle};
-pub use logs::{stream_logs, LogSendError, LogSender};
-pub use node::NodeBuilder;
-pub use pod::Pod;
-#[doc(inline)]
```
thomastaylor312, line 70: Should we still inline the docs from `Provider` since it is such a critical component?

crates/kubelet/src/log/mod.rs
```diff
@@ -1,60 +1,62 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Likewise

crates/kubelet/src/log/mod.rs
```diff
@@ -168,3 +170,12 @@ pub async fn stream_logs<R: AsyncRead + std::marker::Unpin>(
 
     Ok(())
 }
+
+/// Trait to describe necessary behavior for creating multiple log readers.
+/// TODO: Both providers make a handle containing a tempfile. If this is a common pattern,
```
thomastaylor312, line 175: This TODO should use `//` so it doesn't get generated into the docs

crates/kubelet/src/node/mod.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Likewise

crates/kubelet/src/node/mod.rs
```diff
@@ -0,0 +1,5 @@
...
+
+mod node;
+
+pub use node::{create, drain, evict_pods, update, Builder, Node};
```
thomastaylor312, line 5: For my own edification, this is to make the publicly exported node functions more clear (e.g. make them stand out from the huge file), correct?

crates/kubelet/src/pod/handle.rs
```diff
@@ -1,129 +1,42 @@
-//! A convenience handle type for providers
```
thomastaylor312, line 1: Do we still need these docs?

crates/kubelet/src/pod/mod.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here

crates/kubelet/src/store/mod.rs
```diff
@@ -0,0 +1,6 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here

crates/kubelet/src/volume/mod.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here

crates/kubelet/src/webserver/mod.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(missing_docs)]
```
thomastaylor312, line 1: Same here

crates/wascc-provider/src/lib.rs
```diff
@@ -77,16 +80,16 @@ const VOLUME_DIR: &str = "volumes";
 /// Kubernetes' view of environment variables is an unordered map of string to string.
 type EnvVars = std::collections::HashMap<String, String>;
 
-/// A [kubelet::handle::Stop] implementation for a wascc actor
-pub struct ActorStopper {
+/// A [kubelet::handle::handle] implementation for a wascc actor
```
thomastaylor312, line 83: I think this is supposed to be `Handle` (capitalized)

crates/kubelet/src/container/status.rs
bacongobbler: Good catch. Must’ve been a blanket replace all and I missed that replacement

crates/kubelet/src/node/mod.rs
bacongobbler: Bingo

Splitting them out into separate files makes it simpler to split out work into separate modules in case we need to implement further node API calls in the future (node resource stats, for example?)

bacongobbler:
okay. I'll take a crack at writing some package documentation for these.

crates/kubelet/src/lib.rs
bacongobbler: The provider's now part of a separate module: `kubelet::provider::Provider`. I'll leave it up to you whether it makes sense to inline the provider documentation as part of module `kubelet`.

PR #296, opened 2020-06-19
Image pull policy

### LINKED ISSUE #256: FileModuleStore should respect imagePullPolicy=Always, 'latest' tag, or no tag
Currently `FileModuleStore` caches modules pulled from a registry.

However, it does not force a pull when `imagePullPolicy` is set to `Always` or when there is no `imagePullPolicy` specified and the image tag is `latest` or there is no tag.

[According to the documentation](https://kubernetes.io/docs/concepts/configuration/overview/#container-images), this should force a pull of the image from the repository.

We should also respect the other defaults for the image pull policy as specified in the documentation.

itowlson:
Fixes #256.

This implements the Kubernetes `imagePullPolicy` behaviour as specified at https://kubernetes.io/docs/concepts/configuration/overview/#container-images.


crates/kubelet/src/module_store.rs
```diff
@@ -12,6 +12,40 @@ use std::convert::TryFrom;
...
+    /// present in the local store
+    IfNotPresent,
+    /// Never fetch the module from the upstream registry; if it is not
+    /// available locally then return an error
+    Never,
+}
+
+impl ModulePullPolicy {
+    /// Parses a module pull policy from a Kubernetes ImagePullPolicy string
+    pub fn parse(name: Option<String>) -> anyhow::Result<Option<ModulePullPolicy>> {
```
thomastaylor312, line 32: I believe `IfNotPresent` is the default. Should we implement `Default` for `ModulePullPolicy` and then have this function just return `Result<ModulePullPolicy>`?

crates/kubelet/src/module_store.rs
```diff
@@ -29,20 +63,85 @@ use std::sync::Arc;
...
+    /// It is up to the implementation to establish caching policies.
+    /// However, the implementation must fail if the image is not present
+    /// locally.
+    async fn get_local(&self, image_ref: &Reference) -> anyhow::Result<Vec<u8>>;
+
+    /// Whether the specified module is already present in the store.
+    async fn is_present(&self, image_ref: &Reference) -> bool;
+
+    /// Whether the specified module is already present in the store with the specified digest.
+    async fn is_present_with_digest(&self, image_ref: &Reference, digest: String) -> bool;
```
thomastaylor312, line 103: Just spitballing here: Because this store is k8s specific, would it be better to abstract away all these functions so we don't have to manage implementing and calling each of these separately? We can provide the default implementation here at least?

I dunno, maybe it wouldn't work.

crates/kubelet/src/module_store.rs
```diff
@@ -97,39 +202,83 @@ impl<C> FileModuleStore<C> {
...
         path
     }
 
     fn pull_file_path(&self, r: &Reference) -> PathBuf {
         self.pull_path(r).join("module.wasm")
     }
 
-    async fn store(&self, image_ref: &Reference, contents: &[u8]) -> anyhow::Result<()> {
+    fn digest_file_path(&self, r: &Reference) -> PathBuf {
+        self.pull_path(r).join("digest.txt")
```
thomastaylor312, line 214: Isn't the digest some sort of SHA? Can we recalculate it instead of storing locally? I feel like you could do some weird fiddling if you can modify the digest file

crates/kubelet/src/image_client.rs
```diff
@@ -10,33 +10,63 @@ use oci_distribution::Reference;
 /// An image client capable of fetching images from a storage location
 #[async_trait]
 pub trait ImageClient {
```
radu-matei, line 12: Not related to this PR necessarily - but the naming here is still a bit misleading, and can lead people to think these are actually images.

crates/kubelet/src/module_store.rs
```diff
@@ -12,6 +12,40 @@ use std::convert::TryFrom;
...
+    /// Parses a module pull policy from a Kubernetes ImagePullPolicy string
+    pub fn parse(name: Option<String>) -> anyhow::Result<Option<ModulePullPolicy>> {
+        match name {
+            None => Ok(None),
+            Some(n) => ModulePullPolicy::parse_str(&n[..]),
+        }
+    }
+
+    fn parse_str(name: &str) -> anyhow::Result<Option<ModulePullPolicy>> {
+        match name {
```
radu-matei, line 40: Should this be case sensitive?
i.e. - if I try to set the pull policy as "always", I get an error:

```
---- module_store::test::can_parse_pull_policies stdout ----
thread 'module_store::test::can_parse_pull_policies' panicked at 'called `Result::unwrap()` on an `Err` value: unrecognized pull policy always', crates/kubelet/src/note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

thomastaylor312: I think kubernetes expects capitalized

radu-matei: Just checked the Kubelet code, you are right - https://github.com/kubernetes/kubernetes/blob/9cd906e9322f75ce48b47613dd9e025c2b06c143/pkg/kubelet/images/image_manager.go#L65-L76

crates/kubelet/src/image_client.rs
```diff
@@ -10,33 +10,63 @@ use oci_distribution::Reference;
...
-    ///         Ok(image.clone())
+    ///         Ok((image.clone(), Some(digest.to_string())))
     ///     }
     /// }
     /// ```
-    async fn pull(&mut self, image: &Reference) -> anyhow::Result<Vec<u8>>;
+    async fn pull_with_digest(
+        &mut self,
+        image_ref: &Reference,
+    ) -> anyhow::Result<(Vec<u8>, Option<String>)>;
```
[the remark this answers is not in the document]
itowlson: Yeah, it should be its own type I think, this is kind of "make it work first" code that I got lazy and talked myself into not tidying up.  Thanks for keeping me honest.

crates/kubelet/src/image_client.rs
```diff
@@ -10,33 +10,63 @@ use oci_distribution::Reference;
...
-    ///             .get(image)
+    ///             .get(image_ref)
     ///             .ok_or(anyhow::anyhow!("Couldn't find image"))?;
-    ///         Ok(image.clone())
+    ///         Ok((image.clone(), Some(digest.to_string())))
     ///     }
     /// }
     /// ```
-    async fn pull(&mut self, image: &Reference) -> anyhow::Result<Vec<u8>>;
+    async fn pull_with_digest(
```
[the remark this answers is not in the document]
itowlson: You are probably right that we can just have `pull` return the content plus digest.  I did this because I didn't know how widely the existing `pull` was used.

crates/kubelet/src/module_store.rs
itowlson: If only it were that simple!  The default is `IfNotPresent` _if_ the reference contains a tag _and_ the tag is not `:latest`; otherwise it is `Always`.  (https://kubernetes.io/docs/concepts/configuration/overview/#container-images)  Hence the shenanigans to compute `effective_pull_policy` in `FileModuleStore::get`.

crates/kubelet/src/module_store.rs
itowlson: This is an excellent question and one of the things I was unsure about in the API design.  What I wanted to do was encode the mandated logic from https://kubernetes.io/docs/concepts/configuration/overview/#container-images so that other module stores didn't have to replicate that.  But that logic depends deeply on all the stuff about local presence and registry digests, so that means _implementers_ of ModuleStore need to surface that stuff for `get` to use, even though _consumers_ of ModuleStore shouldn't care.

So maybe the pattern is wrong here.  Maybe there should be a `ModuleStore` facade struct which exposes _only_ the `get` method (which requires a pull policy) - as a struct impl but not a trait - but the facade struct has `storage` and `fetcher` members which take care of the local cache and registry interaction side of things respectively.  `get` could use those members but they would be private and so not surfaced to users of the `ModuleStore`.  This shifts us away from the idea that `ModuleStore` is just a magic black box that you ask it for a module and lo! it dispenses one - it mandates some specifics about _how_ it works internally.

Though I guess we could still have a `ModuleStore` trait for the magic `get` dispenser interface, and a `TheOnlyModuleStoreYoullEverNeed` struct for the standard `get` implementation over arbitrary storers and fetchers.  That's probably nicest.

Bah, between this and that rogue Fisher moving all the files, I'm almost thinking I should start over.  The work and tests would come straight across but I'm not sure trying to change in place on top of a rebase/merge from hell is going to be worth it...!

crates/kubelet/src/module_store.rs
itowlson: I thought we were means to use the digest from the registry.

And yeah, if someone fiddles with the internals of the store, then they could mess things up.  But if we're worried about that then shouldn't we also worry about someone taking a hex editor to the `module.wasm` file?

crates/kubelet/src/module_store.rs
thomastaylor312: That idea sounds interesting for sure. Are you saying the `storage` and `fetcher` things would be Traits?

crates/kubelet/src/module_store.rs
thomastaylor312: Both are viable attacks, but the digest feels pretty simple to change. As for the digest, we do get it from the registry, but can't we calculate it locally as well? I honestly don't know enough here

itowlson: From my memory of the CNAB distribution wars, different registries can calculate digests differently (and differently from a client).  I am pretty sure one of the difficulties with that spec was that we could not get a digest for a bundle without first pushing it to a registry.

crates/kubelet/src/module_store.rs
itowlson: Yes, like the `Client` trait we currently pass into `FileStore`.

itowlson:
I think I've addressed everything except the possible refactor to avoid surfacing the implementation methods of `Store`.  Will noodle on that, or we can tackle it in a follow-up PR so this one doesn't sit around!

itowlson:
Okay, I did the refactor of Store, and while I'm not 100% satisfied with it, I agree it's better than it was.  I did have a deadlock bug - now fixed, but would be interested to know if there is anything else in all the maze of Mutexes and RwLocks that I should be watching out for.  (I've also noticed slowness in the pull tests, but this also occurs on main, so I think this is due to slowness at ACR or on the tubes, rather than any blunderous lock contention that I might have introduced.)

crates/kubelet/src/module_store.rs
thomastaylor312: Ah ok that makes sense

crates/kubelet/src/store/mod.rs
```diff
@@ -72,3 +124,82 @@ pub trait Store {
...
+/// but caches it in local storage.
+pub struct CachingStore<S: Storer, C: Client> {
+    storer: Arc<RwLock<S>>,
+    client: Arc<Mutex<C>>,
+}
+
+impl<S: Storer, C: Client> CachingStore<S, C> {
+    async fn pull(&self, image_ref: &Reference) -> anyhow::Result<()> {
+        debug!("Pulling image ref '{:?}' from registry", image_ref);
+        let image_data = self.client.lock().await.pull(image_ref).await?;
```
thomastaylor312, line 138: Does the client need a lock? Just thinking through it and if all it is doing is pulling an image, I don't think there is any need to lock it, but I could be way off here

crates/kubelet/src/store/oci/file.rs
```diff
@@ -1,76 +1,386 @@
...
+            ));
         }
 
         debug!("Fetching image ref '{:?}' from disk", image_ref);
         Ok(tokio::fs::read(path).await?)
     }
+    async fn store(&mut self, image_ref: &Reference, image_data: ImageData) -> anyhow::Result<()> {
+        tokio::fs::create_dir_all(self.pull_path(image_ref)).await?;
+        let digest_path = self.digest_file_path(image_ref);
+        if digest_path.exists() {
```
thomastaylor312, line 79: I think this can be removed. `tokio::fs::write` completely overwrites the content of a file if it exists

crates/kubelet/src/store/mod.rs
```diff
@@ -72,3 +124,82 @@ pub trait Store {
             .collect()
     }
 }
+
+/// A `Store` implementation which obtains module data from remote registries
+/// but caches it in local storage.
+pub struct CachingStore<S: Storer, C: Client> {
```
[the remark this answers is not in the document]
itowlson: I thought about `Local` and rejected it because the 'local' store might not be actually local, e.g. a blob store.  But I guess the intent is about locality, that the private storage is nearby - and the doc comment already says local.  Yeah, all right.  Sold.

crates/kubelet/src/store/mod.rs
```diff
@@ -62,7 +108,13 @@ pub trait Store {
                 .clone()
                 .expect("FATAL ERROR: container must have an image");
             let reference = Reference::try_from(image).unwrap();
-            async move { Ok((container.name.clone(), self.get(&reference).await?)) }
+            let pull_policy = ModulePullPolicy::parse(container.image_pull_policy.clone()).unwrap();
```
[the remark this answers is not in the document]
itowlson: Good catch.  They should.  Thanks.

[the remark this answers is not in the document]
itowlson: Oh except this is in the closure where we are mapping over the containers collection so I'm not sure we can.

crates/kubelet/src/store/mod.rs
itowlson: Not sure.  The lock was there before so I cargo-culted it.

crates/kubelet/src/store/oci/file.rs
itowlson: No, this is important (and so should maybe be commented).  If the write of the `module.wasm` fails (possibly part way through) - or if the write of the new digest file fails - I do _not_ want to leave the old digest file lying around saying "hey I have this module and its digest is blah."  We have to delete the digest file first, so that the failure mode is "I don't know the digest of this module file, so better to re-fetch than to assume it's any particular digest."

crates/kubelet/src/store/mod.rs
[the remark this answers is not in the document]
thomastaylor312: Ok, let's just create an issue to improve error handling here. Trying to get rid of all the places that panic

crates/kubelet/src/store/mod.rs
thomastaylor312: Very well, let's keep it

crates/kubelet/src/store/oci/file.rs
thomastaylor312: Ok, I wondered if that was the case. Let's comment this just to be safe

crates/kubelet/src/store/mod.rs
```diff
@@ -1,14 +1,53 @@
...
 use async_trait::async_trait;
 use log::debug;
 use oci_distribution::Reference;
 
 use crate::pod::Pod;
+use crate::store::oci::Client;
+
+/// Specifies how the store should check for module updates
+#[derive(PartialEq, Debug)]
+pub enum ModulePullPolicy {
```
bacongobbler, line 20: Since this can also be used for non-module projects without any changes, should we call this a `PullPolicy` instead?

I imagine you could use this with OCI images as well, for example.

PR #301, opened 2020-06-23
Fix typo and Bootstrapping clarification

kflansburg:
I'm still concerned that it isn't very clear. The bootstrapping procedure appears to be duplicated in several places (each Kubernetes flavor). Might be better to point them all at the one Bootstrapping README.

### DIFF
--- a/docs/howto/bootstrapping.md
+++ b/docs/howto/bootstrapping.md
@@ -131,7 +131,7 @@ Once you have the bootstrap config in place, you can run Krustlet:
 $ KUBECONFIG=~/.krustlet/config/kubeconfig krustlet-wasi --port 3000 --bootstrap-file /path/to/your/bootstrap.conf
 ```
 
-Krustlet will begin the bootstraping process, and then **await manual certificate approval** (described below) before launching.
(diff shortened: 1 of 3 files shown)


docs/howto/bootstrapping.md
```diff
@@ -131,6 +131,8 @@ Once you have the bootstrap config in place, you can run Krustlet:
 $ KUBECONFIG=~/.krustlet/config/kubeconfig krustlet-wasi --port 3000 --bootstrap-file /path/to/your/bootstrap.conf
 ```
 
+Krustlet will begin the bootstraping process, and then **await manual certificate approval** (described below) before launching.
```
bacongobbler, line 134: typo: bootstrapping*

kflansburg: Ugh. Embarrassing

thomastaylor312:
@kflansburg I totally agree it is duplicated. The problem is the separation between approving the cert and the other steps you have to do before (like setting up the systemd unit, downloading krustlet, etc.). So you end up jumping back and forth between docs. I'll try to take another shot at it during this milestone to see if I can simplify it

PR #303, opened 2020-06-29
The wasmerciser

### LINKED ISSUE #205: Create configurable testing module for e2e tests

itowlson:
Fixes #205.  At least minimally!

This produces a WASM module that can be run with a series of 'commands' (provided as command line arguments) that instruct it to:

* Assert that a file or environment variable exists
* Read from a file or environment variable
* Verify the contents of anything it read
* Write to a file, or stdout or stderr

This can be used in integration tests to verify that volume mounts, secrets, config maps, etc. are working correctly.

For example, to verify that a secret containing `mysecret` has been mounted at `/the_secret`, you would run the wasmerciser with the arguments:

`assert_exists(file:/the_secret) read(file:/the_secret)to(var:secret_text) assert_value(var:secret_text)is(lit:mysecret)`

The syntax is a bit foul but the idea was to avoid having to deal with embedded spaces.  I'm very open to bikeshedding it!

I am not sure how to add the wasmerciser tests to the CI tests but it would probably be a good idea.  In the meantime they can be run by running `just test` from the `demos/wasi/wasmerciser` directory.  There is also a `just build-wasm` target which produces a `.wasm` file that can be run with wasmtime.


itowlson:
Adding @rylev to advise on if there is a less worse way to handle the shared vector reference in the unit tests (or any other better way to unit-test the 'write to...' feature).

PR #304, opened 2020-06-29
Add automatic port mapping for waSCC demos

### LINKED ISSUE #288: Thread panics when running multiple waSCC demos on the same port
I ran into this problem when I tried running 2 waSCC demos at once. Since two examples are running on the same port (8080), the thread panicked and failed. After identifying the issue, I tried recreating the problem and one of the two scenarios would occur.

1. Use `kubectl apply` to create 2 pods from 2 examples. Only the 1st pod is running and the 2nd pod failed. Deleting the 2nd pod and running a `curl localhost:8080` on the 1st pod produces **Failed to handle request**. 
2. Run one demo following the instructions, then start running a second demo. `curl localhost:8080`won't produce any error, it will just show the output of the first demo. Deleting the pod created in the first demo, run `curl localhost:8080` again, it says **curl: (7) Failed to connect to localhost port 8080: Connection refused.**  In the `just run-wascc` terminal, this log can be found: **wascc actor executing <unnamed>' panicked at 'called Result::unwrap() on an Err value: Os { code: 48, kind: AddrInUse, message: "Address already in use" }'.**

Currently, the only way to avoid this problem is to run one demo and delete the pod created,  before running another demo. 

Looking for suggestions to be able to run multiple examples concurrently on different ports automatically. @thomastaylor312 had mentioned in docker, an arbitrary port can be mapped to a port in the container. Kubernetes uses automatic mapping behind the scenes. It might be possible to take advantage of k8s and use it here.

wangyira:
Fixes #288 
When multiple waSCC demos are running on the same port (containerPort or hostPort), to avoid thread panics, automatically maps the ports following these behaviors: 
1. If containerPort is specified and hostPort is not, then automatically assign a port number
2. If hostPort is specified (whether or not containerPort is specified), then try to assign that port number. If it is taken, the pod will error (the pod fails)

### DIFF
--- a/crates/wascc-provider/src/lib.rs
+++ b/crates/wascc-provider/src/lib.rs
@@ -52,7 +52,9 @@ use wascc_host::{Actor, NativeCapability, WasccHost};
 use wascc_httpsrv::HttpServerProvider;
 use wascc_logging::{LoggingProvider, LOG_PATH_KEY};
 
-use std::collections::HashMap;
+extern crate rand;
+use rand::Rng;
+use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
@@ -128,6 +130,7 @@ pub struct WasccProvider<S> {
     log_path: PathBuf,
     kubeconfig: kube::Config,
     host: Arc<Mutex<WasccHost>>,
+    port_set: Arc<Mutex<HashSet<i32>>>,
 }
 
 impl<S: Store + Send + Sync> WasccProvider<S> {
@@ -141,6 +144,7 @@ impl<S: Store + Send + Sync> WasccProvider<S> {
         let host = Arc::new(Mutex::new(WasccHost::new()));
         let log_path = config.data_dir.join(LOG_DIR_NAME);
         let volume_path = config.data_dir.join(VOLUME_DIR);
+        let port_set = Arc::new(Mutex::new(HashSet::<i32>::new()));
         tokio::fs::create_dir_all(&log_path).await?;
         tokio::fs::create_dir_all(&volume_path).await?;
 
@@ -188,6 +192,7 @@ impl<S: Store + Send + Sync> WasccProvider<S> {
             log_path,
             kubeconfig,
             host,
+            port_set,
         })
     }
 }
@@ -217,6 +222,53 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
         let client = kube::Client::new(self.kubeconfig.clone());
         let volumes = Ref::volumes_from_pod(&self.volume_path, &pod, &client).await?;
         for container in pod.containers() {
+            //switch code to here, before env var gets set
+            let mut port_assigned: i32 = 0;
+            if let Some(container_vec) = container.ports.as_ref() {
+                for c_port in container_vec.iter() {
+                    let container_port = c_port.container_port;
+                    let host_port = c_port.host_port;
+                    //self.port_set.lock().unwrap().insert(30000);
+                    if c_port.host_port.is_none() {
+                        println!("host port is not specified");
+                        if container_port >= 0 && container_port <= 65536 {
+                            //find a port that's not taken and assign it
+                            port_assigned = find_available_port(&self.port_set);
+                            if port_assigned == -1 {
+                                error!("Failed to assign port {}, all ports between 30000 and 32767 are taken", &host_port.unwrap());
+                                return Err(anyhow::anyhow!(
+                                    "All ports between 30000 and 32767 are taken"
+                                ));
+                            }
+                        }
+                    } else {
+                        println!("host port is specified");
+
+                        if !self.port_set.lock().unwrap().contains(&host_port.unwrap()) {
+                            //not taken, assign that port
+                            port_assigned = host_port.unwrap();
+                            println!("host port is available, so assign it");
+                            self.port_set.lock().unwrap().insert(port_assigned);
+                        } else {
+                            //port is taken, error
+                            error!(
+                                "Failed to assign hostport {}, because it's taken",
+                                &host_port.unwrap()
+                            );
+                            return Err(anyhow::anyhow!(
+                                "Port {} is currently in use",
+                                port_assigned
+                            ));
+                        }
+                    }
+                }
+            }
+            println!(
+                "ports that are already taken: {:?}",
+                self.port_set.lock().unwrap()
+            );
+            println!("new port assigned is: {}", port_assigned);
+            //container.env
             let env = Self::env_vars(&container, &pod, &client).await;
             let volume_bindings: Vec<VolumeBinding> =
                 if let Some(volume_mounts) = container.volume_mounts().as_ref() {
@@ -255,7 +307,16 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
             });
             let host = self.host.clone();
             let http_result = tokio::task::spawn_blocking(move || {
-                wascc_run_http(host, module_data, env, volume_bindings, &lp, status_recv)
+                println!("wascc run http");
+                wascc_run_http(
+                    host,
+                    module_data,
+                    env,
+                    volume_bindings,
+                    &lp,
+                    status_recv,
+                    port_assigned,
+                )
             })
             .await?;
             match http_result {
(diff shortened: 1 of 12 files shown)


crates/wascc-provider/src/lib.rs
```diff
@@ -437,21 +498,40 @@ struct VolumeBinding {
...
     caps.push(Capability {
         name: HTTP_CAPABILITY,
         binding: None,
-        env,
+        env: env_map, //replaced _env
     });
     wascc_run(host, data, &mut caps, volumes, log_path, status_recv)
 }
 
+fn find_available_port(port_set: &Arc<Mutex<HashSet<i32>>>) -> i32 {
```
bacongobbler, line 519: I notice here that you are returning an `i32` with a negative integer in the case of an error. Can you please have a look at [Result types](https://doc.rust-lang.org/std/result/) in Rust to return errors?

crates/wascc-provider/src/lib.rs
```diff
@@ -214,6 +219,53 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
         let client = kube::Client::new(self.kubeconfig.clone());
         let volumes = Ref::volumes_from_pod(&self.volume_path, &pod, &client).await?;
         for container in pod.containers() {
+            //switch code to here, before env var gets set
+            let mut port_assigned: i32 = 0;
+            if let Some(container_vec) = container.ports.as_ref() {
+                for c_port in container_vec.iter() {
+                    let container_port = c_port.container_port;
+                    let host_port = c_port.host_port;
+                    //self.port_set.lock().unwrap().insert(30000);
```
bacongobbler, line 228: if this is irrelevant, can you remove this?

crates/wascc-provider/src/lib.rs
```diff
@@ -214,6 +219,53 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
...
+                    if c_port.host_port.is_none() {
+                        println!("host port is not specified");
```
bacongobbler, line 230: can you review these `println` statements and see if they would be worth turning into `debug!` calls?

crates/wascc-provider/src/lib.rs
bacongobbler: What do you think about writing some unite tests to test some of the common cases this function covers?

cc @thomastaylor312, do you think we need to write some unit tests before merging this code?

crates/wascc-provider/src/lib.rs
```diff
@@ -437,21 +498,40 @@ struct VolumeBinding {
...
 
+fn find_available_port(port_set: &Arc<Mutex<HashSet<i32>>>) -> i32 {
+    let mut range = rand::thread_rng();
+    let mut port: i32 = -1;
+
+    let mut empty_port: HashSet<i32> = HashSet::new();
+    while empty_port.len() < 2768 {
+        port = range.gen_range(30000, 32768);
+        empty_port.insert(port);
+        if !port_set.lock().unwrap().contains(&port) {
```
bacongobbler, line 527: do we want to handle cases such as when another process running outside of Krustlet is bound to this port, we fail here, or do we consider this an acceptable use case and error out as we do in #288?

crates/wascc-provider/src/lib.rs
```diff
@@ -214,6 +221,38 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
         let client = kube::Client::new(self.kubeconfig.clone());
         let volumes = Ref::volumes_from_pod(&self.volume_path, &pod, &client).await?;
         for container in pod.containers() {
+            let mut port_assigned: i32 = 0;
+            info!(
+                "The following ports are already taken: {:?}",
+                self.port_set.lock().unwrap().keys()
```
thomastaylor312, line 227: Because this is a locking operation (that involves some overhead and locks other threads from using it). This should probably be a `trace!`.

crates/wascc-provider/src/lib.rs
```diff
@@ -393,6 +440,13 @@ impl<S: Store + Send + Sync> Provider for WasccProvider<S> {
     }
 
     async fn delete(&self, pod: Pod) -> anyhow::Result<()> {
+        let mut delete_key: i32 = 0;
+        for (key, val) in self.port_set.lock().unwrap().iter() {
```
thomastaylor312, line 444: This could become a pretty expensive operation. You'll have to iterate over every taken port (which at minimum will be hundreds, but could be more). I would probably stick with the hash set and just iterate over all the ports from the pod and remove those from the hash set

thomastaylor312: Nit: one thing you could do here too is just take the lock once

crates/wascc-provider/src/lib.rs
```diff
@@ -437,21 +491,66 @@ struct VolumeBinding {
...
 ) -> anyhow::Result<ContainerHandle<ActorHandle, LogHandleFactory>> {
     let mut caps: Vec<Capability> = Vec::new();
 
+    let mut env_map: HashMap<String, String> = HashMap::new();
+    env_map.insert("PORT".to_string(), port_assigned.to_string());
     caps.push(Capability {
         name: HTTP_CAPABILITY,
         binding: None,
-        env,
+        env: env_map, //replaced _env
```
thomastaylor312, line 507: I think you need to add the port to the EnvVars that are passed in instead of the new hashmap. Otherwise, I don't think the actor will get the environment variables from the pod definition

crates/wascc-provider/src/lib.rs
```diff
@@ -437,21 +491,66 @@ struct VolumeBinding {
...
+    fn new() -> PortAllocationError {
+        PortAllocationError {}
+    }
+}
+impl fmt::Display for PortAllocationError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "all ports are currently in use")
+    }
+}
+impl Error for PortAllocationError {
```
thomastaylor312, line 525: Nit: We could probably just use the `anyhow` crate derives instead of manually implementing `Error`, but not going to block the PR on it

crates/wascc-provider/src/lib.rs
```diff
@@ -437,21 +491,66 @@ struct VolumeBinding {
...
+        "all ports are currently in use"
+    }
+}
+
+fn find_available_port(
+    port_set: &Arc<Mutex<HashMap<i32, String>>>,
+    pod_name: String,
+) -> Result<i32, PortAllocationError> {
+    let mut range = rand::thread_rng();
+    let mut port: i32 = -1;
```
thomastaylor312, line 536: Nit: It would probably be more rusty to have this be an `Option<i32>` and then just do `port.replace` when we find a port

demos/wascc/greet/greet-wascc.yaml
```diff
@@ -10,8 +10,8 @@ spec:
       imagePullPolicy: Always
       name: greet
       env:
-      - name: PORT
-        value: "8080"
+        - name: PORT
```
thomastaylor312, line 13: This can be removed as you are now setting `PORT` in the code

demos/wascc/uppercase/uppercase-wascc.yaml
```diff
@@ -10,8 +10,8 @@ spec:
       imagePullPolicy: Always
       name: uppercase
       env:
-      - name: PORT
-        value: "8080"
+        - name: PORT
```
thomastaylor312, line 13: Same thing with removing this

crates/wascc-provider/src/lib.rs
bacongobbler: should we assign the lock to another variable? I see that we're acquiring a lock on this same variable on lines 234 and 236, so perhaps it may be easier to block `port_set` for the entire function call?

crates/wascc-provider/src/lib.rs
bacongobbler: agreed. Acquiring the lock once and re-using the lock on lines 444 and 449 would prevent excessive locking/unlocking operations.

crates/wascc-provider/src/lib.rs
bacongobbler: Good call. Then we could change the logic on line 548 to `if port.is_none()` and line 551 to `Ok(port.unwrap())`.

https://doc.rust-lang.org/std/option/enum.Option.html

thomastaylor312: To be even safer, we can just do `port.ok_or(PortAllocationError::new())` at the end to transform it into a `Result`

demos/wascc/fileserver/k8s.yaml
```diff
@@ -11,7 +11,6 @@ spec:
       name: fileserver-wascc
       env:
```
thomastaylor312, line 12: Bad communication on my part. This whole `env` section should be removed as we no longer need anything in it

demos/wascc/greet/greet-wascc.yaml
```diff
@@ -10,8 +10,7 @@ spec:
       imagePullPolicy: Always
       name: greet
       env:
```
thomastaylor312, line 12: Same with removing here

demos/wascc/hello-world-assemblyscript/k8s.yaml
```diff
@@ -8,7 +8,6 @@ spec:
       image: webassembly.azurecr.io/hello-world-wascc-assemblyscript:v0.1.0
       env:
```
thomastaylor312, line 9: Same here

demos/wascc/uppercase/uppercase-wascc.yaml
```diff
@@ -10,8 +10,7 @@ spec:
       imagePullPolicy: Always
       name: uppercase
       env:
```
thomastaylor312, line 12: And here

PR #307, opened 2020-06-29
fix(bootstrap): Makes sure to not write empty file

### LINKED ISSUE #300: Node TLS bootstrap: empty server certificate created when informer times out
## Repro steps

Ensure there are no CSR resources related to the `kubelet-wasi` node and then run:

```text
$ rm -rf ~/.krustlet
$ just run-wasi
```

Do not approve the node certificate and wait for the node to start (may take 5+ minutes).

## Expected results

The node should either continue to wait for up to some configurable time limit (I think kubelet defaults to an hour?) or it should treat an informer timeout as a failure to authorize the node's certificate.

## Actual results

The node starts up anyway and creates an empty certificate at `~/.krustlet/config/krustlet-wasi.crt`.  This will later result in certificate errors when trying to obtain any logs from the node.

The [problem lies in `bootstrap_tls`](https://github.com/deislabs/krustlet/blob/cf04acb2ec8d32668f10cc51712ac2e346266f19/crates/kubelet/src/bootstrapping/mod.rs#L183) where a graceful termination of the informer's poll stream will result in using an empty certificate string.

Note the same bug also exists in `bootstrap_auth`'s implementation, although auto-approval of the auth CSR makes it unlikely to encounter.

## Environment

OS: macOS 10.15.5
k8s: Kubernetes for Docker (1.16.5).

## **Additional related bug**

It appears both the WASCC and WASI providers initialize the logging *after* TLS bootstrapping (e.g. [here](https://github.com/deislabs/krustlet/blob/master/src/krustlet-wasi.rs#L15)), so I had to move the initialization of logging to get the bootstrap's diagnostic messages.

thomastaylor312:
The bootstrap now checks to make sure the certs were approved before
writing anything to a file. I also fixed log init so that it happens
before bootstrap

Fixes #300

### DIFF
--- a/crates/kubelet/src/bootstrapping/mod.rs
+++ b/crates/kubelet/src/bootstrapping/mod.rs
@@ -98,6 +98,8 @@ async fn bootstrap_auth<K: AsRef<Path>>(
 
         let mut watcher = inf.poll().await?.boxed();
         let mut generated_kubeconfig = Vec::new();
+        let mut got_cert = false;
+        let start = std::time::Instant::now();
         while let Some(event) = watcher.try_next().await? {
             match event {
                 WatchEvent::Modified(m) | WatchEvent::Added(m) => {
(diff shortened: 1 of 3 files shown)


thomastaylor312:
@peterhuene I tried to add config to the timeout value, but the k8s API only allows timeouts up to 295s on a watch, so it would be a bigger refactor if we want to support any value. Otherwise, I think this addresses everything.

PR #310, opened 2020-06-30
fix demos to use justfiles

### LINKED ISSUE #281: convert demo Makefiles to justfiles
Some of our demonstrations use a Makefile. We should switch to justfiles to keep the tooling consistent between the demos and the main project

bacongobbler:
closes #281


demos/wascc/uppercase/justfile
```diff
@@ -0,0 +1,34 @@
...
+	@{{CARGO}} build --release
+	wascap sign {{RELEASE}}/{{NAME}}.wasm {{RELEASE}}/{{NAME}}_signed.wasm -i {{KEYDIR}}/account.nk -u {{KEYDIR}}/module.nk -s -f -l -n {{NAME}}
+
+push:
+	wasm-to-oci push {{RELEASE}}/{{NAME}}_signed.wasm webassembly.azurecr.io/{{NAME}}-wascc:v{{VERSION}}
+
+
+keys: account module
+
+account:
```
thomastaylor312, line 26: Future improvement suggestion: We should move this keys target to a top level justfile we can call to avoid copy/pasting

PR #312, opened 2020-07-01
Support init containers

### LINKED ISSUE #25: Implement init containers for all runtimes

itowlson:
Fixes #25 

**NOTES:**

* This first drop _only supports init containers in the WASI provider._  #25 requires them for all runtimes.
* ~waSCC will be harder to test because it doesn't allow us to pass command line args.  We can address this by improving the wasmerciser to accept commands via an environment variable as well as the command line.~ _EV support is now implemented_
* ~This will not currently pass CI because the integrations tests use the wasmerciser (to write a file in the init container and then read it back in the main container).  I've checked the tests pass on my machine by using a hacked store that can load from local `.wasm` files, but they won't work yet with the real OCI store.~

### DIFF
--- a/crates/kubelet/src/pod/mod.rs
+++ b/crates/kubelet/src/pod/mod.rs
@@ -215,6 +215,17 @@ impl Pod {
             .collect()
     }
 
+    /// Get a pod's init containers
+    pub fn init_containers(&self) -> Vec<Container> {
+        // We can't use and_then because that requires the lambda to return an Option,
+        // but we have to return an &Option to avoid a move.
+        let inits = match self.0.spec.as_ref() {
+            None => &None,
+            Some(spec) => &spec.init_containers
+        };
+        inits.as_ref().unwrap_or_else(|| &EMPTY_VEC).iter().map(|c| Container::new(c)).collect()
+    }
+
     /// Turn the Pod into the Kubernetes API version of a Pod
     pub fn into_kube_pod(self) -> KubePod {
         self.0
--- a/crates/kubelet/src/pod/mod.rs
+++ b/crates/kubelet/src/pod/mod.rs
@@ -223,6 +223,33 @@ impl Pod {
         }
     }
 
+    /// Sets the status of all specified containers to Waiting, and optionally
+    /// sets the pod status message.
+    pub async fn initialise_status(
+        &self,
+        client: &kube::Client,
+        container_names: &Vec<&String>,
+        initial_message: Option<String>,
+    ) {
+        let mut all_waiting_map = HashMap::new();
+        for name in container_names {
+            let waiting = crate::container::Status::Waiting {
+                timestamp: chrono::Utc::now(),
+                message: "PodInitializing".to_owned(),
+            };
+            all_waiting_map.insert(name.to_string(), waiting);
+        }
+        let initial_status_message = match initial_message {
+            Some(m) => StatusMessage::Message(m),
+            None => StatusMessage::Clear,
+        };
+        let all_waiting = Status {
+            message: initial_status_message,
+            container_statuses: all_waiting_map,
+        };
+        self.patch_status(client.clone(), all_waiting).await;
+    }
+
     /// Get a pod's containers
     pub fn containers(&self) -> Vec<Container> {
         self.0
--- a/crates/kubelet/src/pod/mod.rs
+++ b/crates/kubelet/src/pod/mod.rs
@@ -217,13 +217,14 @@ impl Pod {
 
     /// Get a pod's init containers
     pub fn init_containers(&self) -> Vec<Container> {
-        // We can't use and_then because that requires the lambda to return an Option,
-        // but we have to return an &Option to avoid a move.
-        let inits = match self.0.spec.as_ref() {
-            None => &None,
-            Some(spec) => &spec.init_containers,
-        };
-        inits.as_ref().unwrap_or_else(|| &EMPTY_VEC).iter().map(|c| Container::new(c)).collect()
+        self.0
+            .spec
+            .as_ref()
+            .and_then(|s| s.init_containers.as_ref())
+            .unwrap_or(&EMPTY_VEC)
+            .iter()
+            .map(|c| Container::new(c))
+            .collect()
     }
 
     /// Turn the Pod into the Kubernetes API version of a Pod
--- a/crates/kubelet/src/pod/mod.rs
+++ b/crates/kubelet/src/pod/mod.rs
@@ -221,7 +221,7 @@ impl Pod {
         // but we have to return an &Option to avoid a move.
         let inits = match self.0.spec.as_ref() {
             None => &None,
-            Some(spec) => &spec.init_containers
+            Some(spec) => &spec.init_containers,
         };
         inits.as_ref().unwrap_or_else(|| &EMPTY_VEC).iter().map(|c| Container::new(c)).collect()
     }
--- a/crates/kubelet/src/pod/mod.rs
+++ b/crates/kubelet/src/pod/mod.rs
@@ -5,7 +5,7 @@ mod status;
 
 pub use handle::{key_from_pod, pod_key, Handle};
 pub(crate) use queue::Queue;
-pub use status::{update_status, Phase, Status};
+pub use status::{update_status, Phase, Status, StatusMessage};
 
 use std::collections::HashMap;
 
@@ -172,7 +172,9 @@ impl Pod {
                 }
             }
         }
+
         // is there ever a case when we get a status that we should end up in Phase unknown?
+
         let phase = if num_succeeded == container_statuses.len() {
             Phase::Succeeded
         } else if failed {
@@ -181,18 +183,36 @@ impl Pod {
             Phase::Running
         };
 
-        let json_status = serde_json::json!(
+        // TODO: init container statuses need to be reported through initContainerStatuses
+
+        let mut json_status = serde_json::json!(
             {
                 "metadata": {
                     "resourceVersion": "",
                 },
                 "status": {
                     "phase": phase,
-                    "message": status.message,
                     "containerStatuses": container_statuses
                 }
             }
         );
+
+        if let Some(map) = json_status.as_object_mut() {
+            if let Some(status_json) = map.get_mut("status") {
+                if let Some(status_map) = status_json.as_object_mut() {
+                    let message_key = "message".to_owned();
+                    match status.message {
+                        StatusMessage::LeaveUnchanged => (),
+                        StatusMessage::Clear => {
+                            status_map.insert(message_key, serde_json::Value::Null);
+                        }
+                        StatusMessage::Message(m) => {
+                            status_map.insert(message_key, serde_json::Value::String(m));
+                        }
+                    }
+                }
+            }
+        }
 
         debug!("Setting pod status for {} using {:?}", name, json_status);
 
--- a/crates/wasi-provider/src/lib.rs
+++ b/crates/wasi-provider/src/lib.rs
@@ -48,7 +48,7 @@ use kubelet::provider::Provider;
 use kubelet::provider::ProviderError;
 use kubelet::store::Store;
 use kubelet::volume::Ref;
-use log::{debug, error, info, trace};
+use log::{debug, error, info, trace, warn};
 use tokio::sync::RwLock;
 use wasi_runtime::{Runtime, WasiRuntime};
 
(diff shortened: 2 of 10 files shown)


crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+    }
+
+    async fn run_one_container(
+        &self,
+        container: &KubeContainer,
+        pod: &Pod,
+        client: &kube::client::Client,
+        modules: &mut HashMap<String, Vec<u8>>,
+        volumes: &HashMap<String, Ref>,
+        container_handles: &mut HashMap<
```
thomastaylor312, line 131: I think it would probably be better to just return a container handle from this function rather that passing a mutable handle. From a code organization standpoint I think it is clearer to manage the full map of handles in the main handler function (`add`)

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+    }
+
+    async fn run_one_container_to_completion(
+        &self,
+        container: &KubeContainer,
+        pod: &Pod,
+        client: &kube::client::Client,
+        modules: &mut HashMap<String, Vec<u8>>,
+        volumes: &HashMap<String, Ref>,
+        container_handles: &mut HashMap<
```
thomastaylor312, line 166: Same thing here with respect to a mutable reference

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+                container.name,
+                result.message
+            ))
+        }
+    }
+
+    async fn wait_for_termination(
+        &self,
+        container: &KubeContainer,
+        container_handles: &mut HashMap<
```
thomastaylor312, line 191: And here with the mutable reference as well.

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+            String,
+            kubelet::container::Handle<wasi_runtime::Runtime, wasi_runtime::HandleFactory>,
+        >,
+    ) -> anyhow::Result<ContainerTerminationResult> {
+        let handle = container_handles.get(&container.name).unwrap();
+        let mut status = handle.status();
+        loop {
+            // TODO: clippy wants us to do it this way but ugh
```
thomastaylor312, line 199: What is the other way you are trying to do it here?

crates/wasi-provider/src/lib.rs
```diff
@@ -107,51 +235,30 @@ impl<S: Store + Send + Sync> Provider for WasiProvider<S> {
...
+                container,
+                &pod,
+                &client,
+                &mut modules,
+                &volumes,
+                &mut container_handles,
+            )
+            .await?;
+        }
+        info!("Finished running init containers for pod {:?}", pod_name);
```
thomastaylor312, line 250: We should probably put a debug log in the loop saying which init container we are running right before it starts and right after it finishes

crates/kubelet/src/pod/mod.rs
```diff
@@ -211,6 +211,17 @@ impl Pod {
...
 
+    /// Get a pod's init containers
+    pub fn init_containers(&self) -> &Vec<KubeContainer> {
+        // We can't use and_then because that requires the lambda to return an Option,
+        // but we have to return an &Option to avoid a move.
+        let inits = match self.0.spec.as_ref() {
+            None => &None,
+            Some(spec) => &spec.init_containers,
+        };
+        inits.as_ref().unwrap_or_else(|| &EMPTY_VEC)
```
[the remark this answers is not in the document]
itowlson: Hmm, looks like this doesn't work, because `spec.init_containers` is an `Option<Vec<...>>` so doesn't have the `borrow` method.  I think this whole fiddle-faddle was to get around the optionality of `init_containers`...

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+                    // been validated by the k8s API
+                    Ok((vol.deref().clone(), Some(guest_path)))
+                })
+                .collect::<anyhow::Result<HashMap<PathBuf, Option<PathBuf>>>>()
+        } else {
+            Ok(HashMap::default())
+        }
+    }
+
+    async fn run_one_container(
```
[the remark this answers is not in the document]
itowlson: Should probably also be `start_container` for avoidance of ambiguity.

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +90,127 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+        &self,
+        container: &KubeContainer,
+        pod: &Pod,
+        client: &kube::client::Client,
+        modules: &mut HashMap<String, Vec<u8>>,
```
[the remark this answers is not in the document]
itowlson: I think that's a good idea, a sort of "runtime pod" type, though it would be quite provider-specific.

crates/wasi-provider/src/lib.rs
itowlson: That sounds desirable.  Need to figure out how it will play with removing completed init containers from the map (which I found I had to do and am probably doing in the wrong place at the moment).

crates/wasi-provider/src/lib.rs
itowlson: I think it was a `match` but can't remember now.  I just don't like the way the destructuring of the Terminated enum breaks up the flow.  I'll take another look.

_copy of "Refactoring" eerily falls open at the entry for 'Extract Method'_

crates/wasi-provider/src/lib.rs
itowlson: The start message is already there in `start_container` (fka `run_one_container`).  Added an end message to `run_container_to_completion`.

crates/wasi-provider/src/lib.rs
itowlson: Here we go, during refactoring I made it the way I wanted it to look and:

```rust
warning: you seem to be trying to use match for destructuring a single pattern. Consider using `if let`
   --> crates/wasi-provider/src/lib.rs:193:13
    |
193 | /             match status {
194 | |                 Some(kubelet::container::Status::Terminated {
195 | |                     message, failed, ..
196 | |                 }) => {
...   |
202 | |                 _ => (),
203 | |             }
    | |_____________^
```

With the refactoring and the explaining variable, though, the `if let` won't look quite as horrid...

itowlson: Update: it still looks foul #damnyouclippy

crates/kubelet/src/pod/mod.rs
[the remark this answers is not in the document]
thomastaylor312: Yeah, I figured it was. Guess we are just doomed to handle the All Optional™ Kubernetes API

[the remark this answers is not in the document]
itowlson: When I understand Rust better I will probably rage-code something for working with all-optional chains in a less painful way.  `and_then` is meant to support a monadic style but in this case it didn't work either... but if there is a consistent pattern to the k8s APIs then maybe we can figure something out.

"If there is a consistent pattern to the k8s APIs."  Ha ha, what a comedian.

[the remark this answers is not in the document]
peterhuene: Perhaps:

```rust
fn init_containers(&self) -> &[KubeContainer] {
    self.0
        .spec
        .as_ref()
        .and_then(|s| s.init_containers.as_deref())
        .unwrap_or(&[])
}
```

There's not much benefit for returning `&Vec<T>` vs a slice.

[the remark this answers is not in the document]
itowlson: Ah, `as_deref` might be the secret sauce.  Thanks Peter!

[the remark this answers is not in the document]
peterhuene: It'll return an `Option<&[KubeContainer]>` in this case, though, as `&[T]` is the deref target for `Vec<T>`.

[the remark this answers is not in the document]
peterhuene: If we do want to keep the `&Vec<KubeContainer>` return type, this will work:

```rust
self.0
    .spec
    .as_ref()
    .and_then(|s| s.init_containers.as_ref())
    .unwrap_or(&EMPTY_VEC)
```

itowlson:
It looks like init containers don't make much sense for waSCC, so I'm proposing to go ahead without them.  Open to feedback on that plan!

itowlson:
I can't put this back into draft state but please do not merge - I found a bug where we can't get logs from a pod whose init container failed.

tests/pod_builder.rs
```diff
@@ -0,0 +1,113 @@
...
+use serde_json::json;
+use std::sync::Arc;
+
+pub struct PodLifetimeOwner {
+    pub pod: Pod,
+    _tempdirs: Vec<Arc<tempfile::TempDir>>, // only to keep the directories alive
+}
+
+pub struct WasmerciserContainerSpec {
+    pub name: &'static str,
```
thomastaylor312, line 11: Any reason to not just make these `String`s?

tests/integration_tests.rs
```diff
@@ -84,51 +147,34 @@ async fn test_wascc_provider() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    // Send a request to the pod to trigger some logging
-    reqwest::get("http://127.0.0.1:8080")
-        .await
-        .expect("unable to perform request to test pod");
+impl Drop for WasccTestResourceCleaner {
+    fn drop(&mut self) {
+        let t = std::thread::spawn(move || {
+            let mut rt =
+                tokio::runtime::Runtime::new().expect("Failed to reate Tokio runtime for cleanup");
```
thomastaylor312, line 161: Nit: s/reate/create

tests/pod_builder.rs
itowlson: I don't want to clutter the calling code with `to_owned()`.  I want spec creation code focused on spec creation and not on optimising memory usage.  (If this were production code then yes I would totally make them `String`s though.)

tests/integration_tests.rs
itowlson: Ooh thanks - also fixed in the place I copied this from.  (Which was also me.)

itowlson:
I've got a known issue where init container statuses are reported through `containerStatus` instead of `initContainerStatus`; otherwise I think this works (fingers crossed for CI).  How do people feel about merging what we have and fixing that bug in a new PR?  I know it's a bit iffy but I'd like to avoid too many more merges/rebases - this is big enough and ugly enough already.

crates/kubelet/src/pod/mod.rs
```diff
@@ -181,19 +183,37 @@ impl Pod {
...
+        // TODO: init container statuses need to be reported through initContainerStatuses
+
+        let mut json_status = serde_json::json!(
             {
                 "metadata": {
                     "resourceVersion": "",
                 },
                 "status": {
                     "phase": phase,
-                    "message": status.message,
```
rylev, line 191: Is it not possible to insert the correct json value here directly?

crates/kubelet/src/pod/mod.rs
```diff
@@ -203,6 +223,33 @@ impl Pod {
...
+        container_names: &Vec<&String>,
+        initial_message: Option<String>,
+    ) {
+        let mut all_waiting_map = HashMap::new();
+        for name in container_names {
+            let waiting = crate::container::Status::Waiting {
+                timestamp: chrono::Utc::now(),
+                message: "PodInitializing".to_owned(),
+            };
+            all_waiting_map.insert(name.to_string(), waiting);
```
rylev, line 240: nit: I think `clone` is preferred to `to_string` in this case

crates/wasi-provider/src/lib.rs
```diff
@@ -83,6 +91,178 @@ impl<S: Store + Send + Sync> WasiProvider<S> {
...
+        info!("Finished running init containers for pod {:?}", pod.name());
+        (container_handles, None)
+    }
+
+    async fn wait_for_terminated_status(
+        status_receiver: &mut tokio::sync::watch::Receiver<kubelet::container::Status>,
+    ) -> ContainerTerminationResult {
+        loop {
+            let status = status_receiver.next().await;
+            if let Some(Status::Terminated {
```
rylev, line 255: I think `while let` should still work here no?

thomastaylor312:
@itowlson I think you'll be ok to merge once tests pass and then we can fix that issue (please open up one and put it in the milestone). We are still in a non-stable state, so an issue like this isn't the worst

crates/kubelet/src/pod/mod.rs
itowlson: @rylev It's easy if the value is "clear the setting" (`None`) or "set it to X" (`Some(X)`) but I couldn't see how to specify "don't include the element at all" which is what k8s seems to need to allow it to be left unchanged.

Actually... I guess we are retrieving the existing pod status, so for leave unchanged I could read the existing status and write it back explicitly.  There's a race condition with that but maybe it's not an issue.

crates/kubelet/src/pod/mod.rs
itowlson: That was what I thought but it doesn't work because `clone` produces an `&String` and I need a `String`.

crates/wasi-provider/src/lib.rs
itowlson: I couldn't see a way to make it work - it looked like a failed match would drop out of the loop rather than go round again, and the compiler certainly thought it would result in ways out of the function other than via the `return`.  I'd be happy to take this suggestion but I'd need a more specific pointer on the correct phrasing of the `while let` - sorry.

itowlson:
Ugh.  No local repro on the wascc integration test...

PR #313, opened 2020-07-03
Implement an API similar to Pod for KubeContainer

kflansburg:
Basically, mirror Pod API behavior for Container, wrapping KubeContainer and providing nice accessor methods. 

As a start, I moved Image and PullPolicy parsing into the container module.

### DIFF
--- a/crates/kubelet/src/store/oci/file.rs
+++ b/crates/kubelet/src/store/oci/file.rs
@@ -125,7 +125,7 @@ async fn file_content_is(path: PathBuf, text: String) -> bool {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::store::PullPolicy;
+    use crate::container::PullPolicy;
     use crate::store::Store;
     use oci_distribution::client::ImageData;
     use std::collections::HashMap;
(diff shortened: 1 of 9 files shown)


crates/kubelet/src/store/oci/file.rs
```diff
@@ -336,12 +330,13 @@ mod test {
         let fake_ref = Reference::try_from("foo/bar:2.0")?;
         let scratch_dir = create_temp_dir();
         let store = FileStore::new(fake_client.clone(), &scratch_dir.path);
-        let module_bytes_orig = store.get(&fake_ref, None).await?;
+        let policy = PullPolicy::parse_with_ref(None, Some(fake_ref.clone()))?;
```
itowlson, line 333: I wonder if this might be better moved to the `Container` tests?  That way you can have a set of tests that focus on the correct resolution of the pull policy, and when you get to the store tests, you can take it as read that we resolve pull policies correctly and can focus on testing that we implement each policy correctly - no need for the "None" tests.

kflansburg: Yes, I wasn't sure about removing those Store tests entirely as I wasn't sure if there was other functionality being tested there as well.

bacongobbler:
Thanks for this @kflansburg! I'll review this in a short while after I'm finished with today's meetings.

PR #316, opened 2020-07-07
Option to load modules from local filesystem

itowlson:
During the dev-test cycle it can be useful to load modules that have been built locally but not pushed to a registry.  This PR supports this via a new `--allow-local-modules=true` command line flag (also available via environment variable or config file).  If `--allow-local-modules` is `true`, references of the form `fs/<path>` are loaded directly from the file system instead of from a registry.  For example:

* `fs//wasm/mymodule.wasm` - loads from absolute path `/wasm/mymodule.wasm`
* `fs/./out/mymodule.wasm` or `fs/out/mymodule.wasm` - loads from path `out/mymodule.wasm`, relative to the working directory of the kubelet process

As it is the custom kubelet (e.g. `krustlet-wasi`) which constructs the module store, it is up to that code to honour the `--allow-local-modules` flag.  Awareness of the flag is _not_ baked into the FileStore or the  module loader.

### DIFF
--- a/crates/kubelet/src/config.rs
+++ b/crates/kubelet/src/config.rs
@@ -53,6 +53,8 @@ pub struct Config {
     pub max_pods: u16,
     /// The location of the tls bootstrapping file
     pub bootstrap_file: PathBuf,
+    /// Whether to allow modules to be loaded directly from the filesystem
+    pub allow_local_modules: bool,
 }
 /// The configuration for the Kubelet server.
 #[derive(Clone, Debug)]
@@ -106,6 +108,8 @@ struct ConfigBuilder {
     pub server_tls_cert_file: Option<PathBuf>,
     #[serde(default, rename = "tlsPrivateKeyFile")]
     pub server_tls_private_key_file: Option<PathBuf>,
+    #[serde(default, rename = "allowLocalModules")]
+    pub allow_local_modules: Option<bool>,
 }
 
 struct ConfigBuilderFallbacks {
@@ -136,6 +140,7 @@ impl Config {
             data_dir,
             max_pods: DEFAULT_MAX_PODS,
             bootstrap_file: PathBuf::from(BOOTSTRAP_FILE),
+            allow_local_modules: false,
             server_config: ServerConfig {
                 addr: match preferred_ip_family {
                     // Just unwrap these because they are programmer error if they
@@ -259,6 +264,7 @@ impl ConfigBuilder {
             hostname: opts.hostname,
             data_dir: opts.data_dir,
             max_pods: ok_result_of(opts.max_pods),
+            allow_local_modules: opts.allow_local_modules,
             server_addr: ok_result_of(opts.addr),
             server_port: ok_result_of(opts.port),
             server_tls_cert_file: opts.cert_file,
@@ -293,6 +299,7 @@ impl ConfigBuilder {
             server_port: other.server_port.or(self.server_port),
             server_tls_cert_file: other.server_tls_cert_file.or(self.server_tls_cert_file),
             bootstrap_file: other.bootstrap_file.or(self.bootstrap_file),
+            allow_local_modules: other.allow_local_modules.or(self.allow_local_modules),
             server_tls_private_key_file: other
                 .server_tls_private_key_file
                 .or(self.server_tls_private_key_file),
@@ -339,6 +346,7 @@ impl ConfigBuilder {
             data_dir,
             max_pods,
             bootstrap_file,
+            allow_local_modules: self.allow_local_modules.unwrap_or(false),
             server_config: ServerConfig {
                 cert_file: server_tls_cert_file,
                 private_key_file: server_tls_private_key_file,
@@ -466,6 +474,13 @@ pub struct Opts {
         default_value = BOOTSTRAP_FILE
     )]
     bootstrap_file: PathBuf,
+
+    #[structopt(
+        long = "allow-local-modules",
+        env = "KRUSTLET_ALLOW_LOCAL_MODULES",
+        help = "Whether to allow loading modules directly from the filesystem"
+    )]
+    allow_local_modules: Option<bool>,
 }
 
 fn default_hostname() -> anyhow::Result<String> {
@@ -590,7 +605,8 @@ mod test {
             "nodeName": "krusty-node",
             "tlsCertificateFile": "/my/secure/cert.pfx",
             "tlsPrivateKeyFile": "/the/key",
-            "bootstrapFile": "/the/bootstrap/file.txt"
+            "bootstrapFile": "/the/bootstrap/file.txt",
+            "allowLocalModules": true
         }"#,
         );
         let config = config_builder.unwrap().build(fallbacks()).unwrap();
@@ -613,6 +629,7 @@ mod test {
         assert_eq!(config.data_dir.to_string_lossy(), "/krusty/data/dir");
         assert_eq!(format!("{}", config.node_ip), "173.183.193.2");
         assert_eq!(config.max_pods, 400);
+        assert_eq!(config.allow_local_modules, true);
         assert_eq!(config.node_labels.len(), 2);
         assert_eq!(config.node_labels.get("label1"), Some(&("val1".to_owned())));
     }
@@ -669,6 +686,7 @@ mod test {
         assert_eq!(config.hostname, "fallback-hostname");
         assert_eq!(config.data_dir.to_string_lossy(), "/fallback/data/dir");
         assert_eq!(format!("{}", config.node_ip), "4.4.4.4");
+        assert_eq!(config.allow_local_modules, false);
         assert_eq!(config.node_labels.len(), 0);
     }
 
@@ -699,6 +717,7 @@ mod test {
                 "label2": "val2"
             },
             "nodeName": "krusty-node",
+            "allowLocalModules": true,
             "tlsCertificateFile": "/my/secure/cert.pfx",
             "tlsPrivateKeyFile": "/the/key"
         }"#,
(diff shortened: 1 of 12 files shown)


crates/kubelet/src/store/composite/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+use crate::store::PullPolicy;
+use crate::store::Store;
+use async_trait::async_trait;
+use oci_distribution::Reference;
+use std::sync::Arc;
+
+/// A `Store` that has additional logic to determine if it can satisfy
+/// a particular reference. An `InterceptingStore` can be composed with
+/// another Store to satisfy specific requests in a custom way.
+pub trait InterceptingStore: Store {
```
thomastaylor312, line 12: Is an interceptor model better here or should we just have the separate "file system only" implementation of store? This feels like a lot of code (including 2 publicly exposed traits) for something that might be better as a the separate implementation

crates/wascc-provider/src/lib.rs
```diff
@@ -122,20 +123,20 @@ impl StopHandler for ActorHandle {
...
     handles: Arc<RwLock<HashMap<String, Handle<ActorHandle, LogHandleFactory>>>>,
-    store: S,
+    store: Arc<dyn Store + Sync + Send>,
     volume_path: PathBuf,
     log_path: PathBuf,
     kubeconfig: kube::Config,
     host: Arc<Mutex<WasccHost>>,
 }
 
-impl<S: Store + Send + Sync> WasccProvider<S> {
```
thomastaylor312, line 134: I would probably keep this constraint as is and then just have:
```rust
store: Arc<S>
```

crates/wasi-provider/src/lib.rs
```diff
@@ -56,18 +57,18 @@ const VOLUME_DIR: &str = "volumes";
...
+pub struct WasiProvider {
     handles: Arc<RwLock<HashMap<String, Handle<Runtime, wasi_runtime::HandleFactory>>>>,
-    store: S,
+    store: Arc<dyn Store + Sync + Send>,
     log_path: PathBuf,
     kubeconfig: kube::Config,
     volume_path: PathBuf,
 }
 
-impl<S: Store + Send + Sync> WasiProvider<S> {
```
thomastaylor312, line 67: Same here with the constraint

crates/kubelet/src/store/composite/mod.rs
itowlson: Just to clarify, there's no "filesystem only" implementation of `Store`.  There's a mode where you can get things from the filesystem _as well as_ registries.  But as written this doesn't turn off registries, because you may have some established modules in a registry and some under-development modules on the filesystem (this was certainly the case with the integration tests while the wasmerciser was under construction).

That aside, it's a good question: rather than offering a mechanism for arbitrary composition, should we just offer two Store implementations, registry-only and hybrid?  Internally the hybrid one would certainly still sit over the existing `FileStore` - no need to reimplement that stuff - and intercept certain requests, but it could be hardwired rather than generic.  And even if internally it was generic, we don't have to surface that composability to kubelet implementers.

I'm kind of agnostic about this.  It sort of irked me to build a one-off composite rather than a general mechanism for composition, which is why I ended up going this route.  But we have no other use case for composition, and as you observe it does increase the public API.

Anyone else have opinions?

crates/wascc-provider/src/lib.rs
itowlson: I'm not sure I see the advantage.  Conceptually the provider cares only that it _has_ a `Store`, not the _kind_ of `Store` - what do we gain by requiring it to be generic in the type of store?

crates/kubelet/src/config.rs
```diff
@@ -466,6 +474,13 @@ pub struct Opts {
         default_value = BOOTSTRAP_FILE
     )]
     bootstrap_file: PathBuf,
+
+    #[structopt(
+        long = "allow-local-modules",
+        env = "KRUSTLET_ALLOW_LOCAL_MODULES",
+        help = "Whether to allow loading modules directly from the filesystem"
```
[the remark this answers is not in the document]
itowlson: I've had a go at improving this - if it's still confusing let's put our heads together and wordsmith it.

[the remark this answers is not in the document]
thomastaylor312: tenor-21987896

crates/kubelet/src/store/composite/mod.rs
thomastaylor312: I am not sure if there are many other needs for a hybrid/composite repo in normal k8s land, but it could be really valuable here. I'm imagining fetching modules from various sources for IoT/Edge devices (even the current FS + registries model is great for that). 

@bacongobbler You have any opinions here?

crates/wascc-provider/src/lib.rs
thomastaylor312: Mostly I think of it as describing that we have a Provider of the given type with the store. A `WasccProvider<FileStore>` is a distinct provider and type from a `WasccProvider<InterceptingStore>` in my mind. However, using generics also avoids dynamic dispatching as it will have a concrete type

crates/kubelet/src/store/composite/mod.rs
technosophos: To be honest, OCI registry is not an ideal storage format for WASM modules. And we've seen CapOne and Wasmer already introduce competing module distribution services. So there is a good chance that at some point in the not-too-distant future we are going to need to be able to fetch modules from other network-centric storage systems.

squillace:
This is interesting. At this point, while I wouldn't prevent file system direct loading, no production or debug system should in theory use anything other than some "network connected storage" like OCI registries or other, similar but more specific types per @technosophos (and I agree with the importance of his point).

But I am also trying to think of anything that isn't "developer laziness" to NOT push a wasm while debugging. Mind you, I'M THAT LAZY DEVELOPER, so I get it. But what scenarios other than the lazy dev one could such a file-system loading be useful? @itowlson could you help out with the precise user case for the switch that isn't the lazy dev -- or in this case, the lazy Ralph?

crates/kubelet/src/store/composite/mod.rs
thomastaylor312: Ok, then I think composeability is the right course here @itowlson. Let's keep it as is

crates/wascc-provider/src/lib.rs
itowlson: image

itowlson:
@squillace There's a bit of "do I really want to pollute the registry with 800 not-working versions of the same application just because of this one damned bug I'm chasing down."  But yeah, mostly it's the lazy dev, or, to use the official marketing term, "lowering the barrier to rapid iteration and a tight inner development loop."

squillace:
what tight inner dev loop do we need here? tightly developing the krustlet inner loop, or the module dev inner loop? how long does it take to push a module?

crates/kubelet/src/config.rs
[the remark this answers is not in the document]
thomastaylor312: I think we just stick with this as is. We can always change it later

bacongobbler:
I get the use case here. But this same use case can technically be accomplished by using a pull policy of `Never` and just overwriting the module in `~/.krustlet/.oci` each time you rebuild the module. You cannot read from an arbitrary directory in the filesystem, but it does fix the use case where you're concerned about polluting a container registry with 800+ modules you wrote during development.

My main concern with this flag: this could introduce security issues where a malicious attacker could load arbitrary modules they should not have access to. However, this logic is all behind a feature flag, so I'm not too concerned about bringing this in. I figured it should be noted somewhere though.

Is this feature flag something the kubelet project supports as well? I'm curious how they handle development with containers.

thomastaylor312:
I know kind and minikube (at least I think so) have a way to push images to the node's local registry or something of the sort. I don't think normal kubelet has a flag like this. Either way, I think the composable store is a good thing to have around

bacongobbler:
If this is targeted towards developers, would it be worth writing something down in the developer documentation about this feature flag?

thomastaylor312:
The flag definitely is for devs or specific cases where you want to load from the filesystem

squillace:
I am officially concerned about flags enabling a circumvention of the registry system, but I LOVE the composability aspect of potential registries. But merge anyway, and we'll have a better look at it.

thomastaylor312:
After talking in person, we decided that the flag is going to be temporary (and documented as such) and we like the idea of having the composable stores. So we will merge this and then add more tooling to remove the need for a local file store option

docs/topics/configuration.md
```diff
@@ -70,3 +75,19 @@ configuration file, for example by writing `MAX_PODS=200 krustlet-wascc` or
...
+
+## Notes to kubelet implementers
+
+Some flags require you to support them in your provider or main code - they are
+not implemented automatically by the kubelet core. These flags are as follows:
+
+* `--bootstrap-file` - should be passed to `kubelet::bootstrap` if you use the
+  bootstrapping feature
+* `--data-dir` - this should be used to construct the `FileStore` if you use one
+* `--allow-local-modules` - if specified you should compose a `FileSystemStore`
```
bacongobbler, line 87: is this now `--x-allow-local-modules`?

docs/topics/configuration.md
```diff
@@ -70,3 +75,19 @@ configuration file, for example by writing `MAX_PODS=200 krustlet-wascc` or
...
+  onto your normal store
+
+See the `krustlet-wasi.rs` file for examples of how to honour these flags.
+
+If you can't honour a flag value in your particular scenario, then you should
+still check for it and return an error, rather than silently ignoring it.
```
bacongobbler, line 93: 👍 I like the call to action here for developers on how to properly handle this behaviour.

docs/topics/configuration.md
itowlson: **doh**

itowlson:
Fixed that docs glitch @bacongobbler - please re-review.

PR #319, opened 2020-07-09
Fix kubelet exit on module failure

itowlson:
There is an issue where if a WASM module fails (for example, with Rust source, if `main` returns an `Err`), it takes down the entire `krustlet-wasi` process.  This fixes that issue: now, if a module fails, the pod will terminate with a nonzero exit code (the exact value will depend on the WASM language binding but in Rust it appears to default to 1).

NOTE: The bad behaviour was down to an issue in wasmtime 0.16: a Rust error return was compiled into a call to a runtime-provided `proc_exit` function, which simply called the Rust stdlib function `std::process::exit`.  This was fixed in 0.18 - `proc_exit` now raises a trap, which the wasmtime CLI handles by exiting, but which we simply treat as a failure.  However, we can't use 0.18 because it doesn't allow us to redirect stdout and stderr to temp files (or if it does then I can't figure out how to do it).  So for now this uses an arbitrary commit of wasmtime from GitHub; hopefully when 0.19 comes out it will not have changed again and we can reset the reference back to that!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,12 +290,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.12.1"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a49806b9dadc843c61e7c97e72490ad7f7220ae249012fbda9ad0609457c0543"
+checksum = "1b6a2d3371669ab3ca9797670853d61402b03d0b4b9ebf33d677dfa720203072"
 dependencies = [
- "gimli 0.21.0",
+ "gimli 0.22.0",
 ]
+
+[[package]]
+name = "adler"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ccc9a9dd069569f212bc4330af9f17c4afb5e8ce185e83dbb14f1349dda18b10"
 
 [[package]]
 name = "adler32"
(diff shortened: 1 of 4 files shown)


peterhuene:
Hmm, it seems that bytecodealliance/wasmtime#1600 was tagged in 0.18.0, so I'm not sure why it wouldn't work for your changes, but I haven't looked too closely at it.

peterhuene:
👍  re: the wasi provider changes.

itowlson:
@peterhuene Hmm. Something weird is going on! When I download the source of the 0.18 release from GitHub, it includes the changes... but when I specify 0.18 in `Cargo.toml`, the code that gets pulled into the Cargo registry does _not_ include those changes!

Thanks for spotting - maybe I just have a corrupted registry or something...

peterhuene:
I'll take a quick look at the published crate.

itowlson:
The published crate doesn't seem to have the changes, unless I'm looking in the wrong place...

peterhuene:
Sure enough, wasi-common `0.18.0` doesn't appear to have the changes.  I'll talk to folks here to figure out why the change was tagged as `0.18.0` if it didn't end up in the published crate.

itowlson:
Thanks @peterhuene, much appreciated!

thomastaylor312:
@itowlson Would you still like a review on this or do you want to wait until we find out some more from @peterhuene (Thanks again btw!)?

peterhuene:
The current theory is that the tag is wrong.  It looks like `wasi-common` was published on the same day that bytecodealliance/wasmtime#1600 went in.  I think it probably missed that commit.  The `wasmtime` crate, oddly, was published a few days later.  I think the tag is wrong.

I would suggest taking the commit reference as a workaround until `0.19.0` is published, which will hopefully be soon as there's a really important Windows unwind fix I'd like out there.

itowlson:
@thomastaylor312 It would be good to get the code reviewed, with the obvious exception of "you can't take a dependency on some random commit you just can't" _grin_.  Depending on whether 0.19 is imminent we can then make the call of whether to temporarily take the nasty dependency or hold off merging until there is a formal release.  But it would be nice to have the code reviewed and ready to go.  Thanks!

PR #320, opened 2020-07-09
bump wascc-host to 0.10

bacongobbler:
closes #315

PR #321, opened 2020-07-09
Feat/windows support

### LINKED ISSUE #20: Windows support
Currently, trying to run the build on a clean Windows machine fails (see [action run](https://github.com/radu-matei/krustlet/runs/481109377) and [workflow definition](https://github.com/radu-matei/krustlet/actions/runs/48502584/workflow)):

```
2020-03-03T00:27:54.3571851Z error: failed to run custom build command for `openssl-sys v0.9.53`
2020-03-03T00:27:54.3571945Z 
2020-03-03T00:27:54.3572085Z Caused by:
2020-03-03T00:27:54.3572259Z   process didn't exit successfully: `D:\a\krustlet\krustlet\target\debug\build\openssl-sys-97d80117f436f29e\build-script-main` (exit code: 101)
2020-03-03T00:27:54.3572584Z --- stdout
2020-03-03T00:27:54.3572658Z cargo:rustc-cfg=const_fn
2020-03-03T00:27:54.3572786Z cargo:rerun-if-env-changed=X86_64_PC_WINDOWS_MSVC_OPENSSL_LIB_DIR
2020-03-03T00:27:54.3572911Z X86_64_PC_WINDOWS_MSVC_OPENSSL_LIB_DIR unset
2020-03-03T00:27:54.3573028Z cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
2020-03-03T00:27:54.3573102Z OPENSSL_LIB_DIR unset
2020-03-03T00:27:54.3573226Z cargo:rerun-if-env-changed=X86_64_PC_WINDOWS_MSVC_OPENSSL_INCLUDE_DIR
2020-03-03T00:27:54.3573353Z X86_64_PC_WINDOWS_MSVC_OPENSSL_INCLUDE_DIR unset
2020-03-03T00:27:54.3573477Z cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
2020-03-03T00:27:54.3573552Z OPENSSL_INCLUDE_DIR unset
2020-03-03T00:27:54.3573676Z cargo:rerun-if-env-changed=X86_64_PC_WINDOWS_MSVC_OPENSSL_DIR
2020-03-03T00:27:54.3573807Z X86_64_PC_WINDOWS_MSVC_OPENSSL_DIR unset
2020-03-03T00:27:54.3573929Z cargo:rerun-if-env-changed=OPENSSL_DIR
2020-03-03T00:27:54.3574001Z OPENSSL_DIR unset
2020-03-03T00:27:54.3574166Z note: vcpkg did not find openssl as libcrypto and libssl: Could not find Vcpkg tree: No vcpkg.user.targets found. Set the VCPKG_ROOT environment variable or run 'vcpkg integrate install'
2020-03-03T00:27:54.3574355Z note: vcpkg did not find openssl as ssleay32 and libeay32: Could not find Vcpkg tree: No vcpkg.user.targets found. Set the VCPKG_ROOT environment variable or run 'vcpkg integrate install'
2020-03-03T00:27:54.3574468Z 
2020-03-03T00:27:54.3574537Z --- stderr
2020-03-03T00:27:54.3574650Z thread 'main' panicked at '
2020-03-03T00:27:54.3574738Z 
2020-03-03T00:27:54.3574819Z Could not find directory of OpenSSL installation, and this `-sys` crate cannot
2020-03-03T00:27:54.3574955Z proceed without this knowledge. If OpenSSL is installed and this crate had
2020-03-03T00:27:54.3575090Z trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
2020-03-03T00:27:54.3575213Z compilation process.
2020-03-03T00:27:54.3575255Z 
2020-03-03T00:27:54.3575482Z Make sure you also have the development packages of openssl installed.
2020-03-03T00:27:54.3575582Z For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.
2020-03-03T00:27:54.3575670Z 
2020-03-03T00:27:54.3575784Z If you're in a situation where you think the directory *should* be found
2020-03-03T00:27:54.3575940Z automatically, please open a bug at https://github.com/sfackler/rust-openssl
2020-03-03T00:27:54.3576089Z and include information about your system as well as this message.
2020-03-03T00:27:54.3576181Z 
2020-03-03T00:27:54.3576252Z $HOST = x86_64-pc-windows-msvc
2020-03-03T00:27:54.3576368Z $TARGET = x86_64-pc-windows-msvc
2020-03-03T00:27:54.3576484Z openssl-sys = 0.9.53
2020-03-03T00:27:54.3576524Z 
2020-03-03T00:27:54.3576612Z 
2020-03-03T00:27:54.3576689Z It looks like you're compiling for MSVC but we couldn't detect an OpenSSL
2020-03-03T00:27:54.3576825Z installation. If there isn't one installed then you can try the rust-openssl
2020-03-03T00:27:54.3576967Z README for more information about how to download precompiled binaries of
2020-03-03T00:27:54.3577074Z OpenSSL:
2020-03-03T00:27:54.3577113Z 
2020-03-03T00:27:54.3577243Z https://github.com/sfackler/rust-openssl#windows
```

Note that the static compilation of the OCI library might also affect the ability to support Windows.

thomastaylor312:
This PR implements full Windows functionality and build/release pipelines. It is by no means perfect, but it is a good start for us to build on. Below is a full list of what this PR contains:

- A working `justfile` for Windows
- A Powershell bootstrap script
- Refactors to how we are handling waiting on futures. The large structs and env variables needed for the config are now boxed as to avoid stack allocation. Futures are no longer pinned to the stack either. This avoids large stack usage and a stack overflow on Windows
- Documentation and caveats for the new windows builds
- A build, test, and release pipeline for Windows (I tested the release pipeline [here](https://github.com/deislabs/krustlet/actions/runs/176256544)) .

NOTE: This is missing e2e tests. This is due to rustls not supporting IP SANs. I've opened #329 to track this

Closes #20

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,11 +8,28 @@ on:
 jobs:
   build:
     name: build release assets
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.config.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest]
-        arch: [amd64]
+        config:
+          - {
+              os: "ubuntu-latest",
+              arch: "amd64",
+              args: "--release",
+              extension: "",
+            }
+          - {
+              os: "macos-latest",
+              arch: "amd64",
+              args: "--release",
+              extension: "",
+            }
+          - {
+              os: "windows-latest",
+              arch: "msvc",
+              args: "--release --no-default-features --features rustls-tls",
+              extension: ".exe",
+            }
     steps:
       - uses: actions/checkout@v2
 
@@ -22,9 +39,11 @@ jobs:
 
       - name: set the release version (master)
         if: github.ref == 'refs/heads/master'
+        shell: bash
         run: echo ::set-env name=RELEASE_VERSION::canary
 
       - name: lowercase the runner OS name
+        shell: bash
         run: |
           OS=$(echo "${{ runner.os }}" | tr '[:upper:]' '[:lower:]')
           echo ::set-env name=RUNNER_OS::$OS
@@ -33,20 +52,20 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release
+          args: ${{ matrix.config.args }}
 
       - name: package release assets
+        shell: bash
         run: |
           mkdir _dist
-          cp README.md LICENSE target/release/krustlet-wasi target/release/krustlet-wascc _dist/
+          cp README.md LICENSE target/release/krustlet-wasi${{ matrix.config.extension }} target/release/krustlet-wascc${{ matrix.config.extension }} _dist/
           cd _dist
-          tar czf krustlet-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.arch }}.tar.gz README.md LICENSE krustlet-wasi krustlet-wascc
+          tar czf krustlet-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}.tar.gz README.md LICENSE krustlet-wasi krustlet-wascc
 
       - uses: actions/upload-artifact@v1
         with:
           name: krustlet
-          path: _dist/krustlet-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.arch }}.tar.gz
-
+          path: _dist/krustlet-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}.tar.gz
   publish:
     name: publish release assets
     runs-on: ubuntu-latest
@@ -55,28 +74,23 @@ jobs:
       - name: set the release version
         if: startsWith(github.ref, 'refs/tags/v')
         run: echo ::set-env name=RELEASE_VERSION::${GITHUB_REF/refs\/tags\//}
-
       - name: set the release version
         if: github.ref == 'refs/heads/master'
         run: echo ::set-env name=RELEASE_VERSION::canary
-
       - name: download release assets
         uses: actions/download-artifact@v1
         with:
           name: krustlet
-
       - name: generate checksums
         run: |
           cd krustlet
           sha256sum * > checksums-${{ env.RELEASE_VERSION }}.txt
-
       - name: upload to azure
         uses: bacongobbler/azure-blob-storage-upload@v1.0.0
         with:
           source_dir: krustlet
           container_name: releases
           connection_string: ${{ secrets.AzureStorageConnectionString }}
-
   crates:
     name: publish to crates.io
     runs-on: ubuntu-latest
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -45,12 +45,14 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
     pub async fn new(
         provider: T,
         kube_config: kube::Config,
-        config: Box<Config>,
+        config: Config,
     ) -> anyhow::Result<Self> {
         Ok(Self {
             provider: Arc::new(provider),
             kube_config,
-            config,
+            // The config object can get a little bit for some reason, so put it
+            // on the heap
+            config: Box::new(config),
         })
     }
 
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -36,7 +36,7 @@ use tokio::sync::mpsc;
 pub struct Kubelet<P> {
     provider: Arc<P>,
     kube_config: kube::Config,
-    config: Config,
+    config: Box<Config>,
 }
 
 impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
@@ -45,7 +45,7 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
     pub async fn new(
         provider: T,
         kube_config: kube::Config,
-        config: Config,
+        config: Box<Config>,
     ) -> anyhow::Result<Self> {
         Ok(Self {
             provider: Arc::new(provider),
@@ -79,10 +79,8 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
         let node_updater = start_node_updater(client.clone(), self.config.node_name.clone()).fuse();
 
         // If any of these tasks fail, we can initiate graceful shutdown.
-        let services = async {
-            futures::pin_mut!(webserver, error_handler, node_updater, signal_task);
-
-            futures::select! {
+        let services = Box::pin(async {
+            tokio::select! {
                 res = error_handler => if let Err(e) = res {
                     error!("Error handler task completed with error: {:?}", &e);
                 },
@@ -97,7 +95,7 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
             // Use relaxed ordering because we just need other tasks to eventually catch the signal.
             signal.store(true, Ordering::Relaxed);
             Ok::<(), anyhow::Error>(())
-        };
+        });
 
         // Periodically checks for shutdown signal and cleans up resources gracefully if caught.
         let signal_handler = start_signal_handler(
@@ -118,10 +116,8 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
         .fuse();
 
         // These must all be running for graceful shutdown. An error here exits ungracefully.
-        let core = async {
-            futures::pin_mut!(pod_informer, signal_handler);
-
-            futures::select! {
+        let core = Box::pin(async {
+            tokio::select! {
                 res = signal_handler => res.map_err(|e| {
                     error!("Signal handler task joined with error {:?}", &e);
                     e.into()
@@ -131,12 +127,12 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
                     e.into()
                 })
             }
-        };
+        });
 
         // Services will not return an error, so this will wait for both to return, or core to
         // return an error. Services will return if signal is set because pod_informer will drop
         // error_sender and error_handler will exit.
-        futures::try_join!(core, services)?;
+        tokio::try_join!(core, services)?;
         Ok(())
     }
 }
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -120,11 +120,11 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
             tokio::select! {
                 res = signal_handler => res.map_err(|e| {
                     error!("Signal handler task joined with error {:?}", &e);
-                    e.into()
+                    e
                 }),
                 res = pod_informer => res.map_err(|e| {
                     error!("Pod informer task joined with error {:?}", &e);
-                    e.into()
+                    e
                 })
             }
         });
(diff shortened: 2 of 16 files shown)


thomastaylor312:
This is getting close, but I need to upgrade the version of `wascc-fs` to solve this. To do that, I will need the changes in #320, which is currently blocked

bacongobbler:
#320 has been merged.

crates/kubelet/src/kubelet.rs
```diff
@@ -36,7 +36,7 @@ use tokio::sync::mpsc;
 pub struct Kubelet<P> {
     provider: Arc<P>,
     kube_config: kube::Config,
-    config: Config,
+    config: Box<Config>,
```
rylev, line 39: Is it necessary to take the argument as a box? I would think boxing at the beginning of the function would also work, no?

docs/community/developers.md
```diff
@@ -66,6 +62,24 @@ sudo apt-get install pkg-config
 run on WSL2 using the WSL2-enabled Docker Kubernetes or Azure Kubernetes.  If you're
 on WSL1 you may be better off running in a full Linux VM under Hyper-V.
 
+### Building on Windows
+
+As of version 0.4, we have support for building on Windows. For convenience sake, there is a windows
+version of the justfile called `justfile-windows`. This justfile uses Powershell and has the proper
```
rylev, line 68: Nit: Powershell is written PowerShell

crates/kubelet/src/kubelet.rs
thomastaylor312: When I tested this, I had to box it as it was coming out, but I made a few more changes since then, so I can try again. Currently my windows computer is having problems though, so I'll need someone else to test

bacongobbler: I think this is because of a borrow.

After removing the `Box`es from a variety of locations:

```
PS C:\Users\me\tmp\krustlet> cargo build --no-default-features --features rustls-tls
   Compiling wasi-provider v0.3.0 (C:\Users\me\tmp\krustlet\crates\wasi-provider)
   Compiling wascc-provider v0.3.0 (C:\Users\me\tmp\krustlet\crates\wascc-provider)
   Compiling krustlet v0.3.0 (C:\Users\me\tmp\krustlet)
error[E0308]: mismatched types
  --> src/krustlet-wascc.rs:24:46
   |
24 |     let provider = WasccProvider::new(store, &config, kubeconfig.clone()).await?;
   |                                              ^^^^^^^
   |                                              |
   |                                              expected struct `kubelet::config::Config`, found `&kubelet::config::Config`
   |                                              help: consider removing the borrow: `config`
```

removing the borrow resulted in the following build error:

```
PS C:\Users\me\tmp\krustlet> cargo build --no-default-features --features rustls-tls
   Compiling krustlet v0.3.0 (C:\Users\me\tmp\krustlet)
error[E0382]: use of moved value: `config`
  --> src/krustlet-wascc.rs:25:54
   |
12 |     let config = Config::new_from_file_and_flags(
   |         ------ move occurs because `config` has type `kubelet::config::Config`, which does not implement the `Copy` trait
...
24 |     let provider = WasccProvider::new(store, config, kubeconfig.clone()).await?;
   |                                              ------ value moved here
25 |     let kubelet = Kubelet::new(provider, kubeconfig, config).await?;
   |                                                      ^^^^^^ value used here after move
```

This is because the config is being borrowed by both the provider and the kubelet. There's likely some code cleanup we could do here.

In either case, cloning the config object before passing to the provider fixes the issue.

here's my diff (missed a few spots in the docstrings, but the code's all there):

```
diff --git a/crates/kubelet/src/kubelet.rs b/crates/kubelet/src/kubelet.rs
index fe0252d..0dc96c4 100644
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -36,7 +36,7 @@ use tokio::sync::mpsc;
 pub struct Kubelet<P> {
     provider: Arc<P>,
     kube_config: kube::Config,
-    config: Box<Config>,
+    config: Config,
 }
 
 impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
@@ -45,7 +45,7 @@ impl<T: 'static + Provider + Sync + Send> Kubelet<T> {
     pub async fn new(
         provider: T,
         kube_config: kube::Config,
-        config: Box<Config>,
+        config: Config,
     ) -> anyhow::Result<Self> {
         Ok(Self {
             provider: Arc::new(provider),
diff --git a/crates/wascc-provider/src/lib.rs b/crates/wascc-provider/src/lib.rs
index 6b1ebc0..012ae36 100644
--- a/crates/wascc-provider/src/lib.rs
+++ b/crates/wascc-provider/src/lib.rs
@@ -147,7 +147,7 @@ impl WasccProvider {
     /// (including creating it if necessary)
     pub async fn new(
         store: Arc<dyn Store + Sync + Send>,
-        config: &Box<kubelet::config::Config>,
+        config: kubelet::config::Config,
         kubeconfig: kube::Config,
     ) -> anyhow::Result<Self> {
         let host = Arc::new(Mutex::new(WasccHost::new()));
diff --git a/src/krustlet-wascc.rs b/src/krustlet-wascc.rs
index 8dd4a29..8043408 100644
--- a/src/krustlet-wascc.rs
+++ b/src/krustlet-wascc.rs
@@ -9,10 +9,10 @@ use wascc_provider::WasccProvider;
 async fn main() -> anyhow::Result<()> {
     // The provider is responsible for all the "back end" logic. If you are creating
     // a new Kubelet, all you need to implement is a provider.
-    let config = Box::new(Config::new_from_file_and_flags(
+    let config = Config::new_from_file_and_flags(
         env!("CARGO_PKG_VERSION"),
         None,
-    ));
+    );
 
     // Initialize the logger
     env_logger::init();
@@ -21,7 +21,7 @@ async fn main() -> anyhow::Result<()> {
 
     let store = make_store(&config);
 
-    let provider = WasccProvider::new(store, &config, kubeconfig.clone()).await?;
+    let provider = WasccProvider::new(store, config.clone(), kubeconfig.clone()).await?;
     let kubelet = Kubelet::new(provider, kubeconfig, config).await?;
     kubelet.start().await
 }
diff --git a/src/krustlet-wasi.rs b/src/krustlet-wasi.rs
index 8266ef1..5b8d67d 100644
--- a/src/krustlet-wasi.rs
+++ b/src/krustlet-wasi.rs
@@ -9,10 +9,10 @@ use wasi_provider::WasiProvider;
 async fn main() -> anyhow::Result<()> {
     // The provider is responsible for all the "back end" logic. If you are creating
     // a new Kubelet, all you need to implement is a provider.
-    let config = Box::new(Config::new_from_file_and_flags(
+    let config = Config::new_from_file_and_flags(
         env!("CARGO_PKG_VERSION"),
         None,
-    ));
+    );
 
     // Initialize the logger
     env_logger::init();
```

docs/intro/install.md
```diff
@@ -5,13 +5,15 @@ This guide shows how to install Krustlet.
...
+compiled releases can be manually downloaded and installed. Please note these instructions will work
+on Linux, MacOS, and Windows (in Powershell)
 
 1. Download your desired version from [the releases
    page](https://github.com/deislabs/krustlet/releases)
 1. Unpack it (`tar -xzf krustlet-v0.3.0-linux-amd64.tar.gz`)
 1. Find the desired Krustlet provider in the unpacked directory, and move it to its desired
-   destination (`mv krustlet-wasi /usr/local/bin/`)
+   destination somewhere in your `$PATH` (e.g. `mv krustlet-wasi /usr/local/bin/` on unix-like
+   systems or `mv krustlet-wasi.exe C:\Windows\system32\` on Windows)
```
bacongobbler, line 16: I think it'd be better to place it under `C:\\Windows` rather than `system32\`. system32 should be reserved for system tools, much like how we don't touch `/bin` in unix.

docs/intro/install.md
```diff
@@ -20,6 +22,13 @@ find Krustlet, check to make sure that your `$PATH` environment variable is set
 If you'd like to validate the download, checksums can be downloaded from
 https://krustlet.blob.core.windows.net/releases/checksums-v0.3.0.txt
 
+### Windows
+As of Krustlet 0.4, there are now Windows builds available. However, there are some caveats. The
```
bacongobbler, line 26: can you add a newline character between the heading and the paragraph?

crates/kubelet/src/kubelet.rs
thomastaylor312: Ok, I double checked this as well and the box is no longer needed. It has been removed

docs/community/developers.md
thomastaylor312: Fixed

docs/intro/install.md
thomastaylor312: I couldn't find a good equivalent to `/usr/local/bin`. `C:\Windows` isn't in the default path (at least on my recent fresh install of windows 10. It did have `%USERPROFILE%\AppData\Local\Microsoft\WindowsApps`, but I am not sure if that exists on every machine. What do you think would be the best option?

.github/workflows/release.yml
```diff
@@ -8,11 +8,28 @@ on:
...
+          - {
+              os: "macos-latest",
+              arch: "amd64",
+              args: "--release",
+              extension: "",
+            }
+          - {
+              os: "windows-latest",
+              arch: "msvc",
+              args: "--release --no-default-features --features rustls-tls",
```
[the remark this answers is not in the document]
thomastaylor312: So we can _maybe_ make it work with things like `[target.'cfg(windows)'.dependencies]`, but the problem is that we want people on *nix based systems to be able to select rustls as well, and I don't know if you can have `features` be OS specific. To put it another way, it seems to be difficult to have this as a feature as well as a conditional dependency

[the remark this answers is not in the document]
bacongobbler: good point. I couldn't find another way to do this either. This looks good to go :)

docs/intro/install.md
bacongobbler: So here's what I find in my system variables (slightly different than user variables) for PATH: 

image

`%SystemRoot%` equates to C:\\ on my system:

```
PS C:\Users\me> echo $Env:SystemRoot
C:\Windows
```

And it looks like there are a few user packages I installed that went there:

image

I do not truly know what is considered the equivalent of `/usr/local/bin` for Windows, unfortunately. Maybe there isn't one. Most self-installers I've seen add their install directory to the user PATH. Git for Windows, VSCode, Docker...

Maybe we need to look into providing a self-installer for Windows to resolve this issue, and recommend using `winget` or downloading the self-installer directly?

Those don't seem like action items we need to make here to address this issue.

Anecdotally though, my `%SystemRoot%\system32` does not appear to have any apps I installed recently. They all seem to be Windows DLLs or system executables like `print.exe` and `replace.exe`. If we infer that system32 is meant for OS executables, I do not think that is the correct location we should recommend to users to move their executables.

Did you check your system environment variables?

image

bacongobbler: I do appear to have WindowsApps as part of my default user path, though I wonder if that's meant to be reserved for UWP apps from the Microsoft Store.

image

bacongobbler: What I have done in the past has created a `bin/` directory in my `%USERPROFILE%`, then added that to my user environment variables.

https://superuser.com/questions/983138/what-is-the-equivalent-of-the-bin-directory-for-windows/983171 seems to have the same answer here too. There is no real equivalent here. It's a convention for Linux users, but there doesn't appear to be a standard convention for Windows.

thomastaylor312: Yeah, so I think a windows installer might be in order here. My environment variables were quite bare. I only had 3 different items in my PATH, and one of those was from Cargo

PR #330, opened 2020-07-23
Modularise WASI integration tests

itowlson:
This PR breaks up the single WASI megatest into a bunch of more granular test functions, currently one per pod used in the testing.  The tests are now able to run in parallel and are kept separate through namespaces.  The pattern for creating new tests is intended to be clear and to require minimal programmer effort, so that we will hopefully be inclined to write new tests for new things and not bolt things onto existing tests out of laziness!

I also moved a bunch of stuff out of the main test file into helper files - some of this, such as the `TestResourceManager` is behavioural, but other parts are just moving things to make the main file a bit more navigable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ rustls-tls = ["kube/rustls-tls", "kubelet/rustls-tls", "wasi-provider/rustls-tls
 
 [dependencies]
 anyhow = "1.0"
-tokio = { version = "0.2", features = ["macros", "rt-threaded"] }
+tokio = { version = "0.2", features = ["macros", "rt-threaded", "time"] }
 kube = { version= "0.35", default-features = false }
 env_logger = "0.7"
 kubelet = { path = "./crates/kubelet", version = "0.3", default-features = false, features = ["cli"] }
--- a/tests/assert.rs
+++ b/tests/assert.rs
@@ -45,22 +45,6 @@ pub async fn pod_container_log_contains(
     Ok(())
 }
 
-// pub async fn pod_log_does_not_contain(
-//     pods: &Api<Pod>,
-//     pod_name: &str,
-//     unexpected_log: &str,
-// ) -> anyhow::Result<()> {
-//     let logs = pods.logs(pod_name, &LogParams::default()).await?;
-//     assert!(
-//         !logs.contains(unexpected_log),
-//         format!(
-//             "Expected log NOT containing {} but got {}",
-//             unexpected_log, logs
-//         )
-//     );
-//     Ok(())
-// }
-
 pub async fn pod_exited_successfully(pods: &Api<Pod>, pod_name: &str) -> anyhow::Result<()> {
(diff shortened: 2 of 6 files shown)


tests/assert.rs
```diff
@@ -0,0 +1,147 @@
...
+    log_params.container = Some(container_name.to_owned());
+    let logs = pods.logs(pod_name, &log_params).await?;
+    assert!(
+        logs.contains(expected_log),
+        format!("Expected log containing {} but got {}", expected_log, logs)
+    );
+    Ok(())
+}
+
+// pub async fn pod_log_does_not_contain(
```
bacongobbler, line 48: was this intended to be left in, or should we remove this?

Cargo.toml
```diff
@@ -40,7 +40,7 @@ rustls-tls = ["kube/rustls-tls", "kubelet/rustls-tls", "wasi-provider/rustls-tls
 
 [dependencies]
 anyhow = "1.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "0.2", features = ["macros", "time"] }
```
thomastaylor312, line 43: This is 2020, I don't think we should be messing with time 😜

tests/assert.rs
```diff
@@ -0,0 +1,147 @@
...
+            .clone()
+    })()
+    .expect("Could not fetch terminated states");
+    assert_eq!(state.exit_code, 1);
+
+    Ok(())
+}
+
+pub async fn container_file_contains(
+    pod_name: &str,
```
thomastaylor312, line 128: Super nit: You could probably make all the `&str` references in these helper functions be `AsRef<str>` for more flexibility. So:

```rust
pub async fn container_file_contains<T: AsRef<str>>(
     pod_name: T,
     pod_namespace: T,
     container_file_path: T,
     expected_content: T,
     file_error: T,
)
```

However, the `&str` here are probably fine

tests/test_resource_manager.rs
```diff
@@ -0,0 +1,220 @@
...
+                        "metadata": {
+                            "name": namespace
+                        },
+                        "spec": {}
+                }))?,
+            )
+            .await?;
+
+        // k8s seems to need a bit of time for namespace permissions to flow
+        // through the system.  TODO: make this less worse
```
thomastaylor312, line 77: This could probably be checked in the status to make sure `phase` == `Active` like we do for other waits. Not a big deal for this PR though

tests/assert.rs
itowlson: It's from a test that I was meaning to implement for the init containers work but didn't get straight in my mind.  I would kind of like to come back to it so hadn't purged it, but it would be no loss to get rid of it; it wasn't quite right for what I wanted, and would be easy to reconstruct.

tests/assert.rs
itowlson: After looking at this, I don't think it's worth it in this case, though it will definitely be a useful technique to have in other places - thanks!

PR #332, opened 2020-07-24
Document tolerations and node affinities for WASM

itowlson:
Because I done got bitten and I don't want other folks to get bitten too.

### DIFF
--- a/docs/howto/wasm.md
+++ b/docs/howto/wasm.md
@@ -1,0 +1,57 @@
+# Running Web Assembly (WASM) workloads in Kubernetes
+
+The Krustlet repository contains two projects, `krustlet-wasi` and
+`krustlet-wascc`, for running WASM workloads in Kubernetes. These kubelets
+run workloads implemented as Web Assembly (WASM) modules rather than
+as OCI containers. Each running instance appears to Kubernetes as a
+node; Kubernetes schedules work to the instance as to any other node.
+
+## Running WASM modules on the right kubelet
+
+WASM modules are not interchangeable with OCI containers: `krustlet-wasi`
+and `krustlet-wascc` can't run OCI containers, and normal OCI nodes
+can't run WASM modules. In order to run your WASM workloads on the right
+nodes, you should use the Kubernetes tolerations system; in some cases you
+will also need to use node affinity.
+
+The `krustlet-wasi` and `krustlet-wascc` 'virtual nodes' both have
+`NoExecute` taints with the key `krustlet/arch` and a provider-defined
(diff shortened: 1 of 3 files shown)


docs/howto/wasm.md
```diff
@@ -0,0 +1,57 @@
...
+kind: Pod
+metadata:
+  name: hello-wasm
+spec:
+  containers:
+  - name: hello-wasm
+    image: webassembly.azurecr.io/hello-wasm:v1
+  tolerations:
+  - effect: NoExecute
+    key: krustlet/arch
```
DerekStrickland, line 33: Doesn't this need to match the nodeSelector

```suggestion
    key: kubernetes.io/arch
```

itowlson: @DerekStrickland Good question. I used `kubernetes.io/arch` for the node selector in the integration tests and it worked, because this is set in the kubelet `node` module.  I copied the tolerations from the integration tests but they were using `krustlet/arch`, which looks like it is set separately by the WASI and WASCC providers.  I agree it seems like `kubernetes.io/arch` should work in both places.

@bacongobbler @thomastaylor312 What is the distinction between `kubernetes.io/arch` and `krustlet/arch`?  Is one preferred over the other?

thomastaylor312: `krustlet/arch` is different in this case. `kubernetes.io/arch` is supposed to indicate the actual OS (like Windows and Linux) as opposed to the type of WASM modules. However, now that we have other provider that are not wasm, maybe we do need to retire `krustlet/arch` and stick with `kubernetes.io/arch`

thomastaylor312: Yeah, thinking about it some more, we should probably standardize on `kubernetes.io/arch` now. That will require a small update to code @itowlson. Did you want to handle that in this PR or would you rather not?

itowlson: @thomastaylor312 I can take that for sure.  Thanks!

itowlson: @DerekStrickland Closing the loop on this... per discussion with @thomastaylor312, I had a look at rationalising the taint key to be the same as the node label key.  However, I realised that it would have a drastic compatibility impact - existing pod specs would need to be modified to tolerate the new taint, and we couldn't think of a way to run a transition period.  For the time being, the plan is to go ahead with this PR just documenting what the current situation is, and I've raised https://github.com/deislabs/krustlet/issues/334 to discuss how we might progress this.  Thanks for flagging it!

DerekStrickland: No worries.  It was a lever I pulled trying to solve my FailedScheduling woes, and thought I'd bring it up.

Issue #318, opened 2020-07-08
Improve local e2e testing experience


thomastaylor312:
*User story*:

As a user, I would like to easily run the e2e tests without fiddling around and have test output that lets me know exactly what went wrong

*Requirements*:

- A single `just` target that spins up and automatically configures both providers (including approving certs if necessary) and then runs the tests
- If a test fails due to a pod issue (not an output assertion), then output relevant details (like the pod status and the log if possible) before failing

thomastaylor312:
This has been closed by the issues mentioned above

PR #336, opened 2020-07-29
Notify user if they need to approve a bootstrap CSR

itowlson:
This adds a notification to the bootstrap process to inform the user when they need to approve a CSR.  It serves two purposes:

1. Currently if you do `bootstrap.sh` and then, say, `just run-wasi`, the kubelet just sits there but all your tests fail because it's waiting on TLS cert approval.  Now the user will have a visual prompt of what they need to do.

2. In preparation for #318, it provides a point in the output where a test harness can know that the CSR is ready, and can extract the CSR name in order to approve it.

Note that the actual interaction is NOT provided by the `kubelet` crate, either by logging macros or by `println!`.  This is because we cannot control the log level on the `kubelet` crate, and we do not want to use `println!` in the crate because it could in principle be hosted in a non-console environment.  Therefore instead the kubelet _application_ passes in a callback representing how it would like to notify its users - in our cases via `println!` again because the justfile logging settings didn't show info messages.

### DIFF
--- a/crates/kubelet/src/bootstrapping/mod.rs
+++ b/crates/kubelet/src/bootstrapping/mod.rs
@@ -23,10 +23,11 @@ const APPROVED_TYPE: &str = "Approved";
 pub async fn bootstrap<K: AsRef<Path>>(
     config: &KubeletConfig,
     bootstrap_file: K,
+    notify: impl Fn(String),
 ) -> anyhow::Result<Config> {
     debug!("Starting bootstrap for {}", config.node_name);
     let kubeconfig = bootstrap_auth(config, bootstrap_file).await?;
-    bootstrap_tls(config, kubeconfig.clone()).await?;
+    bootstrap_tls(config, kubeconfig.clone(), notify).await?;
     Ok(kubeconfig)
 }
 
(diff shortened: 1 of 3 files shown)


itowlson:
The UI looks like this:

image

NOTE: To be clear: the prompt does not introduce any new wait into the kubelets; the bootstrap module does not wait for a response to the message.  It is **purely** informational.

thomastaylor312:
> In preparation for #318, it provides a point in the output where a test harness can know that the CSR is ready, and can extract the CSR name in order to approve it.

Non-blocking question on this? If we are expecting a computer to parse this, should we have it in a computer parsable format rather than a human one?

crates/kubelet/src/bootstrapping/mod.rs
```diff
@@ -23,10 +23,11 @@ const APPROVED_TYPE: &str = "Approved";
 pub async fn bootstrap<K: AsRef<Path>>(
     config: &KubeletConfig,
     bootstrap_file: K,
+    notify: impl Fn(String),
 ) -> anyhow::Result<Config> {
     debug!("Starting bootstrap for {}", config.node_name);
     let kubeconfig = bootstrap_auth(config, bootstrap_file).await?;
```
thomastaylor312, line 29: Nit: Maybe we should have the same notify feature for auth in case there are distros that don't auto approve. That can probably be a low priority issue we create

itowlson:
@thomastaylor312 I wondered about having the format more machine-readable. But the primary use is to provide a human with an explanation of what the program is waiting for.  I did wonder about splitting it across two lines e.g.

```
BOOTSTRAP: blah blah blah, run next line
BOOTSTRAP: the actual command  # easy to parse
```

but I didn't think that would particularly help with the machine processing case.  Can we think of a format that would work well for both?  If not I'll stick to what I got and RELEASE THE REGEXES.

thomastaylor312:
The closest I can think of is JSON, which is only kinda human readable. So RELEASE THE ~KRAKEN~ REGEXES

crates/kubelet/src/bootstrapping/mod.rs
itowlson: Deferring this after discussion with Taylor.

PR #337, opened 2020-08-04
fix(oci-distribution): check for auth token before fetching manifest digest

### LINKED ISSUE #311: Demos failing due to OCI authentication issues.
When running `greet`, my Pod fails with this message: 

```
OCI API error: authentication required, visit https://aka.ms/acr/authorization for more information. on https://webassembly.azurecr.io/v2/greet-wascc/manifests/v0.5
```

bacongobbler:
Calls to fetch_manifest_digest would not fetch the auth token before fetching the manifest digest, causing the call to fail due to an authentication failure.

Several functions were noticed as marked public. Those should not be made available for external use.

Big thanks to @radu-matei for helping me with identifying the issue.

Closes #311

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -104,7 +104,7 @@ impl Client {
     /// For this implementation, it will return v2 or an error result. If the error is a
     /// `reqwest` error, the request itself failed. All other error messages mean that
     /// v2 is not supported.
-    pub async fn version(&self, host: &str) -> anyhow::Result<String> {
+    async fn version(&self, host: &str) -> anyhow::Result<String> {
         let url = format!("{}://{}/v2/", self.config.protocol.as_str(), host);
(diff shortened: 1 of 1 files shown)


bacongobbler:
Note that this is a breaking change to the oci-distribution API, as a client now has to be made mutable before calling `fetch_manifest_digest`, so this should be called out in the 0.4.0 milestone release notes.

radu-matei:
```
---- client::test::test_fetch_digest stdout ----
thread 'client::test::test_fetch_digest' panicked at 'pull manifest should fail: "sha256:51d9b231d5129e3ffc267c9d455c49d789bf3167b611a07ab6e4b3304c96b0e7"', crates/oci-distribution/src/client.rs:455:9
```

Because we now perform the auth workflow for `fetch_manifest_digest`, the test should no longer expect a failure.

PR #340, opened 2020-08-05
downgrade to azure-cli 2.9.1

bacongobbler:
A recent bug was introduced to the azure CLI preventing `az storage blob upload-batch` from working. In the meantime, a hotfix has been provided to the HEAD of azure-blob-storage-upload reverting back to azure-cli v2.9.1

The `sync` flag was also introduced since v1.0.0, which is enabled by default. sync must be disabled to preserve files currently present in the storage account.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,11 +86,12 @@ jobs:
           cd krustlet
           sha256sum * > checksums-${{ env.RELEASE_VERSION }}.txt
       - name: upload to azure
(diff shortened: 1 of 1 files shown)


bacongobbler:
For reference:

- `sync` being enabled by default: https://github.com/bacongobbler/azure-blob-storage-upload/issues/6
- bug in azure-cli 2.10: https://github.com/bacongobbler/azure-blob-storage-upload/issues/7

PR #342, opened 2020-08-05
fix(release): sleep for 60 seconds before publishing kubelet

### LINKED ISSUE #341: CI: publishing a new version of oci-distribution and kubelet fails
There is a bit of a timing issue where kubelet relies on a new version of oci-distribution to be made available before it can be published. `cargo publish` for oci-distribution does run before publishing kubelet, but there is a timing issue before it is discoverable by cargo.

bacongobbler:
often, a new publish of the kubelet crate may need to wait until the new version of oci-distribution is available.

closes #341

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,6 +104,8 @@ jobs:
         working-directory: ./crates/oci-distribution
         run: cargo publish --token ${{ secrets.CargoToken }}
         continue-on-error: true
+      - name: wait for oci-distribution to be published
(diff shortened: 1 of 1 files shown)


radu-matei:
Not blocking this PR, but it would be interesting to have a look at how the crates in Wasmtime are published - https://github.com/bytecodealliance/wasmtime/blob/4cb36afd7b20c9cf0c4737f07425983f057b2eec/scripts/publish.rs#L1

PR #343, opened 2020-08-10
Self-contained integration tester

itowlson:
Addresses #318.

This PR adds a new binary which, when run, does the following:

* Builds the WASI and WASCC kubelets
* Checks if bootstrapped certificates exist; if not, cleans up what does exist and runs the bootstrap script
* Launches both kubelets
* If bootstrapping, monitors both kubelets for the certificate approval message, and approves the certificate
* Runs the integration tests and waits for them to complete
* If the tests fail, captures the kubectl stdout and stderr to files, and prints the test failure details
* Terminates the running kubelets

There is still a lot that could be improved about this, e.g. stream test results as they arrive, send the logs to a better place than the current directory, etc.  But I think it's a useful first step.

## Why write it in Rust, not bash?

Because we can now build and test on Windows, and I don't want to have to maintain parallel bash and Powershell scripts.  Rust is fiddlier than is ideal for this sort of glue code, but using Rust instead of, say, Python means we don't incur another dependency.

Which reminds me...

* ~TODO: invoke `bootstrap.ps1` instead of `bootstrap.sh` when running on Windows~ Wait, I already did that ~and forgot about it~ - excellent work, Ivan, high five

## How do I use it?

Run `just test-e2e-standalone`, or `cargo run --bin oneclick`.

## Can we change the names, reorganise the code, etc.?

Yes.

## Any other to-dos?

I'm glad you asked.

* TODO: incorporate it in the CI
* ~TODO: add it to the documentation~

### DIFF
--- a/justfile
+++ b/justfile
@@ -16,6 +16,9 @@ test:
 test-e2e:
     cargo test --test integration_tests
 
+test-e2e-standalone:
+    cargo run --bin oneclick
+
 run-wascc +FLAGS='': bootstrap
     KUBECONFIG=$(eval echo $CONFIG_DIR)/kubeconfig-wascc cargo run --bin krustlet-wascc {{FLAGS}} -- --node-name krustlet-wascc --port 3000 --bootstrap-file $(eval echo $CONFIG_DIR)/bootstrap.conf --cert-file $(eval echo $CONFIG_DIR)/krustlet-wascc.crt --private-key-file $(eval echo $CONFIG_DIR)/krustlet-wascc.key
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1635,6 +1635,7 @@ dependencies = [
  "kube",
  "kubelet",
  "oci-distribution",
+ "regex",
  "reqwest",
  "serde",
  "serde_derive",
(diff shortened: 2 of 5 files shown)


justfile
```diff
@@ -16,6 +16,9 @@ test:
 test-e2e:
```
bacongobbler, line 16: is there still a need for `just test-e2e` with `test-e2e-standalone` present, or do they perform separate tasks?

I'm wondering if we can replace the github workflow using `just test-e2e` with `just test-e2e-standalone`.

itowlson: That's a good question @bacongobbler. I would keep `test-e2e` around for two reasons:

1. For now, the standalone tester is, er, untested. I would be gobsmacked if we didn't find issues, annoyances or limitations with it.  So, at least for a shakeout period, we need a reliable manual process to fall back on.

2. The standalone tester does not give the live feedback that the manual process does.  You don't see the kubelet logs whizzing by in their terminals; you don't see the tests passing and failing as they go.  This may be particularly relevant to "printf debugging" scenarios where the developer is adding ad hoc logging to the kubelet to figure out why something isn't working.

There is definitely a case for having `test-e2e` run the standalone tester, especially once we have ironed out the initial problems.  But I'd still want to keep, say, a `test-e2e-tests-only` target for those lower-level moments.

PR #344, opened 2020-08-10
[WIP] State Machine

kflansburg:


crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Provider supplies JSON status patch to apply when entering this state.
+    async fn json_status(
+        &self,
+        provider: Arc<Provider>,
+        pod: &Pod,
+    ) -> anyhow::Result<serde_json::Value>;
+}
+
+#[async_recursion::async_recursion]
+/// Recursively evaluate state machine until a state returns Complete.
```
bacongobbler, line 48: Have you considered refactoring this to iterate through the states until one returns "complete" instead? That would greatly reduce the stack size, and it would allow the removal of the `async-recursion` crate.

```
while let Transition::Advance(t) = transition {
}
Transition::Complete(transition)
```

crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,107 @@
...
+    info!(
+        "Pod {} state execution result: {:?}",
+        pod.name(),
+        transition
+    );
+
+    // Handle transition
+    match transition {
+        Transition::Advance(s) => run_to_completion(client, s, provider, pod).await,
+        Transition::Error(s) => run_to_completion(client, s, provider, pod).await,
```
bacongobbler, line 78: If errors and advance states are handled in the same fashion, why do we have two separate types? Shouldn't they be both considered Advance types?

crates/kubelet/src/kubelet.rs
```diff
@@ -295,16 +240,10 @@ mod test {
...
-        async fn add(&self, _pod: Pod) -> anyhow::Result<()> {
-            Ok(())
-        }
-        async fn modify(&self, _pod: Pod) -> anyhow::Result<()> {
-            Ok(())
-        }
-        async fn delete(&self, _pod: Pod) -> anyhow::Result<()> {
-            Ok(())
-        }
+        async fn modify(&self, _pod: Pod) {}
```
bacongobbler, line 245: How do you add a new pod to the Provider if no `add` function is available?

crates/kubelet/src/state.rs
kflansburg: @thomastaylor312 and I spend several days before settling on this solution. We could not get an iterative approach to compile.

crates/kubelet/src/state.rs
kflansburg: This is simply for legibility, before creating this enum I was using Either

crates/kubelet/src/kubelet.rs
kflansburg: A new pod is created by running the state machine to completion. modify and delete remain useful.

kflansburg:
@bacongobbler , @thomastaylor312 and I expected this to be a pretty big lift to integrate. Perhaps we should schedule a meeting to review some of the progress. There is still a lot left to do, and I was hoping to have some help. 

I will say that the only temporary change I have in there right now is disabling compilation of wasi provider. If there are other parts that look clunky or changes that you don't agree with, that is WIP stuff that we should probably discuss.

crates/kubelet/src/kubelet.rs
thomastaylor312: Yeah, this is something we talked about quite a bit. Basically each pod's machine is held in memory, running the pod until it has reached the end of its lifetime. Modify handles the cases where you get a deletion timestamp and the like

crates/wascc-provider/src/lib.rs
```diff
@@ -288,80 +292,75 @@ impl WasccProvider {
...
+    volumes: HashMap<String, Ref>,
+    event_tx: tokio::sync::mpsc::Sender<WatchEvent<KubePod>>,
+    event_rx: tokio::sync::mpsc::Receiver<WatchEvent<KubePod>>,
 }
 
-#[async_trait]
-impl Provider for WasccProvider {
-    const ARCH: &'static str = TARGET_WASM32_WASCC;
+#[async_trait::async_trait]
+impl DefaultStateProvider for WasccProvider {
```
thomastaylor312, line 307: After reading through it here, I am not sure I really like this trait. Might be better to have things defined by the macros instead. Let's talk some more when we meet

crates/kubelet/src/state/default.rs
```diff
@@ -0,0 +1,407 @@
...
+    async fn error(&self, _pod: &Pod) -> anyhow::Result<()> {
+        tokio::time::delay_for(std::time::Duration::from_secs(30)).await;
+        Ok(())
+    }
+}
+
+//
+// * Would be nice to support passing types to the next state (error messages, etc.).
+// * We probably need to explore a more concise way for describing status patches.
+// * Can we offer a macro that doesnt require a trait?
```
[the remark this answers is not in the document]
thomastaylor312: The more I keep thinking about this, the more I am thinking it would be cleaner. I am not sure if we can enforce macro parameter names, but if we can we could enforce it so people using the shortcut method would do something like:

```rust
graph!(
    ImagePull => logic_stuff,
    Volumes => logic_stuff,
    ...
)
```

So if someone put something like `MyCustomThing =>` we could make it fail. People with more advanced needs can then glue their own together using `state!`

crates/wascc-provider/src/states/image_pull.rs
```diff
@@ -0,0 +1,30 @@
...
+        pod_state.run_context.modules = match pod_state.store.fetch_pod_modules(&pod).await {
+            Ok(modules) => modules,
+            Err(e) => {
+                error!("{:?}", e);
+                return Ok(Transition::Error(ImagePullBackoff));
+            }
+        };
+        Ok(Transition::Advance(VolumeMount))
+    },
+    { make_status(Phase::Pending, "ImagePull") }
```
thomastaylor312, line 29: Note for other reviewers: This doesn't have any container status updating yet, so any ideas will be helpful. Most likely we can do something using the current pod state. 

However, we can possibly keep things using the current methodology. Only a few of the steps need to update container statuses (namely the init containers and starting steps), so perhaps it is ok to ask for a full status implementation

crates/wascc-provider/src/states/starting.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        let pod_handle = Handle::new(container_handles, pod.clone(), None).await?;
+        pod_state.handle = Some(pod_handle);
+
+        info!("All containers started for pod {:?}.", pod.name());
+
+        Ok(Transition::Advance(Running))
+    },
+    { make_status(Phase::Pending, "Starting") }
```
thomastaylor312, line 177: Should this be `Phase::Running`? Pending is generally only used during scheduling and image pull

crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,112 @@
...
+    info!(
+        "Pod {} state execution result: {:?}",
+        pod.name(),
+        transition
+    );
+
+    // Handle transition
+    match transition {
+        Transition::Advance(s) => run_to_completion(client, s, pod_state, pod, state_rx).await,
+        Transition::Error(s) => run_to_completion(client, s, pod_state, pod, state_rx).await,
```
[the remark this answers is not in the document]
thomastaylor312: Spitballing here: What if any backoff state actually triggers an early exit which triggers a "reschedule" of the pod on the queue? Starting over would just be a reconciling process. The images would already exist and volumes would already be provisioned, so it should replay over those steps easily

crates/wascc-provider/src/lib.rs
```diff
@@ -133,11 +134,12 @@ impl StopHandler for ActorHandle {
 /// from Kubernetes.
 #[derive(Clone)]
 pub struct WasccProvider {
+    client: kube::Client,
     handles: Arc<RwLock<HashMap<String, Handle<ActorHandle, LogHandleFactory>>>>,
+    run_contexts: Arc<RwLock<HashMap<String, ModuleRunContext>>>,
```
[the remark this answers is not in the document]
kflansburg: You are right, I think several of these fields can be removed now. I was hoping that rust would detect that they are unused but I guess that is asking too much. I will make a pass to see what else is now superfluous in the root module.

crates/kubelet/src/state.rs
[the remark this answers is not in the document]
kflansburg: I think this makes sense, the states are self contained, so there is no functional difference between a CrashLoopBackoff state that is at the first level of function call and one that is deep in the nested calls. Now that I think about it, I'm wondering if Rust unwraps this tail call? Probably best not to rely on that though.

crates/wascc-provider/src/states/crash_loop_backoff.rs
```diff
@@ -0,0 +1,23 @@
...
+use super::registered::Registered;
+
+state!(
+    /// Pod has failed multiple times.
+    CrashLoopBackoff,
+    PodState,
+    Registered,
+    CrashLoopBackoff,
+    {
+        // TODO: Handle pod delete?
```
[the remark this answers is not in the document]
kflansburg: I do think exponential backoff is something kubelet should provide. Probably not in this pull request though. The note about delete was that if you enter this state and just block for 60 or more seconds, the user should be able to delete the pod (as I frequently want to do if it is in crash-loop-backoff), but I think we discussed a better way to handle this that doesn't rely on the user watching a channel, so I will give that a shot.

crates/wascc-provider/src/states/error.rs
```diff
@@ -0,0 +1,44 @@
+use kubelet::pod::{Phase, Pod};
+use kubelet::state::{PodChangeRx, State, Transition};
+
+use super::crash_loop_backoff::CrashLoopBackoff;
```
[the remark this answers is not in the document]
kflansburg: Yes, similarly I think we should spend some time thinking about a prelude type export for the state machine types, you can see there are a lot of boilerplate imports in each of these state files.

crates/wascc-provider/src/states/starting.rs
kflansburg: I'm not sure what Kubernetes does. I figure it will mark it as running as soon as all the containers start and it transitions to running, which was what I expected. Note that the status is currently applied when it enters the state, not when it exits.

thomastaylor312: > Note that the status is currently applied when it enters the state, not when it exits

I missed this when I went through the code

itowlson:
I love the way this makes the various stages of the pod lifecycle so much more explicit, more distinct and easier to understand.  However, there are some details of the design which I'm finding a bit tricky to get my head around...

1. The recursive `run_to_completion` function.  @bacongobbler already mentioned this and you explained the reasons for the current approach, but I'm not sure we're making the right trade off here.  The trade off appears to be compile time validation of state transitions, which require a recursive `run_to_completion` because each call must be parameterised by the current state type, vs. a looser approach (e.g. using an enum) which would allow an iterative (or, in theory, tail recursive) implementation of the machine.  I assume the argument is that the pod will never make so many transitions that stack depth could be a problem, but @thomastaylor312's comment about a potential infinite loop in CrashLoopBackoff makes me worry about that; and, okay, this is subjective and aesthetic, but conceptually the big stack unwind when you finally get `Transition::Complete` feels awkward to me (though it will have few or no _practical_ implications).  I do quite like the sense, though, that `run_to_completion` goal-seeks from wherever it finds itself, and a tail recursive implementation would be a lovely way to express that.

2. Perhaps I am missing something but a `Pod` doesn't seem to have a state member; rather, the state seems to be an independent thing which travels around with the Pod (and, I think, also with the `PodState` though I'm not quite clear about that).  I can see reasons for this design in order to avoid parameterising the `kubelet::Pod` type by the type of its state machine, but it was surprising to me.  This feels like something it would be good to investigate.

3. The transition design seems to bake in a linear flow - that on the happy path a pod will move through a fixed sequence of states, and that even on an unhappy path there is only one possible error state at a given time.  This may be true of the current WASCC pod breakdown but seems a dangerous assumption to build in.  For example, imagine an environment where the runtime could temporarily suspend a pod - then the Running state might have success transitions to both Paused and Finished.  Or you could envisage error transitions to TimedOut and ExitedWithError.  I realise this is speculative and for now the linear flow is fine, but I'm anxious that we may be building in an assumption that will be very hard to change.

Hope this is useful feedback, and many apologies for raising concerns without yet having any positive suggestions for solutions/improvements.  I'm having a tinker and explore but it's going slowly - will definitely chime in if I come up with anything that I think could be useful!

kflansburg:
Hey @itowlson, thanks for the feed back. I agree on most of your concerns:

1. I think you have done a good job of summarizing the tradeoffs here. I also think filling up the stack is a very real concern, for instance, in my own testing, the `kube-proxy` pods that get scheduled result in an image_pull / image_pull_backoff loop that will eventually manifest this problem. I'd really like to find a way to move to a non-recursive approach, I'm optimistic that this is something we can iterate on though and doesn't require us to totally rework this solution. RE: tail-recursion, @thomastaylor312 and I looked into this and it _appears_ that Rust does no tail-recursion unwrapping whatsoever, and so this would not be a solution. 

2. This is an interesting point. I've thought about it a bit as I've worked with Krustlet. To me, `Pod` is actually just a convenience wrapper around `KubePod` manifest, and I think some sort of execution context for a Pod that you describe would be a different object. There may be a way to work this into refactoring what I have that would simplify some things. For now I have simply had `Pod::queue` and `run_to_completion` maintain ownership of the Pod execution state since they represent the "Kubelet control loop" and have been a big focus of this project. 

3. This is definitely a valid point. I wish there were an obvious way to provide more flexibility in terms of outgoing edges. In particular I've found the case where a container image gets modified and Running has to transition back to ImagePull which would be an additional outgoing edge. 

I plan to take one more pass this week, including reviewing the full code changes in full and trying to polish off anything I find. I will also see if I can address these issues in particular. If anyone has any solutions, they would be greatly appreciated.

itowlson:
@kflansburg Re point 1, I wonder how much the static checking is gaining us relative to what it's costing us.  An enum-based approach would support an iterative approach, but you wouldn't be able to enforce that only specific transitions happened, but I'm not sure this is a huge loss given the relatively simple graphs we expect.  How much do you feel we are benefiting from the compile-time checks?  (Also, great example with the `kube-proxy` image pull - that definitely sounds like something we would have to resolve.)

itowlson:
FAQ confirms that Rust almost never does the tail call optimisation, and even when it does we shouldn't rely on it doing so.

kflansburg:
@itowlson I think the compile-time checks are pretty useful. There is utility in just developer convenience, I think having that double check can simplify the cognitive load when working with the state machine graph. The biggest thing though is that in my experience using the Enum wrapper would result in a lot of places where you would have to match on the enum and throw errors at runtime for many invalid states, I think this is pretty undesirable. 

I'd like to give this a few more days. I wonder if hoverbear (author of original blog post) would be willing to review this work.

itowlson:
@kflansburg Re point 2, that distinction makes sense.  I was thinking about something like this for containers (but put it aside pending the state refactor).  Something like a `RuntimePod` that contained the Kubernetes pod spec (as represented by the current `Pod` or `KubePod` type), the Krustlet/provider resources involved in running that pod, and Krustlet's view of the status of the pod, would seem to be quite a meaningful abstraction.  Agree it may be too much to attempt to fold into this PR, but it feels like a probably endgame for the state refactor.

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
-    where
-        P: 'static + Provider + Sync + Send,
-    {
-        let (sender, mut receiver) = watch::channel(initial_event);
-        let worker = tokio::spawn(async move {
+    ) -> anyhow::Result<tokio::sync::mpsc::Sender<WatchEvent<KubePod>>> {
+        let (sender, mut receiver) = tokio::sync::mpsc::channel::<WatchEvent<KubePod>>(16);
+
+        let pod_deleted = Arc::new(RwLock::new(false));
+        let check_pod_deleted = Arc::clone(&pod_deleted);
```
thomastaylor312, line 53: Nit: Can't you just call `pod_deleted.clone()`?

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
-    ) -> Self
-    where
-        P: 'static + Provider + Sync + Send,
-    {
-        let (sender, mut receiver) = watch::channel(initial_event);
-        let worker = tokio::spawn(async move {
+    ) -> anyhow::Result<tokio::sync::mpsc::Sender<WatchEvent<KubePod>>> {
+        let (sender, mut receiver) = tokio::sync::mpsc::channel::<WatchEvent<KubePod>>(16);
+
+        let pod_deleted = Arc::new(RwLock::new(false));
```
thomastaylor312, line 52: Any reason not to use an `AtomicBool` here?

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
-        provider: Arc<P>,
-        mut error_sender: Sender<(KubePod, anyhow::Error)>,
-    ) -> Self
-    where
-        P: 'static + Provider + Sync + Send,
-    {
-        let (sender, mut receiver) = watch::channel(initial_event);
-        let worker = tokio::spawn(async move {
+    ) -> anyhow::Result<tokio::sync::mpsc::Sender<WatchEvent<KubePod>>> {
+        let (sender, mut receiver) = tokio::sync::mpsc::channel::<WatchEvent<KubePod>>(16);
```
thomastaylor312, line 50: Do we need to handle multiple events or just the latest event since we'll reconcile it to the latest definition anyway?

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
+                                let api: kube::Api<KubePod> = kube::Api::namespaced(task_client.clone(), pod.namespace());
+                                let patch = serde_json::json!(
+                                    {
+                                        "metadata": {
+                                            "resourceVersion": "",
+                                        },
+                                        "status": {
+                                            "phase": Phase::Failed,
+                                            "reason": format!("{:?}", e),
+                                            "containerStatuses": Vec::<()>::new(),
```
thomastaylor312, line 88: These are still a TODO, right? Just making sure we don't lose track

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
+
+                    info!("Pod {} waiting for deregistration.", name);
+                    loop {
+                        tokio::time::delay_for(std::time::Duration::from_secs(1)).await;
+                        if *check_pod_deleted.read().await {
+                            info!("Pod {} deleted.", name);
+                            break;
+                        }
+                    }
+                    pod_state.async_drop().await;
```
thomastaylor312, line 116: Nit: Any way we could combine these into a single `drop`?

crates/kubelet/src/pod/queue.rs
```diff
@@ -19,79 +30,166 @@ use crate::provider::Provider;
...
+                        info!("Pod {} modified.", Pod::new(pod.clone()).name());
+                        // Not really using this right now but will be useful for detecting changes.
+                        let pod = Pod::new(pod);
+                        // TODO, detect other changes we want to support, or should this just forward the new pod def to state machine?
+                        if let Some(_timestamp) = pod.deletion_timestamp() {
+                            *(pod_deleted.write().await) = true;
+                        }
+                    }
+                    WatchEvent::Deleted(pod) => {
+                        // I'm not sure if this matters, we get notified of pod deletion with a
```
thomastaylor312, line 153: I think the proper thing here would be to do `WatchEvent::Modified(pod) | WatchEvent::Deleted(pod)` for the block above if we want to run our deregister logic anyway. Right now, if someone force deletes the pod, it wouldn't run the cleanup on our end. The only thing is that error handling would have to manage if the pod is already deleted and ignore it instead of logging the error

crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Advance to next node.
+    Advance(S),
+    /// Transition to error node.
+    Error(E),
+    /// This is a terminal node of the state graph.
+    Complete(anyhow::Result<()>),
+}
+
+#[async_trait::async_trait]
+/// Allow for asyncronous cleanup up of PodState.
```
thomastaylor312, line 24: Spelling nit: "asynchronous"

crates/wascc-provider/src/states/finished.rs
```diff
@@ -0,0 +1,20 @@
...
+use crate::{make_status, PodState};
+
+state!(
+    /// Pod execution completed with no errors.
+    Finished,
+    PodState,
+    Finished,
+    Finished,
+    {
+        // TODO: Wait for deleted.
```
thomastaylor312, line 16: Should "Finished" just be a pseudo-state always included by the krustlet crate itself?

crates/wascc-provider/src/states/running.rs
```diff
@@ -0,0 +1,26 @@
...
+
+state!(
+    /// The Kubelet is running the Pod.
+    Running,
+    PodState,
+    Finished,
+    Error,
+    {
+        // Wascc has no notion of exiting so we just sleep.
+        // I _think_ that periodically awaiting will allow the task to be interrupted.
```
thomastaylor312, line 20: I think you are right here. Anything that is awaiting can be parked. But I don't know all of the implementation details under the hood

crates/kubelet/src/pod/queue.rs
kflansburg: I don't think that's best practice, I think clippy yells at you or something.

crates/kubelet/src/pod/queue.rs
kflansburg: That would work too, seems more heavy than the RwLock though.

crates/kubelet/src/pod/queue.rs
kflansburg: I agree that you only need to handle the latest event, but I also like that you don't have to initialize the channel with the first event, since we are using it immediately.

crates/kubelet/src/pod/queue.rs
kflansburg: Yeah, I figured this was something we can iterate on.

crates/kubelet/src/pod/queue.rs
kflansburg: Maybe, I'm not convinced that AsyncDrop is really what we want to call this. It would be good to make sure that it is always dropped when async_drop is called.

crates/kubelet/src/pod/queue.rs
kflansburg: Ah, good point.

crates/wascc-provider/src/states/finished.rs
kflansburg: Hmm, maybe so it would transition to that to exit? Would we also provide failed or terminated or something like that?

crates/wascc-provider/src/states/running.rs
kflansburg: I'm hoping that as long as we tell our providers to follow the rule of thumb of never running blocking code in their async handlers, that we don't have to worry too much about zombie workloads.

crates/wascc-provider/src/states/finished.rs
thomastaylor312: Either one would work I think. My personal preference would be to have a "return `Completed` to exit". Either way helps avoid provider implementations having to do the final k8s waiting/cleanup

crates/kubelet/src/pod/queue.rs
thomastaylor312: Hmmm...I definitely thought for Arcs you just call clone. Let me know what clippy thinks!

crates/kubelet/src/pod/queue.rs
thomastaylor312: I honestly don't know which is heavier. Though conceptually in my head it seems like an atomic bool makes more sense

crates/wascc-provider/src/states/finished.rs
kflansburg: Ah, I see. so this is actually handled in the Pod::queue, I need to remove this comment.

crates/kubelet/src/pod/queue.rs
thomastaylor312: Should we at least lower the buffer size then?

crates/kubelet/src/pod/queue.rs
kflansburg: Well clippy didn't complain but I think it used to https://github.com/rust-lang/rust/pull/63252. It may have been rylev that first told me to do it this way.

crates/kubelet/src/pod/queue.rs
kflansburg: The think I like about `RwLock` is that it is async. We don't need any sort of ordering so I tend to think of atomics as unnecessary.

crates/kubelet/src/pod/queue.rs
kflansburg: We can, I felt like this was a reasonably small default.

crates/wascc-provider/src/states/starting.rs
```diff
@@ -0,0 +1,174 @@
...
+    })
+    .await?
+}
+
+state!(
+    /// The Kubelet is starting the Pod.
+    Starting,
+    PodState,
+    {
+        info!("Starting containers for pod {:?}", pod.name());
```
itowlson, line 140: I've got a concern around discoverability with blocks like this.  For me at least, I'm getting no code completion in VS Code, and there's nothing that tells me what parameters I have available to me.  Is it possible to make these, say, a lambda, or a function name?

kflansburg: There is a second macro option that accepts a function (although now that I look at it, it needs to be updated to accept the `PodState`) I'm not sure if this will fix the VS Code issue. But in general I agree, this requires developers to understand a lot of magic behind the macro. I'm wondering if the macro is really that useful, the trait itself is not especially verbose. We might consider reserving the macro for the `graph` default state graph implementation that @thomastaylor312 has proposed.

thomastaylor312: Yeah, we do have a macro that just takes a function path and then calls it so that should solve the auto completion thing. We could just expect people to implement the `State` trait by hand now that it is a bit simpler

crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,106 @@
...
+/// A trait representing a node in the state graph.
+pub trait State<PodState>: Sync + Send + 'static + std::fmt::Debug {
+    /// Provider supplies method to be executed when in this state.
+    async fn next(
+        &self,
+        pod_state: &mut PodState,
+        pod: &Pod,
+    ) -> anyhow::Result<Transition<Box<dyn State<PodState>>, Box<dyn State<PodState>>>>;
+
+    /// Provider supplies JSON status patch to apply when entering this state.
```
thomastaylor312, line 40: Would it make more sense to have the status patch come _after_ the work is performed?

kflansburg: In my mental model, it makes sense to patch when entering states. If you patch when exiting states, you need to know which state you are going to right? I think it makes sense for each state to own its own patch.

crates/kubelet/src/state.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    loop {
+        info!("Pod {} entering state {:?}", pod.name(), state);
+
+        let patch = state.json_status(pod_state, &pod).await?;
+        info!("Pod {} status patch: {:?}", pod.name(), &patch);
+        let data = serde_json::to_vec(&patch)?;
+        api.patch_status(&pod.name(), &PatchParams::default(), data)
+            .await?;
+        info!("Pod {} executing state handler {:?}", pod.name(), state);
```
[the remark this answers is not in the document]
kflansburg: Will review logging verbosity now that I'm out of the debugging phase. .

crates/wascc-provider/src/states/starting.rs
kflansburg: I'm going to ignore this for now since the use of the macro or the trait is entirely up to the user.

crates/kubelet/src/state/macros.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    };
+    (
+       $(#[$meta:meta])*
+       $name:ident,
+       $state:path,
+       $success:ty,
+       $error: ty,
+       $work:path,
+       $patch:block
```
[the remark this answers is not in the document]
kflansburg: @thomastaylor312  I think I just copied your code for this, If this is wrong let me know.

kflansburg:
@bacongobbler @thomastaylor312 @itowlson 
I have rebased this and reviewed the code one last time. The e2e tests will not pass because of the wasi provider being disabled. 

The OSX build seems to be failing with:
```
error: reached the type-length limit while instantiating `std::pin::Pin::<&mut std::future...::Future>::poll::{{closure}}#0]>`
738
    |
739
    = note: consider adding a `#![type_length_limit="1256844"]` attribute to your crate
```
 
However, it builds on my Mac. 

I see the following as outstanding issues that we may wish to look into after merging this:
* Remove remaining errors and unwraps in `wascc` states, ideally these handlers never fail and instead transition to Error states. 
* Add class methods to `Transition` to remove all the `Box`es the user has to create right now. 
* Possibly support more `Transition` options for more outgoing edges.
* I don't think we are actually type checking the graph as much as we were before with the recursive implementation because all it knows is that the state transitions are `dyn State`. I'd like to think about this more. 
* Refactor `Pod::queue` to break up that big method for running the Pod. 
* And of course, the `wasi` provider.

bacongobbler:
Thanks @kflansburg. I'll open issues for those action items you mentioned. The macOS error seems odd... I'll re-run the e2e tests just to make sure it wasn't some one-off error.

I think we're far enough along here that we're ready to merge this in and iterate upon its design. Thank you so much for your work on this!

bacongobbler:
huh. Now it's failing for both ubuntu and macos. I'll file another ticket to look into this.

PR #348, opened 2020-08-11
OCI client version function was showing as dead code

itowlson:
We have dead code warnings for the `oci_distribution::Client::version` function.  It's not clear whether it was intended to be public, or simply fell out of use.  As there are tests for it, I'm guessing the former.  Reviewers, if you think it's truly dead code, comment and I'll do it that way instead *grin*.


bacongobbler:
It is truly dead code. After pairing with @radu-matei he explained that the only two functions that should be public are `pull_image` and `fetch_manifest_digest`. So it can probably be removed.

itowlson:
I deaded it

itowlson:
image

PR #349, opened 2020-08-11
[oci-distribution] Support digest references

jlegrone:
The OCI distribution spec supports [fetching manifests by tag or digest](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#get-manifest). This PR adds support for parsing and downloading digested manifest references.

## ⚠️ Breaking Changes

The public `to_v2_blob_url` and `to_v2_manifest_url` methods of `Reference` have been removed and are now private methods on `Client`. These seemed like implementation details of the client, but if it would be preferable to hash this out in a separate PR I'm happy to do so.

### DIFF
--- a/crates/oci-distribution/Cargo.toml
+++ b/crates/oci-distribution/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "oci-distribution"
-version = "0.3.0"
+version = "0.4.0"
 authors = [
     "Matt Butcher <matt.butcher@microsoft.com>",
     "Matthew Fisher <matt.fisher@microsoft.com>",
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -185,7 +185,7 @@ impl Client {
             self.auth(image, None).await?;
         }
 
-        let url = image.to_v2_manifest_url(self.config.protocol.as_str());
+        let url = self.to_v2_manifest_url(image);
         debug!("Pulling image manifest from {}", url);
         let request = self.client.get(&url);
 
@@ -216,7 +216,7 @@ impl Client {
     /// If the connection has already gone through authentication, this will
     /// use the bearer token. Otherwise, this will attempt an anonymous pull.
     async fn pull_manifest(&self, image: &Reference) -> anyhow::Result<(OciManifest, String)> {
-        let url = image.to_v2_manifest_url(self.config.protocol.as_str());
+        let url = self.to_v2_manifest_url(image);
         debug!("Pulling image manifest from {}", url);
         let request = self.client.get(&url);
 
@@ -266,7 +266,11 @@ impl Client {
         digest: &str,
         mut out: T,
     ) -> anyhow::Result<()> {
-        let url = image.to_v2_blob_url(self.config.protocol.as_str(), digest);
+        let url = self.to_v2_blob_url(
+            image.registry(),
+            image.repository(),
+            digest,
+        );
         let mut stream = self
             .client
             .get(&url)
@@ -280,6 +284,38 @@ impl Client {
         }
 
         Ok(())
+    }
+
+    /// Convert a Reference to a v2 manifest URL.
+    fn to_v2_manifest_url(&self, reference: &Reference) -> String {
+        if let Some(digest) = reference.digest() {
+            format!(
+                "{}://{}/v2/{}/manifests/{}",
+                self.config.protocol.as_str(),
+                reference.registry(),
+                reference.repository(),
+                digest,
+            )
+        } else {
+            format!(
+                "{}://{}/v2/{}/manifests/{}",
+                self.config.protocol.as_str(),
+                reference.registry(),
+                reference.repository(),
+                reference.tag().unwrap_or("latest")
+            )
+        }
+    }
+
+    /// Convert a Reference to a v2 blob (layer) URL.
+    fn to_v2_blob_url(&self, registry: &str, repository: &str, digest: &str) -> String {
+        format!(
+            "{}://{}/v2/{}/blobs/{}",
+            self.config.protocol.as_str(),
+            registry,
+            repository,
+            digest,
+        )
     }
 
     /// Generate the headers necessary for authentication.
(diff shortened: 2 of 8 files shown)


crates/oci-distribution/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "oci-distribution"
-version = "0.3.0"
+version = "0.4.0"
```
bacongobbler, line 3: can you revert this? We generally don't bump the version number until we're ready to cut a release.

crates/oci-distribution/src/client.rs
```diff
@@ -411,7 +494,7 @@ mod test {
 
     #[tokio::test]
     async fn test_auth() {
-        let image = Reference::try_from(HELLO_IMAGE).expect("failed to parse reference");
+        let image = Reference::try_from(HELLO_IMAGE_TAGGED).expect("failed to parse reference");
```
bacongobbler, line 497: Can we add a separate test instead for HELLO_IMAGE_TAGGED? That way we can continue to test image references with no tag.

bacongobbler: In fact... It might be better to add all three into a matrix, such that each `test_*` function tests untagged, tagged, and digest references to make sure all three work in a consistent fashion.

jlegrone:
Thanks for the feedback so far!

> On a side note, we noticed there's a new oras crate available, but the repository is private. Is there an intention here to replace oci-distribution with oras once that's a bit more stable, or do you see the two as handling separate goals?

I'm still experimenting (read: lots of force pushes 😄 ) but will be open sourcing after a little more iteration. My current thinking is that oci-distribution would provide a low level OCI client to the oras crate.

Right now requirements are pretty minimal, but eventually oras will require a full-fledged OCI client implementation. Is that something that seems likely to fit the scope and maintenance budget for oci-distribution?

crates/oci-distribution/src/client.rs
```diff
@@ -397,7 +429,45 @@ mod test {
     use super::*;
     use std::convert::TryFrom;
 
-    const HELLO_IMAGE: &str = "webassembly.azurecr.io/hello-wasm:v1";
+    const HELLO_IMAGE_NO_TAG: &str = "webassembly.azurecr.io/hello-wasm";
```
jlegrone, line 432: This currently breaks e2e tests because `webassembly.azurecr.io/hello-wasm:latest` is not a valid reference. This can be removed, but we'll have better coverage if we add a `latest` tag pointing to the same digest as `v1`.

Sidebar: implementing a mock registry endpoint to avoid going over the network during test execution would probably be a good idea.

crates/oci-distribution/src/client.rs
jlegrone: I took a crack at this in a8352eb. I'm not too familiar with table driven testing patterns in rust so my approach is probably a little naïve. Notably, when any of the test cases trigger a panic during a loop like the one below, later test cases are skipped.

https://github.com/deislabs/krustlet/blob/a8352ebe797eaa21d22778d1c0efb384f4151cd0/crates/oci-distribution/src/client.rs#L577-L587

It seems like we may need to use a macro to generate test functions in order to avoid this problem, but I'm not sure if this is desirable given the additional complexity that would incur. Are there any other approaches that might work? This [SO question](https://stackoverflow.com/questions/34662713/how-can-i-create-parameterized-tests-in-rust) has some interesting discussion but may be outdated.

jlegrone: I just had some fun going down this rabbit hole. 😄 A refactor like this is probably best for a separate PR, but I came up with two macro based solutions, each resulting in identical cargo test output.

1. Define our own macro for each test type using [macro_rules](https://doc.rust-lang.org/rust-by-example/macros.html#macro_rules) (no dependencies)

```rust
macro_rules! test_to_v2_manifest {
    ($($name:ident: $value:expr,)*) => {
        mod to_v2_manifest {
            use super::*;
            $(
                #[test]
                fn $name() {
                    let (input, expected) = $value;
                    let reference = Reference::try_from(input).expect("failed to parse reference");
                    assert_eq!(Client::default().to_v2_manifest_url(&reference), expected);
                }
            )*
        }
    }
}

test_to_v2_manifest! {
    no_tag: (
        HELLO_IMAGE_NO_TAG,
        "https://webassembly.azurecr.io/v2/hello-wasm/manifests/latest",
    ),
    tag: (
        HELLO_IMAGE_TAG,
        "https://webassembly.azurecr.io/v2/hello-wasm/manifests/v1",
    ),
    digest: (
        HELLO_IMAGE_DIGEST,
        "https://webassembly.azurecr.io/v2/hello-wasm/manifests/sha256:51d9b231d5129e3ffc267c9d455c49d789bf3167b611a07ab6e4b3304c96b0e7",
    ),
    tag_and_digest: (
        HELLO_IMAGE_TAG_AND_DIGEST,
        "https://webassembly.azurecr.io/v2/hello-wasm/manifests/sha256:51d9b231d5129e3ffc267c9d455c49d789bf3167b611a07ab6e4b3304c96b0e7",
    ),
}
```

2. Use the [test_case](https://docs.rs/test-case/1.0.0/test_case/) crate (downside: it doesn't currently work with `#[tokio::test]`)

```rust
#[test_case(
    HELLO_IMAGE_NO_TAG,
    "https://webassembly.azurecr.io/v2/hello-wasm/manifests/latest";
    "no tag"
)]
#[test_case(
    HELLO_IMAGE_TAG,
    "https://webassembly.azurecr.io/v2/hello-wasm/manifests/v1";
    "tag"
)]
#[test_case(
    HELLO_IMAGE_DIGEST,
    "https://webassembly.azurecr.io/v2/hello-wasm/manifests/sha256:51d9b231d5129e3ffc267c9d455c49d789bf3167b611a07ab6e4b3304c96b0e7";
    "digest"
)]
#[test_case(
    HELLO_IMAGE_TAG_AND_DIGEST,
    "https://webassembly.azurecr.io/v2/hello-wasm/manifests/sha256:51d9b231d5129e3ffc267c9d455c49d789bf3167b611a07ab6e4b3304c96b0e7";
    "tag and digest"
)]
fn to_v2_manifest(image: &str, manifest_uri: &str) {
    let reference = Reference::try_from(image).expect("failed to parse reference");
    assert_eq!(
        Client::default().to_v2_manifest_url(&reference),
        manifest_uri
    );
}
```

crates/oci-distribution/src/reference.rs
```diff
@@ -3,13 +3,15 @@ use std::convert::{Into, TryFrom};
 /// An OCI image reference
 ///
 /// currently, the library only accepts modules tagged in the following structure:
-/// <registry>/<repository>:<tag> or <registry>/<repository>
+/// <registry>/<repository>, <registry>/<repository>:<tag>, or <registry>/<repository>@<digest>
```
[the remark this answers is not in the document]
jlegrone: It does, I was planning to follow up with a second PR to improve the documentation (this comment actually [doesn't render properly](https://docs.rs/oci-distribution/0.3.0/oci_distribution/struct.Reference.html)), but if we're ok increasing the scope of this PR a bit more I can include those changes too.

crates/oci-distribution/src/reference.rs
```diff
@@ -111,45 +116,50 @@ mod tests {
...
     fn correctly_parses_string() {
+        // Tag only (String)
         let reference = Reference::try_from("webassembly.azurecr.io/hello:v1".to_owned())
             .expect("Could not parse reference");
 
         assert_eq!(reference.registry(), "webassembly.azurecr.io");
         assert_eq!(reference.repository(), "hello");
         assert_eq!(reference.tag(), Some("v1"));
 
+        // Tag only (&str)
```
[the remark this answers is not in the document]
jlegrone: Done in f3ffb70

crates/oci-distribution/src/reference.rs
```diff
@@ -32,62 +34,65 @@ impl Reference {
...
         })?;
-        let colon = string[slash + 1..].find(':');
+        let digest_start = string[repo_start + 1..].find('@').map(|i| repo_start + i + 1);
+        let mut tag_start = string[repo_start + 1..].find(':').map(|i| repo_start + i + 1);
+
+        let repo_end = match (digest_start, tag_start) {
+            (Some(d), Some(t)) => {
+                if t > d {
+                    // tag_start is after digest_start, so no tag is actually present
+                    tag_start = None;
```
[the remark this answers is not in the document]
jlegrone: Done in 6bf96423c1ed382cbb4f7081089c0a3ea2433d9e

crates/oci-distribution/src/client.rs
bacongobbler: done

PR #353, opened 2020-08-17
ref(*): Renames node taint to `kubernetes.io/arch`

### LINKED ISSUE #334: Rationalise krustlet/arch and kubernetes.io/arch
The kubelet node builder automatically sets the node `kubernetes.io/arch` label to the provider's `ARCH` constant.  The in-repo providers also taint the nodes with the key `krustlet/arch`.  This means that pods need to _tolerate_ `krustlet/arch` but _select_ `kubernetes.io/arch`, which feels a bit inconsistent.  We should consider adopting `kubernetes.io/arch` as the taint key for the in-repo providers.

One factor to consider, though, is that this would be a breaking change for existing pod definitions, as they would not have tolerations for the new key.  There may not be any way to mitigate this.  For example, having both taints side by side would simply mean that pods need to tolerate both, which is still breaking.  If we can't find a way to parallel-run for a bit then we will need to communicate very clearly and coordinate carefully with the community around the changeover.

thomastaylor312:
Please note that this is a breaking change that should be called out
in the release notes. This updates all documentation and examples
to use the new taint name.

Closes #334

### DIFF
--- a/crates/wascc-provider/src/lib.rs
+++ b/crates/wascc-provider/src/lib.rs
@@ -336,7 +336,7 @@ impl Provider for WasccProvider {
 
     async fn node(&self, builder: &mut Builder) -> anyhow::Result<()> {
         builder.set_architecture("wasm-wasi");
-        builder.add_taint("NoExecute", "krustlet/arch", Self::ARCH);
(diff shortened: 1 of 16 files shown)


itowlson (approved):
LGTM

PR #354, opened 2020-08-18
Mount individual items from configmaps or secrets

### LINKED ISSUE #230: support configmap volume `items` API
Example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: hello-world-wasi-rust
data:
  myval: "cool stuff"
  loremIpsum: |
    Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut
    labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
    laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
    voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
    non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
---
apiVersion: v1
kind: Pod
metadata:
  name: hello-world-wasi-rust
spec:
  containers:
    - name: hello-world-wasi-rust
      image: webassembly.azurecr.io/hello-world-wasi-rust:v0.2.0
      env:
        - name: FOO
          value: bar
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: CONFIG_MAP_VAL
          valueFrom:
            configMapKeyRef:
              key: myval
              name: hello-world-wasi-rust
      volumeMounts:
        - name: storage
          mountPath: /mnt/storage
  volumes:
    - name: storage
      configMap:
        name: hello-world-wasi-rust
        items:
          - key: loremIpsum
            path: lorem_ipsum.txt
  tolerations:
    - key: "krustlet/arch"
      operator: "Equal"
      value: "wasm32-wasi"
      effect: "NoExecute"

```

Here's the important bit:

```
        items:
          - key: loremIpsum
            path: lorem_ipsum.txt
```

The current behaviour will mount the whole ConfigMap into the WASI runtime. The `items` API tells the runtime that only certain keys should be mounted, and at what path.

The example above should be available at /mnt/storage/lorem_ipsum.txt in the runtime, and `myval` should not be present.

itowlson:
Fixes #230.

### DIFF
--- a/crates/kubelet/src/volume/mod.rs
+++ b/crates/kubelet/src/volume/mod.rs
@@ -5,7 +5,7 @@ use std::ops::Deref;
 use std::path::PathBuf;
 
 use k8s_openapi::api::core::v1::Volume as KubeVolume;
-use k8s_openapi::api::core::v1::{ConfigMap, Secret};
+use k8s_openapi::api::core::v1::{ConfigMap, KeyToPath, Secret};
 use k8s_openapi::ByteString;
 use kube::api::Api;
 use log::{debug, error};
@@ -120,6 +120,7 @@ async fn configure(
             namespace,
             client,
             path,
+            &cm.items,
         )
         .await
     } else if let Some(s) = &vol.secret {
@@ -130,6 +131,7 @@ async fn configure(
             namespace,
             client,
             path,
+            &s.items,
         )
         .await
     } else if let Some(hostpath) = &vol.host_path {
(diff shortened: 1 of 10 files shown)


crates/kubelet/src/volume/mod.rs
```diff
@@ -148,14 +150,20 @@ async fn populate_from_secret(
...
     let data = secret.data.unwrap_or_default();
     let data = data.iter().map(|(key, ByteString(data))| async move {
-        let file_path = path.join(key);
-        tokio::fs::write(file_path, &data).await
+        match mount_path_for(key, items) {
+            Some(mount_path) => {
+                let file_path = path.join(mount_path);
+                tokio::fs::write(file_path, &data).await
+            }
+            None => Ok(()),
```
thomastaylor312, line 165: Maybe we should log an warning if there is no path found?

itowlson: I don't think so; the `None` case occurs when the pod spec includes an `items` clause that says not to mount this particular data item.  We could debug-log _all_ the items and whether they are mounted and if so at what path; but "user chose not to mount this item" is definitely not warning material.

However, the fact that the code _looks_ like it might merit a warning suggests that maybe I need to pick more explicit names to make it clear that this is a positive choice rather than a "it's skipping something or dropping something" situation...?

thomastaylor312: Yeah, It might be clearer with some other names as it definitely seems like it is just dropping something. However, I am not going to block

PR #359, opened 2020-08-19
Insecure registries

### LINKED ISSUE #356: handling image from insecure registry
I set up a local registry with [kind](https://kind.sigs.k8s.io/docs/user/local-registry/). I pushed web assembly image via wasm-to-oci, but wasn't able to deploy the workload via kubectl. kubectl describe pod POD_NAME points to error message that krustlet is using https instead of http. 

`Message:      error sending request for url (https://localhost:5000/v2/): error trying to connect: error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../ssl/record/ssl3_record.c:332:`

itowlson:
Fixes #356.

I'm not quite sure how to test this.  I've tested the config and the URL building, but can't see how to verify that it works end to end.  @amar123deep, would you be able to pull this PR and give it a test in your Kind environment please?


crates/kubelet/src/config.rs
```diff
@@ -633,6 +657,9 @@ mod test {
         assert_eq!(config.allow_local_modules, true);
         assert_eq!(config.node_labels.len(), 2);
         assert_eq!(config.node_labels.get("label1"), Some(&("val1".to_owned())));
+        assert_eq!(config.insecure_registries.clone().unwrap().len(), 2);
+        assert_eq!(&config.insecure_registries.clone().unwrap()[0], "local");
```
thomastaylor312, line 661: Nit: Because this is the end of the test, you could just do something like this to make things a bit less gnarly:

```rust
let insecure_registries = config.insecure_registries.expect("registries")
...do assertions here
```

crates/kubelet/src/lib.rs
```diff
@@ -55,6 +55,7 @@ pub(crate) mod kubeconfig;
 pub(crate) mod webserver;
 
 pub mod config;
+pub mod config_interpreter;
```
thomastaylor312, line 58: Just to clarify: This is `pub` because of the trait implementation in `config_interpreter`?

crates/oci-distribution/src/client.rs
```diff
@@ -47,6 +47,12 @@ pub struct Client {
     client: reqwest::Client,
 }
 
+/// A source that can provide a `ClientConfig`.
```
thomastaylor312, line 50: Nit: It might be good to expand the documentation here a bit to say something like "This is useful when writing your own configuration for an application as it can implement this trait to provide a `ClientConfig`"

crates/oci-distribution/src/reference.rs
```diff
@@ -52,21 +52,21 @@ impl Reference {
     }
 
     /// Convert a Reference to a v2 manifest URL.
-    pub fn to_v2_manifest_url(&self, protocol: &str) -> String {
+    pub fn to_v2_manifest_url(&self, protocol: &crate::client::ClientProtocol) -> String {
```
thomastaylor312, line 55: I really like having this be an enum over a bare `&str`

crates/kubelet/src/lib.rs
itowlson: I thought it needed to be but it looks like having it private still works!  Fixed!  Thanks!

crates/kubelet/src/config.rs
itowlson: Might skip that because it could become a source of unrelated changes if a subsequent PR added test lines after these ones... this whole area could do with a bit more granularity though, to avoid this sort of gnarl!

bacongobbler:
needs a rebase, but LGTM!

itowlson:
Rebase was a bit fiddly, might not hurt to take another look...

PR #364, opened 2020-08-26
Tidy up WSL2 setup commands

Porges:
This can be done in one command instead of running each command and waiting for it to finish before progressing to the next one.

Also fix some incorrect/missing Markdown syntax highlighting annotations.


thomastaylor312:
Don't know why the tests didn't run (do we exclude doc changes?), but as this is a doc change, I am not worried about merging

PR #368, opened 2020-09-01
fix(kubelet): Removes node role label for compatibility with k8s 1.19

thomastaylor312:
This was causing failures when trying to register the node. See the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md#proposal) for more information


thomastaylor312:
We are going to wait for a bit to see if we get any feedback [here](https://github.com/kubernetes/kubernetes/pull/90307#discussion_r481340453) before we merge this

PR #370, opened 2020-09-02
ref(pod): switch from tuple to regular struct; use Pod::from

bacongobbler:
Pod is a wrapper struct around k8s_openapi's Pod, but many callers use `Pod::new(pod)` instead of `Pod::from(pod)`.

This commit removes the constructor, forcing the caller to rely on calling `::from` to instantiate a new Pod. This also allows a Pod to expose more fields in the future (such as a pod's current state).

conflicts with #344, which should be merged prior to this PR to prevent another rebase.


itowlson (approved):

itowlson:
Looks like the extra length of `kube_pod` over `0` pushed `cargo fmt` into a full-blown tantrum - otherwise looks fine!

thomastaylor312 (approved):

PR #371, opened 2020-09-02
The podsmiter

### LINKED ISSUE #366: Simplify recovery if integration test cleanup fails
In some circumstance it's possible for the integration tests to leave pods in an unusable state - like they never went ready but aren't able to terminate either - which also blocks deletion of the namespaces, and causes vexing failures next time you try to run the tests.

You can clean up the naughty pods with kubectl, and the namespaces then recover, but this takes a couple of minutes, so we don't want to defer this to next time the integration tests run.  And it's annoying to do manually because you need to run through a bunch of kubectl commands writing out pod names and `-n` options and remembering to write `--force --grace-period=0`.

It would be good the one-click widget could detect this situation and either:

* Iterate through namespaces of the form `wasi-e2e-*` force-deleting any terminating pods; or
* Emit a podsmiter script that the user could run to force-delete the offending pods.

itowlson:
Fixes #366.  If integration tests fail in a way that leaves pods stuck in `Terminating`, you can `cargo run --bin podsmiter` to delete the pods, after which the namespaces will clear up automatically.

(As an alternative to a post-test cleanup script, we could modify the integration tests to force-delete the pods in its cleanup; I haven't done that because I thought there might be cases where we'd want to look at stuck ones before blowing them away; but it would certainly be easier if we didn't care about that.  Having a script also allows cleanup if the test process fails to clean up for whatever reason.)

### DIFF
--- a/docs/community/developers.md
+++ b/docs/community/developers.md
@@ -167,6 +167,17 @@ kind of Kubernetes cluster you're using and so doesn't know how to infer a node 
 _WARNING:_ The standalone integration tester has not been, er, tested on Windows.
 Hashtag irony.
 
+### Integration test debris
+
+There are some failure modes - for example image pull timeout - where the integration
+tests are not able to complete cleanup of their resources.  Specifically you can
+sometimes get pods stuck in `Terminating`, which prevents namespace cleanup and
+causes the next test run to break.
+
+You can forcibly clean up such debris by running `cargo run --bin podsmiter`.  You
+may need to way a couple of minutes after pod deletion for the namespaces to be
+collected.
+
 ## Creating your own Kubelets with Krustlet
 
 If you want to create your own Kubelet based on Krustlet, all you need to do is implement a
--- a/docs/community/developers.md
+++ b/docs/community/developers.md
@@ -175,7 +175,7 @@ sometimes get pods stuck in `Terminating`, which prevents namespace cleanup and
 causes the next test run to break.
 
 You can forcibly clean up such debris by running `cargo run --bin podsmiter`.  You
-may need to way a couple of minutes after pod deletion for the namespaces to be
+may need to wait a couple of minutes after pod deletion for the namespaces to be
 collected.
 
 ## Creating your own Kubelets with Krustlet
--- a/tests/podsmiter/src/main.rs
+++ b/tests/podsmiter/src/main.rs
@@ -1,6 +1,6 @@
 use k8s_openapi::api::core::v1::{Namespace, Pod};
-use k8s_openapi::Metadata;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
+use k8s_openapi::Metadata;
 use kube::api::{Api, DeleteParams, ListParams};
 
 #[tokio::main(threaded_scheduler)]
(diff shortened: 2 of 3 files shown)


docs/community/developers.md
```diff
@@ -167,6 +167,17 @@ kind of Kubernetes cluster you're using and so doesn't know how to infer a node
...
 
+### Integration test debris
+
+There are some failure modes - for example image pull timeout - where the integration
+tests are not able to complete cleanup of their resources.  Specifically you can
+sometimes get pods stuck in `Terminating`, which prevents namespace cleanup and
+causes the next test run to break.
+
+You can forcibly clean up such debris by running `cargo run --bin podsmiter`.  You
+may need to way a couple of minutes after pod deletion for the namespaces to be
```
thomastaylor312, line 178: s/way/wait/

tests/podsmiter/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    println!("Requested force-delete of all pods; requesting delete of namespaces...");
+
+    let ns_smite_operations = namespaces
+        .iter()
+        .map(|ns| smite_namespace(client.clone(), ns));
+    let ns_smite_results = futures::future::join_all(ns_smite_operations).await;
+    let (_, ns_smite_errors) = ns_smite_results.partition_success();
+
+    if !ns_smite_errors.is_empty() {
+        return Err(anyhow::anyhow!(smite_failure_message(&pod_smite_errors)));
```
thomastaylor312, line 57: Nit: Perhaps have `smite_failure_message` return an `anyhow::Error`

tests/podsmiter/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+async fn list_e2e_namespaces(client: kube::Client) -> anyhow::Result<Vec<String>> {
+    println!("Finding e2e namespaces...");
+
+    let nsapi: Api<Namespace> = Api::all(client.clone());
+    let nslist = nsapi.list(&ListParams::default()).await?;
+
+    Ok(nslist
+        .iter()
+        .map(name_of)
+        .filter(|n| is_e2e_namespace(n))
```
thomastaylor312, line 72: Nit: I think this can just be shortened to `filter(is_e2e_namespace)`. Clippy might even complain about it as is right now

tests/podsmiter/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        .collect())
+}
+
+fn name_of(ns: &impl Metadata<Ty = ObjectMeta>) -> String {
+    ns.metadata().unwrap().name.as_ref().unwrap().to_owned()
+}
+
+fn is_e2e_namespace(namespace: &str) -> bool {
+    namespace.starts_with("wascc-e2e") || namespace.starts_with("wasi-e2e")
```
thomastaylor312, line 81: Nit: These namespace strings should probably be constants

thomastaylor312 (commented):
This looks super handy (and is hilarious). Let me know when tests are passing and I'll do a final review

tests/podsmiter/src/main.rs
itowlson: I thought that but apparently not; if I just pass the function directly then I get a type mismatch error.  The lambda form seems to unify `&str` and `&String` in a way that the direct form doesn't.  It's weird, and I agree that what you describe _should_ work... but...

itowlson:
Fixed your comments @thomastaylor312 - thanks!  Ready for you to take another look when convenient.

thomastaylor312 (approved):

itowlson:
After discussion we are merging this despite the ongoing CI issues because we believe those relate to the state machine refactoring in progress.

PR #378, opened 2020-09-03
fix compiler error

bacongobbler:
This hack disables graceful node shutdowns, but it allows the project to compile.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

PR #379, opened 2020-09-04
new-and-improved WASI provider

bacongobbler:
\*slaps roof of car\* This bad boy can fit so much WebAssembly in it

slaps-roof-of-car-meme-2

This PR is temporarily rebased on top of #378 until we find a fix for the root cause. Commit 28adc3b contains the new WASI provider logic.


crates/kubelet/src/provider/mod.rs
```diff
@@ -135,6 +135,38 @@ pub trait Provider: Sized {
...
 }
 
+/// Resolve the environment variables for a container.
+///
+/// This generally should not be overwritten unless you need to handle
+/// environment variable resolution in a special way, such as allowing
+/// custom Downward API fields.
+///
+/// It is safe to call from within your own providers.
+pub async fn env_vars(
```
bacongobbler, line 145: note: this comes from the Provider trait, a few lines above. Now that the logic is internal to the state machine transitions rather than the provider, we should look into refactoring this out of `mod provider` so it can be accessible to all state machines.

crates/wasi-provider/src/states/registered.rs
```diff
@@ -0,0 +1,156 @@
...
+    state,
+};
+
+use super::error::Error;
+use super::image_pull::ImagePull;
+
+fn validate_pod_runnable(pod: &Pod) -> anyhow::Result<()> {
+    if !pod.init_containers().is_empty() {
+        return Err(anyhow::anyhow!(
+            "Cannot run {}: spec specifies init containers which are not supported on wasCC",
```
bacongobbler, line 17: TODO: re-introduce init containers.

crates/wasi-provider/src/states/registered.rs
```diff
@@ -0,0 +1,156 @@
...
+    for container in pod.containers() {
+        validate_container_runnable(&container)?;
+    }
+    Ok(())
+}
+
+fn validate_container_runnable(container: &Container) -> anyhow::Result<()> {
+    if has_args(container) {
+        return Err(anyhow::anyhow!(
+            "Cannot run {}: spec specifies container args which are not supported on wasCC",
```
bacongobbler, line 30: TODO: validate whether the WASI provider accepted container arguments.

itowlson: It did.

crates/wasi-provider/src/lib.rs
```diff
@@ -85,185 +82,63 @@ impl WasiProvider {
...
-            Ok(())
-        } else {
-            Err(anyhow::anyhow!(
-                "Init container {} failed with message {}",
-                container.name(),
-                result.message
-            ))
-        }
-    }
+fn make_status(phase: Phase, reason: &str) -> anyhow::Result<serde_json::Value> {
```
kflansburg, line 101: Should we move this function to `kubelet` for now?

crates/wasi-provider/src/states/error.rs
```diff
@@ -0,0 +1,39 @@
+use kubelet::pod::{Phase, Pod};
```
kflansburg, line 1: Do we want to make a `kubelet::state:prelude` that includes these things?

crates/wasi-provider/src/states/registered.rs
```diff
@@ -0,0 +1,156 @@
...
+use kubelet::{
+    container::Container,
+    pod::{Phase, Pod},
+    state,
+};
+
+use super::error::Error;
+use super::image_pull::ImagePull;
+
+fn validate_pod_runnable(pod: &Pod) -> anyhow::Result<()> {
```
kflansburg, line 14: These look lifted from the `wascc` provider. Are the necessary for `wasi`?

crates/wasi-provider/src/states/running.rs
```diff
@@ -0,0 +1,21 @@
...
+use crate::{make_status, PodState};
+
+state!(
+    /// The Kubelet is running the Pod.
+    Running,
+    PodState,
+    {
+        // Wascc has no notion of exiting so we just sleep.
+        // I _think_ that periodically awaiting will allow the task to be interrupted.
+        loop {
```
kflansburg, line 16: `wascc` has no notion of "completion" but `wasi` appears to (based on init container logic). I think it would make sense for this state to poll for completion of one or more containers and move to completed / failed accordingly.

kflansburg (approved):
Looks good! How much of a burden would you say the rewrite was? It looks like the `wascc` and `wasi` providers will be good examples to guide updates to other providers.

kflansburg (changes requested):
Looks like I got it building with functionality restored. I used suggested `#![type_length_limit = "1271125"]` fix. Hopefully those can be removed with the next rust version. E2E tests still failing, it doesn't look straightforward to get these all working (admittedly I've never run these locally, so some of this may be misconfiguration): 
```
running 10 tests
test test_container_args ... FAILED
test test_container_logging ... FAILED
test test_failing_init_containers ... FAILED
test test_can_mount_individual_values ... FAILED
test test_can_mount_multi_values ... FAILED
test test_wascc_provider ... ok
test test_wasi_node_should_verify ... ok
test test_init_containers ... FAILED
test test_module_exiting_with_error ... FAILED
test test_pod_logs_and_mounts ... FAILED

failures:

---- test_container_args stdout ----
Error: ApiError: No API token found for service account "default", retry after the token is automatically created and added to the service account: ServerTimeout (ErrorResponse { status: "Failure", message: "No API token fo
und for service account \"default\", retry after the token is automatically created and added to the service account", reason: "ServerTimeout", code: 500 })

Caused by:
    No API token found for service account "default", retry after the token is automatically created and added to the service account: ServerTimeout
thread 'test_container_args' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /Users/kevin/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/macros.rs:16:9

---- test_container_logging stdout ----
thread 'test_container_logging' panicked at 'pod loggy-pod never went ready', tests/pod_setup.rs:90:5

---- test_failing_init_containers stdout ----
thread 'test_failing_init_containers' panicked at 'pod faily-inits-pod never went ready', tests/pod_setup.rs:90:5

---- test_can_mount_individual_values stdout ----
thread 'test_can_mount_individual_values' panicked at 'pod multi-mount-items-pod never went ready', tests/pod_setup.rs:90:5

---- test_can_mount_multi_values stdout ----
thread 'test_can_mount_multi_values' panicked at 'pod multi-mount-pod never went ready', tests/pod_setup.rs:90:5

---- test_init_containers stdout ----
thread 'test_init_containers' panicked at 'pod hello-wasi-with-inits never went ready', tests/pod_setup.rs:90:5

---- test_module_exiting_with_error stdout ----
thread 'test_module_exiting_with_error' panicked at 'pod faily-pod never went ready', tests/pod_setup.rs:90:5

---- test_pod_logs_and_mounts stdout ----
thread 'test_pod_logs_and_mounts' panicked at 'Could not fetch terminated states', tests/assert.rs:58:6


failures:
    test_can_mount_individual_values
    test_can_mount_multi_values
    test_container_args
    test_container_logging
    test_failing_init_containers
    test_init_containers
    test_module_exiting_with_error
    test_pod_logs_and_mounts

test result: FAILED. 2 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out
```

bacongobbler:
@kflansburg with the `#![type_length_limit = "1271125"]` workaround, where did you place that in the code? Do you have a diff with your changes available? It would be great to PR that in place of #378!

bacongobbler:
With regards to your earlier questions:

> How much of a burden would you say the rewrite was?

At this point, it was fairly simple. The hard part now has to do with mapping functions that used to be part of the provider into the state machine. For example, the WASI provider had a few useful functions to determine the correct mount paths for volumes (`volume_path_map`), or to start the module (`start_container()`). Those had to be moved over to the state machine.

I found that some of the logic from the waSCC provider were also relevant to the WASI provider, such as `make_status` and `PodState`. I think there's an opportunity to refactor code here to prevent code duplication.

One part that seemed odd during this refactor was the compiler errors that occurred when incorrect syntax was provided in one of the state macros. And as @itowlson pointed out, autocompletion does not work very well within a macro, so it makes it harder to know what fields and functions are available within `pod`, `handles`, and `pod_state`.

All in all, the migration wasn't too difficult. Now I just need to finish off some of the wasmtime specific functionality like init containers and container arguments. Those may take more time to complete, but I don't imagine it'll be too difficult.

crates/wasi-provider/src/states/registered.rs
bacongobbler: No, they're unnecessary. I left these in as a TODO reminder. The first stage was to get it compiling again.

kflansburg:
> @kflansburg with the `#![type_length_limit = "1271125"]` workaround, where did you place that in the code? Do you have a diff with your changes available? It would be great to PR that in place of #378!

This line must be added at the top of the two files in the `src` directory (i.e. it needs to apply to the two binaries). If we want to split this into a separate pull request, I can do that tonight.

kflansburg:
RE: compiler errors in macros, I agree, it seems unable to supply line numbers or even correct file names.

thomastaylor312:
We had mentioned this in the state machine PR, but what do we think about dropping the `state!` macro in favor of manual implementation? It isn't terribly difficult to do and would lead to better autocompletion support (and be easier to understand IMO)

bacongobbler:
> If we want to split this into a separate pull request, I can do that tonight.

Thank you! I went ahead and did that with https://github.com/deislabs/krustlet/pull/381. I'll go and rebase this PR against that one and close #378.

bacongobbler:
> We had mentioned this in the state machine PR, but what do we think about dropping the `state!` macro in favor of manual implementation? It isn't terribly difficult to do and would lead to better autocompletion support (and be easier to understand IMO)

I'm not entirely sure what would be involved here, but it may be worth looking into. Especially since this is a main code loop rather than a helper function like `debug!` or `hashmap!`.

kflansburg:
I’m in favor of removing the state macro but I don’t think it needs to be in this pull request. That feels like more of an ergonomic thing to iterate on. I would rather see us get tests passing / restore any functionality that we’ve broken. Like I mentioned in the last thread, users are free to ignore the macro and use the trait instead so we aren’t blocking that

kflansburg:
That being said, it should be pretty mechanical so if anyone has a spare hour I’m fine either way.

bacongobbler:
Agreed. Let's punt the macro refactor to another PR and focus on re-implementing the WASI provider. The current system's fine to work with. I just want to make sure we can get a second provider working with the new state machine, and we can iterate on its design after we get the e2e tests back on its feet.

kflansburg:
@bacongobbler Do you want to split up the work on these e2e test? I was thinking I would try to tackle the wascc ones.

bacongobbler:
That would be wonderful! I've been focusing on the WASI provider. I'd love some help getting the waSCC provider to pass.

kflansburg:
Great, I think I have my environment configured now for running the e2e tests. I will see what I can do.

kflansburg:
@bacongobbler The `test_wascc_provider` test appears to be the only `wascc` one. I believe it is passing although occasionally it seems to timeout too soon when waiting for the pod to start. Perhaps we should discuss splitting up the `wasi` tests since there are more of them.

crates/wasi-provider/src/states/error.rs
bacongobbler: Good idea!

crates/wasi-provider/src/lib.rs
bacongobbler: I agree.

kflansburg:
@bacongobbler let me know if you want to coordinate on fixing the wasi tests, not sure where you are at on that. I will do the prelude and make_status for now.

thomastaylor312:
@kflansburg We are much closer now, still just figuring out some things with an image pull secret problem and working through each test failure. Can't speak to whether or not @bacongobbler would like another set of eyes

crates/wasi-provider/src/states/running.rs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        let client: Api<KubePod> = Api::namespaced(
+            kube::Client::new(pod_state.shared.kubeconfig.clone()),
+            pod.namespace(),
+        );
+        let mut completed = 0;
+        let total_pods = pod.containers().len();
+
+        while let Some((name, status)) = pod_state.run_context.status_recv.recv().await {
+            if let Err(e) = client
```
kflansburg, line 24: One of my goals of the state machine was to eliminate any pod status patches outside of the context of the state machine. Can this be handled by transitioning to an error state or adding an error message to the Terminated state?

crates/wasi-provider/src/states/running.rs
```diff
@@ -0,0 +1,53 @@
...
+                    return Ok(Transition::Error(Box::new(Error { message })));
+                } else {
+                    completed += 1;
+                    if completed == total_pods {
+                        return Ok(Transition::Advance(Box::new(Terminated)));
+                    }
+                }
+            }
+        }
+        Ok(Transition::Advance(Box::new(Terminated)))
```
kflansburg, line 50: I think technically the states are Completed and CrashLoopBackoff.

bacongobbler:
I'm wiped for the day, but we'd love some help getting the rest of the e2e tests working if you'd like to help out on that.

When we last left our heroes:

```
><> just test-e2e-standalone
[...]
Integration tests running

running 11 tests
test test_pull_from_private_registry ... ok
test test_wasi_node_should_verify ... ok
test test_wascc_provider ... ok
test test_container_args ... ok
test test_failing_init_containers ... FAILED
test test_container_logging ... ok
test test_pod_logs_and_mounts ... FAILED
test test_init_containers ... FAILED
test test_module_exiting_with_error ... FAILED
test test_can_mount_multi_values ... FAILED
test test_can_mount_individual_values ... FAILED

failures:

---- test_failing_init_containers stdout ----
thread 'test_failing_init_containers' panicked at 'Could not get pod message', tests/assert.rs:83:46

---- test_pod_logs_and_mounts stdout ----
thread 'test_pod_logs_and_mounts' panicked at 'Could not fetch terminated states', tests/assert.rs:58:6

---- test_init_containers stdout ----
thread 'test_init_containers' panicked at 'pod hello-wasi-with-inits never went ready', tests/pod_setup.rs:90:5

---- test_module_exiting_with_error stdout ----
thread 'test_module_exiting_with_error' panicked at 'Could not fetch terminated states', tests/assert.rs:108:6

---- test_can_mount_multi_values stdout ----
thread 'test_can_mount_multi_values' panicked at 'Could not fetch terminated states', tests/assert.rs:58:6

---- test_can_mount_individual_values stdout ----
thread 'test_can_mount_individual_values' panicked at 'Could not fetch terminated states', tests/assert.rs:58:6
```

crates/wasi-provider/src/states/starting.rs
```diff
@@ -0,0 +1,136 @@
...
+
+state!(
+    /// The Kubelet is starting the Pod.
+    Starting,
+    PodState,
+    {
+        let mut container_handles = HashMap::new();
+
+        info!("Running init containers for pod {:?}", pod.name());
+        for init_container in pod.init_containers() {
```
kflansburg, line 106: I patched the `init_containers` method, it was returning the Pod's app containers. Was this intentional? How was it passing tests before?

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -294,29 +311,53 @@ impl WasiRuntime {
...
+                &mut cx,
+            );
             Ok(())
         });
         // Wait for the interrupt to be sent back to us
         let interrupt = rx.await?;
         Ok((interrupt, handle))
     }
 }
+
```
kflansburg, line 346: I understand the need for this channel, but it seems to be implemented in very cumbersome way here. Am I missing something?

kflansburg:
@bacongobbler Cool, I will claim `test_pod_logs_and_mounts` and `test_failing_init_containers` and look into those tomorrow.

crates/wasi-provider/src/states/running.rs
bacongobbler: If my hazy memory serves right... This was necessary to send container status updates to the API. `make_status` does not have a way to send patch updates, which is necessary for the e2e tests to pass.

crates/wasi-provider/src/states/running.rs
bacongobbler: Good catch.

crates/wasi-provider/src/states/running.rs
bacongobbler: https://github.com/bacongobbler/krustlet/blob/02d15a8e6104b885a6704c4f8b4eec0722cbe36f/crates/kubelet/src/pod/status.rs#L18-L19

crates/wasi-provider/src/states/starting.rs
bacongobbler: No, I don't think that was intentional. Thank you for catching that.

I see your change in https://github.com/bacongobbler/krustlet/commit/02d15a8e6104b885a6704c4f8b4eec0722cbe36f to `all_containers` accidentally iterating through `self.containers` twice, but I don't see a fix for `pod.init_containers()`. Can you point out the patch for `init_containers()`?

crates/wasi-provider/src/wasi_runtime.rs
bacongobbler: We changed the internals from a `tokio::sync::watch` to a `tokio::sync::mpsc::Sender`. The former did not return a future on a `.broadcast`, but `.send` does. Because async calls like `await` cannot be used inside non-async threads (like the one spawned in `spawn_wasmtime`), thus born the cumbersome `send`.

In reality these are all gross hacks to get the tests passing again. Once that's done we can take a step back and remove some of this tech debt. If you have a suggestion how we could make this better in the meantime, we'd be happy to hear your feedback.

crates/wasi-provider/src/states/running.rs
bacongobbler: But yes, the eventual goal is to eliminate these pod status patches from the control loop entirely.

kflansburg: I do agree we need a better story around container status updates.

crates/wasi-provider/src/states/starting.rs
kflansburg: Ah, I misremembered what I patched, it sounds like `all_containers` was the only method impacted.

crates/wasi-provider/src/wasi_runtime.rs
kflansburg: Ah, I did not catch that the calling function was not Async. Makes sense.

crates/wasi-provider/src/states/running.rs
thomastaylor312: I think the current idea is to make each container have its own state machine (separate from the pod one) and those state machines would take care of updating the status

thomastaylor312:
@kflansburg I think `test_pod_logs_and_mounts` is no longer needed (it is covered by other tests) and @bacongobbler is working on the init containers stuff. We know what was wrong, and we made a bunch of progress today

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -0,0 +1,59 @@
...
+                    completed += 1;
+                    if completed == total_containers {
+                        return Ok(Transition::Advance(Box::new(Starting)));
+                    }
+                }
+            }
+        }
+        Ok(Transition::Advance(Box::new(Starting)))
+    },
+    { make_status(Phase::Pending, "Initializing") }
```
thomastaylor312, line 58: I think the proper phase during init containers is `Running`

kflansburg:
@bacongobbler @thomastaylor312 I made a pretty chunky commit to address two of our open issues and go ahead and remove the `state` macro. Check out the commit message for more details. I will do a proper review tomorrow.

crates/kubelet/src/state.rs
```diff
@@ -1,25 +1,221 @@
...
+    ///     async fn json_status(
+    ///         &self,
+    ///         _pod_state: &mut OtherPodState,
+    ///         _pod: &Pod,
+    ///     ) -> anyhow::Result<serde_json::Value> {
+    ///         Ok(serde_json::json!(null))
+    ///     }
+    /// }
+    /// ```
+    pub fn next<I: State<PodState>, S: State<PodState>>(_i: &I, s: S) -> Transition<PodState>
```
thomastaylor312, line 200: This feels a little weird to make the user pass the old state in order to enforce the edge constraint. Is there any other way around this? I can't think of any and I don't know what you already tried

crates/wascc-provider/src/states/crash_loop_backoff.rs
```diff
@@ -1,20 +1,30 @@
...
+    async fn json_status(
+        &self,
+        _pod_state: &mut PodState,
+        _pod: &Pod,
+    ) -> anyhow::Result<serde_json::Value> {
+        make_status(Phase::Pending, "CrashLoopBackoff")
+    }
+}
+
+impl EdgeTo<Registered> for CrashLoopBackoff {}
```
thomastaylor312, line 30: Could the `EdgeTo` trait be better exposed with a `derive` macro in the future?

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -0,0 +1,83 @@
...
+            let handle = start_container(pod_state, pod, &init_container).await?;
+
+            while let Some((_name, status)) = pod_state.run_context.status_recv.recv().await {
+                // TODO: implement a container state machine such that it will self-update the Kubernetes API as it transitions through these stages.
+                //let s = serde_json::json!({
+                //    "metadata": {
+                //        "resourceVersion": "",
+                //    },
+                //    "status": {
+                //        "containerStatuses": [
```
thomastaylor312, line 41: This will need to be `initContainerStatuses` when you uncomment

thomastaylor312 (commented):
@kflansburg I am still not sure how I feel about the `EdgeTo` trait. It seems like an extra step for implementors that doesn't add tons of additional compile time safety. A user still has to specify the `EdgeTo` manually, which you already had to pass a proper `State` implementation with `Transition` anyway.

crates/wasi-provider/src/states/initializing.rs
bacongobbler: Nice catch.

crates/kubelet/src/state.rs
kflansburg: I spent quite a while to try to avoid this but couldn’t come up with anything. Happy to remove it if we can come up with another way, but I don’t think it’s possible since this is a class method and has no context.

crates/wascc-provider/src/states/crash_loop_backoff.rs
kflansburg: I’m definitely open to exploring other ways of setting the edge trait, I’ve never actually written a derive macro so I’m not sure.

kflansburg:
I think whitelisting the edges is pretty important. While it’s true that you must transition to a valid state, when we moved to the Box we eliminated the explicit specification of valid outgoing edges for each state and essentially made it a fully connected graph. I think, especially with multiple contributors to a provider project, that making the edges explicit will make it much harder for contributors to jump to a state that is unexpected. At the very least they will have to add the edge and reviewers will be reminded to consider the implication of that.

thomastaylor312:
Ok, I think I am fine with that then

bacongobbler:
rebased, squashed, and ready for final review.

crates/wasi-provider/src/states/starting.rs
```diff
@@ -0,0 +1,139 @@
...
+    runtime.start().await
+}
+
+pub(crate) type ContainerHandleMap =
+    HashMap<ContainerKey, kubelet::container::Handle<Runtime, HandleFactory>>;
+
+#[derive(Default, Debug)]
+/// The Kubelet is starting the Pod containers
+pub(crate) struct Starting {
+    init_handles: Arc<Mutex<ContainerHandleMap>>,
```
thomastaylor312, line 87: Note for other reviewers, this was necessary to have interior mutability as the `next` method is not `mut self`

thomastaylor312 (approved):
Looks good. Great work on getting this all together

crates/wasi-provider/src/states/completed.rs
```diff
@@ -0,0 +1,16 @@
+use crate::PodState;
+use kubelet::state::prelude::*;
+
+state!(
+    /// Pod was deleted.
```
itowlson, line 5: Wait.... wouldn't a Deleting state be different from a Completed state?  E.g. a job pod could complete but would remain in existence.

crates/wasi-provider/src/states/error.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[async_trait::async_trait]
+impl State<PodState> for Error {
+    async fn next(
+        &self,
+        pod_state: &mut PodState,
+        _pod: &Pod,
+    ) -> anyhow::Result<Transition<Box<dyn State<PodState>>, Box<dyn State<PodState>>>> {
+        pod_state.errors += 1;
+        if pod_state.errors > 3 {
```
itowlson, line 22: Nit: Magic number?

crates/wasi-provider/src/states/error.rs
```diff
@@ -0,0 +1,38 @@
...
+            pod_state.errors = 0;
+            Ok(Transition::Error(Box::new(CrashLoopBackoff)))
```
itowlson, line 24: Why do you need to distinguish between an error transition and an advance transition?  Aren't both just "move to the specified state"?  Neither transition here seems specifically like an 'error' transition - they are just different ways of proceeding based on the state data.

crates/wasi-provider/src/states/finished.rs
```diff
@@ -0,0 +1,10 @@
+use crate::PodState;
+use kubelet::state::prelude::*;
+
+state!(
+    /// Pod execution completed with no errors.
+    Finished,
```
itowlson, line 6: The difference between `Finished` and `Completed` is elusive, especially as the documentation comment for `Finished` uses the term `completed`!  Can we find names that better express the distinction?  (I understand that the answer may be "no.")

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -0,0 +1,138 @@
...
+use super::error::Error;
+use super::starting::{start_container, ContainerHandleMap, Starting};
+
+async fn patch_init_status(
+    client: &Api<KubePod>,
+    pod_name: &str,
+    name: String,
+    status: &ContainerStatus,
+) -> anyhow::Result<()> {
+    // We need to fetch the current status because there is no way to merge with a strategic merge patch ere
```
itowlson, line 21: Super nit: Possible typo in last word of comment.

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -0,0 +1,138 @@
...
+        .await?;
+    Ok(())
+}
+
+state!(
+    /// The Kubelet is running the init containers for the pod.
+    Initializing,
+    PodState,
+    {
+        let client: Api<KubePod> = Api::namespaced(
```
itowlson, line 68: This all looks fine in itself, but is strong evidence for the macro rework.  This is a 70-line function of some complexity; maintaining it would be a lot easier if the macro allowed for better editor support.

crates/wasi-provider/src/states/starting.rs
itowlson: I'm curious why a state would need a mutable member in the design we have - isn't all the work and interior state essentially local to the `next` method?  How is this used that it could not be local to `next`?  (I'm not challenging you - I am trying to clarify my understanding of the state machine design.)

itowlson (commented):
Had a quick look (although 'quick' being relative in this case) and I don't think I have anything critical, but do have some questions and style nits.  Probably a higher priority to land this - we can always tidy in tandem with the state machine enhancements coming down the pipe.

crates/wasi-provider/src/states/starting.rs
thomastaylor312: We need to pass the current handles from the init container down the line. This is purely internal to this state. In this specific instance, because we need to take ownership of these handles, we must make the vec mutable so as to be able to drain it and take ownership of its contents. That clarify it at all?

crates/wasi-provider/src/lib.rs
```diff
@@ -85,185 +83,49 @@ impl WasiProvider {
...
+            let mut handles = self.shared.handles.write().await;
+            handles.remove(&self.key);
         }
     }
 }
 
 #[async_trait::async_trait]
 impl Provider for WasiProvider {
+    type InitialState = Registered;
+    type TerminatedState = Completed;
```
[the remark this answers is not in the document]
thomastaylor312: I actually think this should be Terminating as it is meant to be the state it kicks to if it receives a delete event. But either way I think this name might be good to change in the other cleanup PR

crates/wasi-provider/src/states/error.rs
thomastaylor312: Most definitely a magic number but I think this will go away with the implementation of exponential backoff

crates/wasi-provider/src/states/starting.rs
kflansburg: Can container handles be a part of `PodState`?

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -294,29 +311,53 @@ impl WasiRuntime {
...
+fn send(mut sender: Sender<(String, Status)>, name: String, status: Status, cx: &mut Context<'_>) {
```
kflansburg, line 347: Sorry, to reopen this conversation, but it seems that the use of `tokio::Sender` is defined in the `ModuleRunContext` of this crate. I had changed the channel in `Pod::queue`, yes, but I dont think that means we have to use it everywhere. It resulted in cleaner code in `Pod::queue`, but clearly not here. I dont like the fact that `tokio::sync::watch` drops messages, but I think it would be ok here.

kflansburg (approved):
Looks great, thanks for all of your work on this @bacongobbler !

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -0,0 +1,138 @@
...
+use super::starting::{start_container, ContainerHandleMap, Starting};
+
+async fn patch_init_status(
+    client: &Api<KubePod>,
+    pod_name: &str,
+    name: String,
+    status: &ContainerStatus,
+) -> anyhow::Result<()> {
+    // We need to fetch the current status because there is no way to merge with a strategic merge patch ere
+    let mut init_container_statuses = match client.get(pod_name).await {
```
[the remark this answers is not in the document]
thomastaylor312: So we actually tried this first. We have to have stable array elements to do so with a JSON Patch, which is possible here as we have to serially iterate through and run each container to completion. So it would technically be possible here, but not be possible in the `running` state because we'd be receiving updates from different containers within the pod at any time and would have to store an index mapping. That could be fine to do, but I didn't want to add yet another mapping to keep track of things right now

crates/wasi-provider/src/wasi_runtime.rs
thomastaylor312: Because we are using a shared channel for all containers, we can't use watch or we'll end up dropping status events if there are multiple containers

crates/wasi-provider/src/states/starting.rs
thomastaylor312: It feels like an anti-pattern to have _anything_ that needs to be sent to a single specific state shoved into a shared pod state across the whole thing. We only need to forward these handles to the `Starting` state and do not need to track them otherwise. We did try to change the trait for `next` to be `self` instead of `&self` as it kinda makes sense for it to consume itself (as it always returns another state), but that gave the compiler problems. We could also trying making it `&mut self` though, which I can't believe I didn't think of doing.

crates/wascc-provider/src/states/registered.rs
```diff
@@ -31,6 +27,11 @@ fn validate_container_runnable(container: &Container) -> anyhow::Result<()> {
             container.name()
         ));
     }
+    if let Some(image) = container.image()? {
```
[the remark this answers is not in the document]
bacongobbler: Yeah. This is slightly orthogonal, but without it we get panic dump traces during the ImagePull state as `kube-proxy` is a docker image, not webassembly. I think we need to investigate this a little further, but I put this in to get things working.

I'll file a ticket to look further into how we can disambiguate between media types when we pull images.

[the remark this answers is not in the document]
bacongobbler: filed https://github.com/deislabs/krustlet/issues/387 to track this work

crates/wasi-provider/src/lib.rs
```diff
@@ -34,45 +34,43 @@
...
 /// WasiProvider provides a Kubelet runtime implementation that executes WASM
-/// binaries conforming to the WASI spec
+/// binaries conforming to the WASI spec.
 #[derive(Clone)]
 pub struct WasiProvider {
+    shared: SharedPodState,
+}
+
+#[derive(Clone)]
+struct SharedPodState {
```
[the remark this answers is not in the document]
bacongobbler: As I understand it, this is a shared state used between all pods, so your assessment is correct. I've copied this over from the waSCC provider, so perhaps @kflansburg can shed some more light onto this.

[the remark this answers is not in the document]
bacongobbler: Looks like he already did in https://github.com/deislabs/krustlet/pull/379#discussion_r489448636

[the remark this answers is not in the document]
bacongobbler: Looks like he already did in https://github.com/deislabs/krustlet/pull/379#discussion_r489448636

crates/wasi-provider/src/states/completed.rs
bacongobbler: okay. Renamed to `Terminated`.

crates/wasi-provider/src/states/finished.rs
bacongobbler: oh! I see the difference between the two states now. Yes, this should be Completed, and the current "Completed" state is actually the Terminated state.

bacongobbler: renamed to reflect this.

crates/wasi-provider/src/states/initializing.rs
bacongobbler: yep

crates/wasi-provider/src/lib.rs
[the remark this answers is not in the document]
bacongobbler: renamed as Terminated.

bacongobbler:
Thank you everyone for your feedback. I'll echo @itowlson's thoughts and say that there certainly are more improvements we can make here, but I think it's a high priority to land this into the main branch sooner rather than later. Keeping this in a PR state for too long will eventually block us from doing other work as the e2e tests are currently failing in the main branch.

That said, I hope that everyone will continue to file tickets or open PRs to address some of the design nits like the `state!` macro being removed in #385, and other changes noted here like #387 to remove the ugly kube-proxy hack.

Thank you all for participating!

PR #380, opened 2020-09-07
Image pull secrets CI test

itowlson:
The testing work for image pull secrets that we deferred to avoid rebase woes

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -598,8 +598,12 @@ async fn create_private_registry_pod(
     let pod_name = PRIVATE_REGISTRY_POD;
 
     let containers = vec![
-        WasmerciserContainerSpec::named("floofycat").with_args(&["write(lit:slats)to(stm:stdout)"]).private(),
-        WasmerciserContainerSpec::named("neatcat").with_args(&["write(lit:kiki)to(stm:stdout)"]).private(),
+        WasmerciserContainerSpec::named("floofycat")
+            .with_args(&["write(lit:slats)to(stm:stdout)"])
+            .private(),
+        WasmerciserContainerSpec::named("neatcat")
+            .with_args(&["write(lit:kiki)to(stm:stdout)"])
+            .private(),
     ];
 
     wasmercise_wasi(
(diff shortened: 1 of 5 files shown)


crates/kubelet/src/secret/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct RegistryAuthResolver {
+    kube_client: kube::Client,
+    pod_namespace: String,
+    image_pull_secret_names: Vec<String>,
+}
+
+impl RegistryAuthResolver {
+    /// Creates a resolver for the given pod
+    pub fn new(client: kube::Client, pod: &crate::pod::Pod) -> Self {
+        // TODO: is it safe to capture this stuff or might we need to re-resolve e.g.
```
thomastaylor312, line 17: As far as I understand about the pod spec and how we treat a modify, I think it is perfectly safe to capture this

tests/integration_tests.rs
```diff
@@ -14,6 +14,10 @@ use pod_builder::{
 use pod_setup::{wait_for_pod_complete, wait_for_pod_ready, OnFailure};
 use test_resource_manager::{TestResource, TestResourceManager, TestResourceSpec};
 
+fn in_ci_environment() -> bool {
+    std::env::var("KRUSTLET_TEST_ENV") == Ok("ci".to_owned())
```
thomastaylor312, line 18: I think it would probably be better to use a `cargo` feature to enable/disable these tests. So then in CI we could run with `cargo test ... --features ci` and annotate the test functions with `#[cfg(feature = "ci")]`. That feels more explicit and Rust-y than an env variable to me. Could just be bike shedding here though

tests/integration_tests.rs
```diff
@@ -584,20 +568,14 @@ async fn create_pod_with_failing_init_container(
...
-            name: "init-that-fails",
-            args: &["assert_exists(file:/nope.nope.nope.txt)"],
-        },
-        WasmerciserContainerSpec {
-            name: "init-that-would-succeed-if-it-ran",
-            args: &["write(lit:slats)to(stm:stdout)"],
-        },
+        WasmerciserContainerSpec::named("init-that-fails")
+            .with_args(&["assert_exists(file:/nope.nope.nope.txt)"]),
+        WasmerciserContainerSpec::named("init-that-would-succeed-if-it-ran")
```
thomastaylor312, line 573: I really like the new builder pattern way of doing this

tests/pod_builder.rs
```diff
@@ -10,6 +10,27 @@ pub struct PodLifetimeOwner {
 pub struct WasmerciserContainerSpec {
     pub name: &'static str,
     pub args: &'static [&'static str],
+    pub use_private_registry: bool,
```
thomastaylor312, line 13: Super Nit: If we are going to use the builder pattern, we should unexport these.

thomastaylor312 (approved):
Looking good. There are some optional points of feedback

tests/integration_tests.rs
itowlson: Oh!  Yes!  I tried that but misunderstood the attribute syntax so it didn't work and I went back to this solution.  I agree the Cargo feature approach is nicer, but I'll defer that to a second PR.

itowlson:
After discussion we are merging this despite the ongoing CI issues because we believe those relate to the state machine refactoring in progress.  If this is still tragic after those are resolved we can come back to it.

PR #381, opened 2020-09-08
increase type length limit

bacongobbler:
supersedes #378 by increasing the type length limit to work around an issue in the Rust compiler with nested closure wrappers (see https://github.com/rust-lang/rust/issues/54540 and https://github.com/rust-lang/rust/issues/75992 for more information).


thomastaylor312 (approved):
Good to go once tests pass

PR #383, opened 2020-09-10
Run secrets tests only when secrets are available

itowlson:
Currently E2E tests on PRs are failing because... well, okay, lots of reasons right now, but a fundamental one is that PRs don't have access to the secret needed to run the image pull secrets test, and there's no way we know of to grant that access to 'approved' PRs.

This PR therefore separates the `build.yml` workflow that does public E2E tests and can run on every PR and a new `ci.yml` workflow that also runs tests that use secrets, and runs only after a PR merges to the main branch.

Ironically, I can't see a way to test the on-merge-to-master workflow without, er, merging it to master.

**REVIEWERS:** It's a bit awful duplicating the bodies of the workflows, but I couldn't see a way to conditionally set environment variables depending on the trigger, nor to pull out the bulk into a shared workflow with configurable knobs.  Let me know if you know ways of doing one of these!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: Build and Test
 on:
-  pull_request:
+  push:
+    branches:
+      - master
     paths-ignore:
       - "docs/**"
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
+name: Build and Test
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - "docs/**"
+jobs:
+  build:
```
thomastaylor312, line 9: We could put these in the same file and then use a YAML anchor to "copy" it in, instead of duplicating. Maybe that will work?

thomastaylor312 (approved):
Approved, with one possible solution to the duplication

itowlson:
Merging on approval alone because the test failures are ongoing and unrelated.  Will review duplication once we have this in main and have confirmed it works.

PR #385, opened 2020-09-15
State machine improvements

bacongobbler:
pulling out @kflansburg's enhancements in #379 into a separate PR as per earlier discussions


crates/kubelet/src/state.rs
```diff
@@ -104,3 +291,61 @@ impl<P: 'static + Sync + Send> State<P> for Stub {
...
+        async fn json_status(
+            &self,
+            _pod_state: &mut PodState,
+            _pod: &Pod,
+        ) -> anyhow::Result<serde_json::Value> {
+            Ok(serde_json::json!(null))
+        }
+    }
+
+    #[test]
```
itowlson, line 325: I think I'm missing the test here.  It seems to verify that the `impl for TestState` compiles, but I'm missing where it checks that a pod operating under this state machine moves correctly through the states.  (This is quite possibly me missing things.)

More significantly, I can't think of a way to test that the `EdgeTo` mechanism is serving its purpose by _preventing_ the state machine from traversing undeclared edges.  This would involve asserting that a piece of code does _not_ compile and I'm not sure how to do that short of invoking the Rust compiler from the test itself e.g. via a process exec call.

crates/wascc-provider/src/states/crash_loop_backoff.rs
```diff
@@ -1,20 +1,30 @@
...
+    async fn json_status(
+        &self,
+        _pod_state: &mut PodState,
+        _pod: &Pod,
+    ) -> anyhow::Result<serde_json::Value> {
+        make_status(Phase::Pending, "CrashLoopBackoff")
+    }
+}
+
+impl EdgeTo<Registered> for CrashLoopBackoff {}
```
itowlson, line 30: I realise this may be one of those irreconcilable fundamental philosophical differences, but I want to make the case that the `EdgeTo` trait is not buying us anything in this particular approach to state machines.

The purpose of `EdgeTo` is for the compiler to prevent code making transitions that are not permitted by the state machine.  This is super useful when exposing a state machine where external code will drive the transitions.  If we are building an IoT coffee machine that users can program to make themselves the perfect brew, then we certainly don't want some unfortunate soul to put it into the Squirting state when it's not in the FullOfBrownStuff state.  (Look, don't @ me about the names, I don't know how coffee works.)

But that _isn't_ how we're doing it here.  In our approach, each state 'drives itself.'  The state machine switches based solely on code it controls.  _There's no other code we need to defend against._  The `EdgeTo` feature only defends the state against itself.

Meanwhile, there appears to be a downside to `EdgeTo` in which the `Transition::next` function ends up with a parameter that is _always_ ignored and only exists to carry type information a la phantom types.  We're cluttering the interface at the point of use, as well as requiring the busywork of declaring an edge as well as using it.

I realise that the 'defends the code against itself' feature does have merit; it's why we declare and type local variables.  But in this case I don't feel that explicitly specifying the subset of permissible transitions is worth the candle.  If it were driven from the outside, absolutely; but not when it's doing no more than checking the return value.

Thanks for coming to my TED talk.

itowlson (changes requested):
LET THE PHILOSOPHISING COMMENCE

crates/kubelet/src/state.rs
```diff
@@ -1,25 +1,221 @@
...
+    /// ```
+    pub fn next<I: State<PodState>, S: State<PodState>>(_i: &I, s: S) -> Transition<PodState>
+    where
+        I: EdgeTo<S>,
+    {
+        Transition::Next(StateHolder { state: Box::new(s) })
+    }
+}
+
+impl<PodState> std::fmt::Display for Transition<PodState> {
```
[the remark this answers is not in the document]
kflansburg: I went with `Display` here because I did not want to include the contents of `PodState` (or require that it is `Debug`).

crates/kubelet/src/state.rs
```diff
@@ -68,15 +264,10 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
 
         let transition = { state.next(pod_state, &pod).await? };
 
-        debug!(
-            "Pod {} state execution result: {:?}",
-            pod.name(),
-            transition
-        );
+        debug!("Pod {} state execution result: {}", pod.name(), transition);
```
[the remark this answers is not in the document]
kflansburg: Will do.

crates/kubelet/src/state.rs
```diff
@@ -88,11 +279,7 @@ pub struct Stub;
 
 #[async_trait::async_trait]
 impl<P: 'static + Sync + Send> State<P> for Stub {
```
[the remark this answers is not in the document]
kflansburg: I'm not sure, it is primarily used for doctests, I dont think they can import from the test module, since it is behind `cfg(test)`

crates/kubelet/src/state.rs
kflansburg: So, the real tests are up in the doctest for for `Transition`, where we are actually able to flag a test as `compile_fail` and confirm that the compiler is checking these things. This test mostly serves as a sanity check that the `compile_fail` isnt due to something else (change in API). I think this test is mostly superfluous, as `Stub`, as well as some of the doctests test roughly the same thing. I just kept it around because we may want to expand this test module in the future and it didnt hurt.

crates/wascc-provider/src/states/crash_loop_backoff.rs
kflansburg: I agree that this is mostly a weak constraint. People are able to simply add edges in order to get their code to compile without considering the consequences, so for some developers we are just asking them to do extra typing (this actually feels a lot like `String` and `&str` for newbie Rust developers). I would also love to find a way to remove the unused parameter from `Transition::next`. But I do think that this gives people the option (and encourages them) to think about what transitions they are defining. It helps maintainers be confident that pull requests aren't introducing new state transitions without explicitly stating that. Personally, I think good Rust code is fairly verbose and explicit, so this matches my expectation of how one would define a state machine. You mention not needing to defend against outside code, and I disagree, having multiple authors contribute code (or even the same other on a different day) results in situations where they don't fully consider the behavior or intent of the rest of the code, and having the compiler there to double check things is one of the things I love about Rust. 

I'm happy to iterate on this API but in general I think there should be a limited set of valid outgoing edges associated with a state that is not simply every other state in the graph.

kflansburg: I'll just add that I think there definitely are invalid state transitions in the graph, and if we can enforce them at compile time we should. Beyond that I think its just a question of ergonomics.

thomastaylor312: Side note: could the `next` function use a `PhantomData`?

thomastaylor312: @kflansburg I am not sure I understand your counterexample. Could you elaborate more? I don't see how `EdgeTo` would protect people contributing code on the same day in the way you explain. For example if both of us added a new state that transitions to starting, that would be caught by the git diff and probably the compiler as I'd probably be returning an incorrect type.

Either way, this is where I agree with @itowlson. If it is only protecting us from very specific edge cases, it does feel like the ceremony to benefit ratio is a little off, especially how adding a new state probably isn't something that happens very often.

kflansburg: Yes, I think `PhantomData` would work here. 

I guess my point is that there are definitely invalid state transitions. For instance `Completed` to `Registered`. As it stands without edge enforcement, you could accidentally call `Transition::next(Registered)`, and you would be relying on your own understanding of the state machine and the code reviewers being vigilant to note that this is a transition which is invalid and has not been considered. I think this is a fairly real possibility, just based on our confusion over state names in the Wasi pull request. If we can make the compiler fail on this, requiring the developer and reviewers to explicitly add the edge, I think we should, and I dont think it is much more verbose (I actually think its helpful having a nice list of outgoing edges in the file). Without this enforcement I think we are leaving a lot of the benefits of a state machine on the table. 

At the very least, I would like to request that this is a feature that Providers can enable so that I can use it in my own CRI provider.

thomastaylor312:
@kflansburg So after talking some more offline with @itowlson, I think we should just go forward with this as is. We still have quite a bit of tweaking to do with the API surface anyway as we learn more about it in practice. However, we think it might be a good idea to rename the trait to `CanTransitionTo` as it lines up with our idea of `Transition` elsewhere, whereas now we don't even mention the term "edge" except in this trait

kflansburg:
@thomastaylor312 I'm definitely in favor of renaming `EdgeTo`. I spent some time last night trying to get `PhantomData` working and/or the State owned by the state handlers, but I think both are impossible unless we can eliminated the `Box<dyn>`.

thomastaylor312:
Ok we can keep iterating on that as needed

crates/kubelet/src/state.rs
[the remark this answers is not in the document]
itowlson: If it's used for doctests can't it be in the documentation code?  It feels super awkward having it in the public API.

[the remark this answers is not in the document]
kflansburg: Well, it is part of examples. So it’s intended to be public but there are no other states defined in this crate we can use. Perhaps a different name would make more sense?

kflansburg:
This should be ready to merge.

crates/wasi-provider/src/states/starting.rs
```diff
@@ -98,10 +98,10 @@ impl Starting {
 #[async_trait::async_trait]
 impl State<PodState> for Starting {
     async fn next(
-        &self,
+        self: Box<Self>,
```
thomastaylor312, line 101: Nit: We can probably change the logic in this function so it doesn't need an Arc Mutex anymore (as we are consuming `self`). However, I can do that in a follow up PR

thomastaylor312 (approved):
This looks great! TIL about `self: Box<Self>`, which I had no idea was possible

itowlson (approved):
LGTM except for the Stub thing but we can tackle that in a subsequent PR - for now let's get this in.  Thanks @kflansburg - the state machine stuff is getting better all the time!

PR #388, opened 2020-09-16
ref(*): Updates state trait to make next have mut self

thomastaylor312:
This facilitates passing of non-shared data from state to state. A good
example of this is Initializing -> Starting in the wasi provider

Also includes a small doc change to account for the new provider interface

### DIFF
--- a/crates/wasi-provider/src/states/starting.rs
+++ b/crates/wasi-provider/src/states/starting.rs
@@ -1,10 +1,8 @@
 use std::collections::HashMap;
 use std::ops::Deref;
 use std::path::PathBuf;
-use std::sync::Arc;
 
 use log::{debug, info};
-use tokio::sync::Mutex;
 
 use kubelet::container::{Container, ContainerKey};
 use kubelet::pod::{key_from_pod, Handle};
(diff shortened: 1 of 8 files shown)


crates/wasi-provider/src/states/starting.rs
```diff
@@ -84,30 +82,24 @@ pub(crate) type ContainerHandleMap =
...
-    init_handles: Arc<Mutex<ContainerHandleMap>>,
+    init_handles: ContainerHandleMap,
 }
 
 impl Starting {
-    pub(crate) fn new(init_handles: ContainerHandleMap) -> Self {
-        Starting {
-            init_handles: Arc::new(Mutex::new(init_handles)),
-        }
+    pub fn new(init_handles: ContainerHandleMap) -> Self {
```
bacongobbler, line 89: I'm curious on the motivation for the visibility change. Is there ever a time where someone would need to start from a Starting state? the Starting struct still has `pub(crate)` visibility which makes me think this was intentional

kflansburg:
I guess I'm not convinced that this isn't better handled by keeping states immutable and storing this information in PodState. Why should container handles not be stored in PodState?.

bacongobbler:
So the reason why the init container handles need to be passed along from the initializing state to the starting state was due to some wonky design issues with the pod handler. To pass them along, we had to create a pod handler and add those container handles to the pod handle. Then, when we came to the starting state, we'd have to re-fetch that pod handle to add more container handles to the container handler (the main run containers).

I'm not sure if there's a way we can address that more cleanly, but i we can, then there might not be a need to make states mutable and just use the shared pod state to share state (as originally intended).

crates/wasi-provider/src/states/starting.rs
thomastaylor312: It should be `pub(crate)` and I just forgot after deleting and re-adding

thomastaylor312:
@kflansburg I think there are plenty of use cases (speaking in the generic sense) where you might want to pass information to a state, but not _every_ state (or other running states, a la handles). Trying to put a bunch of `Options` all over in a shared struct feels like a more clunky way of doing things than being able to pass information like in the `Error` state or in `Starting`. Basically, if something is truly shared by 2 or more states, or by things outside of the state machine, then a shared state makes sense. Otherwise, this gives implementors the option, but not the requirement, to pass along other data to a single state only needed for decoration, messaging, etc.

thomastaylor312:
Please note: I am ok if we choose not to go this route. It just seemed like a decent possible solution

kflansburg:
It seems to me that PodState can simply have a single `container_handles` that is initially given init container handles and then application container handles as it runs through the state. Before those states it is empty and after it is full and does not change, but I don't think it is a bad thing to have that available but unused to all states. To me states themselves are immutable and do not carry information other than some context on how the state was entered (as we do with `Error`).

thomastaylor312:
I'm gonna go ahead and close this for now. I still think we have some sort of a rough edge here, but I'd like to see how this evolves in practice before changing anything

PR #392, opened 2020-09-22
Exponential backoff

### LINKED ISSUE #386: Implement exponential backoff for `*BackOff` states
Right now we just wait 60 seconds. We should actually be increasing that time using an exponential backoff

itowlson:
Fixes #386.

### DIFF
--- a/crates/kubelet/src/backoff/mod.rs
+++ b/crates/kubelet/src/backoff/mod.rs
@@ -23,16 +23,18 @@ pub struct ExponentialBackoffStrategy {
     last_duration: Duration,
 }
 
-impl ExponentialBackoffStrategy {
+impl Default for ExponentialBackoffStrategy {
     /// Gets a backoff strategy that adheres to the Kubernetes defaults.
-    pub fn kubernetes_default() -> Self {
+    fn default() -> Self {
         Self {
             base_duration: Duration::from_secs(10),
             cap: Duration::from_secs(300),
             last_duration: Duration::from_secs(0),
         }
     }
+}
 
+impl ExponentialBackoffStrategy {
     fn capped_next_duration(&self) -> Duration {
         let next_duration = if self.last_duration == Duration::from_secs(0) {
             self.base_duration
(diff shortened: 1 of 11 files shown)


crates/kubelet/src/backoff/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+/// until hitting a cap.
+pub struct ExponentialBackoffStrategy {
+    base_duration: Duration,
+    cap: Duration,
+    last_duration: Duration,
+}
+
+impl ExponentialBackoffStrategy {
+    /// Gets a backoff strategy that adheres to the Kubernetes defaults.
+    pub fn kubernetes_default() -> Self {
```
thomastaylor312, line 28: Should we just make this a `Default` implementation instead or do we intend for this to be used in other non-krustlet contexts?

crates/kubelet/src/backoff/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+            cap: Duration::from_secs(300),
+            last_duration: Duration::from_secs(0),
+        }
+    }
+
+    fn capped_next_duration(&self) -> Duration {
+        let next_duration = if self.last_duration == Duration::from_secs(0) {
+            self.base_duration
+        } else {
+            self.last_duration * 2
```
thomastaylor312, line 40: Most exponential backoff/retry algorithms I've seen have a jitter component. Should we do that here?

thomastaylor312 (commented):
This is looking good. Just a few questions

crates/kubelet/src/backoff/mod.rs
itowlson: I kind of wanted to explicitly make it clear that this was the default _for the context of Kubernetes_.  But you're right.  There _is_ no context other than Kubernetes, so this can just be `default`.

crates/kubelet/src/backoff/mod.rs
itowlson: I wondered about that. The description of backoff in the k8s docs didn't mention jitter, but I can't imagine anything relies on there _not_ being jitter, and it may indeed be desirable!

thomastaylor312: Honestly up to you. I won't block on it at all. I just know from looking into exponential backoff and the logic behind it that jitter helps avoid problems on retries

thomastaylor312 (approved):

PR #396, opened 2020-09-23
ref(*): Removes use of deprecated `Informer` from the `kube` crate

thomastaylor312:
This replaces all uses of the `Informer` with the new `watcher` type from the `kube_runtime` crate. I also fixed a bunch of clippy lints I found along the way

This also contains a refactor around pod keys by introducing the new `PodKey` type. This serves 2 main purposes. The first is it makes it clear that we are using key for a pod rather than it just being a `String`. The second is that instead of it being an arbitrary string, it now has separate fields that define the key. Arbitrary strings and separators (even with an alias type) can come back to bite, so this makes it a concrete type with separate fields. As part of this, I have deprecated, but not removed, the old pod key methods

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: engineerd/setup-kind@v0.4.0
+        with:
+          version: "v0.9.0"
       - uses: engineerd/configurator@v0.0.1
         with:
           name: just
@@ -86,8 +88,8 @@ jobs:
           KUBECONFIG=${CONFIG_DIR}/kubeconfig-wascc ./target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 --bootstrap-file ${CONFIG_DIR}/bootstrap.conf --cert-file ./config/krustlet-wascc.crt --private-key-file ./config/krustlet-wascc.key &
           # Wait for things to start before approving certs and then delete so we don't overlap with the same hostname
           sleep 5 && kubectl certificate approve $(hostname)-tls
-          sleep 2 && kubectl delete csr $(hostname)-tls 
+          sleep 2 && kubectl delete csr $(hostname)-tls
           KUBECONFIG=${CONFIG_DIR}/kubeconfig-wasi ./target/debug/krustlet-wasi --node-name krustlet-wasi --port 3001 --bootstrap-file ${CONFIG_DIR}/bootstrap.conf --cert-file ./config/krustlet-wasi.crt --private-key-file ./config/krustlet-wasi.key &
           sleep 5 && kubectl certificate approve $(hostname)-tls
-          sleep 2 && kubectl delete csr $(hostname)-tls 
+          sleep 2 && kubectl delete csr $(hostname)-tls
           just test-e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
           KUBECONFIG=${CONFIG_DIR}/kubeconfig-wascc ./target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 --bootstrap-file ${CONFIG_DIR}/bootstrap.conf --cert-file ./config/krustlet-wascc.crt --private-key-file ./config/krustlet-wascc.key &
           # Wait for things to start before approving certs and then delete so we don't overlap with the same hostname
           sleep 5 && kubectl certificate approve $(hostname)-tls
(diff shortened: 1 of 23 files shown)


crates/kubelet/src/pod/queue.rs
```diff
@@ -199,8 +134,85 @@ impl<P: 'static + Provider + Sync + Send> Queue<P> {
                 }
                 Ok(())
             }
-            WatchEvent::Bookmark(_) => Ok(()),
-            WatchEvent::Error(e) => Err(e.clone().into()),
+            // TODO: Figure out how to handle restarted. It might need to be handled at a higher level (as to be reenqueued properly)
```
thomastaylor312, line 137: Just realized I forgot to work on this. I'll handle this tomorrow

thomastaylor312: This is all done

thomastaylor312:
This is now ready for review

tests/test_resource_manager.rs
```diff
@@ -93,7 +93,7 @@ impl TestResourceManager {
 
         // k8s seems to need a bit of time for namespace permissions to flow
         // through the system.  TODO: make this less worse
-        tokio::time::delay_for(tokio::time::Duration::from_millis(100)).await;
+        tokio::time::delay_for(tokio::time::Duration::from_millis(1000)).await;
```
thomastaylor312, line 96: Note: This is just a temporary fix. Given the number of namespaces we are creating, it often takes the token controller more than 100ms to create all the default tokens. I will create an issue to fix this with something more robust

.github/workflows/ci.yml
```diff
@@ -68,6 +68,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: engineerd/setup-kind@v0.4.0
+        with:
+          version: "v0.9.0"
```
bacongobbler, line 72: is this because setup-kind uses an older version of kind, or that we have to downgrade?

bacongobbler (approved):
LGTM!

.github/workflows/ci.yml
thomastaylor312: Yeah, it uses an older version, so this was purely to upgrade the version

tests/test_resource_manager.rs
thomastaylor312: https://github.com/deislabs/krustlet/issues/403

.github/workflows/ci.yml
bacongobbler: FYI: https://github.com/engineerd/setup-kind/pull/26

.github/workflows/ci.yml
```diff
@@ -86,8 +88,8 @@ jobs:
           KUBECONFIG=${CONFIG_DIR}/kubeconfig-wascc ./target/debug/krustlet-wascc --node-name krustlet-wascc --port 3000 --bootstrap-file ${CONFIG_DIR}/bootstrap.conf --cert-file ./config/krustlet-wascc.crt --private-key-file ./config/krustlet-wascc.key &
           # Wait for things to start before approving certs and then delete so we don't overlap with the same hostname
           sleep 5 && kubectl certificate approve $(hostname)-tls
-          sleep 2 && kubectl delete csr $(hostname)-tls 
+          sleep 2 && kubectl delete csr $(hostname)-tls && sleep 2
```
thomastaylor312, line 91: This was to solve a race condition where the delete hadn't completed (I'm not sure if `--wait` works with this either). This will be going away next week when I convert this over to our one click testing

thomastaylor312:
Ok, @kflansburg since you asked to review this as well this evening, feel free to merge if you approve it

kflansburg (approved):
LGTM, added a couple of comments, mostly on things that are still TODO in the `Pod::queue` module. Let me know what you think and I can open some issues.

crates/kubelet/src/pod/queue.rs
```diff
@@ -190,17 +124,122 @@ impl<P: 'static + Provider + Sync + Send> Queue<P> {
...
+            debug!("Pod {} terminated. Jumping to state {:?}.", name, state);
+            match run_to_completion(&task_client, state, &mut pod_state, &pod).await {
+                Ok(()) => debug!("Pod {} state machine exited without error", name),
+                Err(e) => error!("Pod {} state machine exited with error: {:?}", name, e),
+            }
+        }
+    }
+
+    debug!("Pod {} waiting for deregistration.", name);
+    loop {
```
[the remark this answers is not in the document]
thomastaylor312: I think that would be a great improvement

PR #398, opened 2020-09-24
Infallible state handlers

### LINKED ISSUE #390: Make State Handlers Infallible
Instead of returning `Result<Transition>`, return `Transition` to prevent unhandled errors in state handlers 

As discussed in #389 .

itowlson:
Fixes #390.

There's currently more boilerplate around error handling than is nice.  I looked at some macro-based shortcuts for these, but haven't yet found anything that looks nice.  I'll keep tinkering though!

### DIFF
--- a/crates/kubelet/src/state.rs
+++ b/crates/kubelet/src/state.rs
@@ -1,5 +1,5 @@
 //! Used to define a state machine of Pod states.
-use log::{debug, error};
+use log::debug;
 
 pub mod prelude;
 
@@ -30,10 +30,6 @@ pub enum Transition<PodState> {
     Next(StateHolder<PodState>),
     /// Stop executing the state machine and report the result of the execution.
     Complete(anyhow::Result<()>),
-    /// Stop executing the state machine and report that the state machine
-    /// failed.
-    // TODO: is this really any different from Complete(Err(...))?
-    Fatal(anyhow::Error),
 }
 
 /// Mark an edge exists between two states.
@@ -266,10 +262,6 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
                     result
                 );
                 break result;
-            }
-            Transition::Fatal(err) => {
-                error!("Pod {} execution fatal error {:?}", pod.name(), err);
-                break Err(err);
             }
         };
     }
--- a/crates/kubelet/src/state.rs
+++ b/crates/kubelet/src/state.rs
@@ -1,5 +1,5 @@
 //! Used to define a state machine of Pod states.
-use log::debug;
+use log::{debug, error};
 
 pub mod prelude;
 
@@ -28,8 +28,12 @@ pub struct StateHolder<PodState> {
 pub enum Transition<PodState> {
     /// Transition to new state.
     Next(StateHolder<PodState>),
-    /// This is a terminal node of the state graph.
+    /// Stop executing the state machine and report the result of the execution.
     Complete(anyhow::Result<()>),
+    /// Stop executing the state machine and report that the state machine
+    /// failed.
+    // TODO: is this really any different from Complete(Err(...))?
+    Fatal(anyhow::Error),
 }
 
 /// Mark an edge exists between two states.
@@ -58,8 +62,8 @@ impl<PodState> Transition<PodState> {
     ///         self: Box<Self>,
     ///         _pod_state: &mut PodState,
     ///         _pod: &Pod,
-    ///     ) -> anyhow::Result<Transition<PodState>> {
-    ///         Ok(Transition::next(self, TestState))
+    ///     ) -> Transition<PodState> {
+    ///         Transition::next(self, TestState)
     ///     }
     ///
     ///     async fn json_status(
@@ -217,11 +221,7 @@ pub trait AsyncDrop: Sized {
 /// A trait representing a node in the state graph.
 pub trait State<PodState>: Sync + Send + 'static + std::fmt::Debug {
     /// Provider supplies method to be executed when in this state.
-    async fn next(
-        self: Box<Self>,
-        pod_state: &mut PodState,
-        pod: &Pod,
-    ) -> anyhow::Result<Transition<PodState>>;
+    async fn next(self: Box<Self>, pod_state: &mut PodState, pod: &Pod) -> Transition<PodState>;
 
     /// Provider supplies JSON status patch to apply when entering this state.
     async fn json_status(
@@ -252,7 +252,7 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
             .await?;
         debug!("Pod {} executing state handler {:?}", pod.name(), state);
 
-        let transition = { state.next(pod_state, &pod).await? };
+        let transition = { state.next(pod_state, &pod).await };
 
         state = match transition {
             Transition::Next(s) => {
@@ -267,6 +267,10 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
                 );
                 break result;
             }
+            Transition::Fatal(err) => {
+                error!("Pod {} execution fatal error {:?}", pod.name(), err);
+                break Err(err);
+            }
         };
     }
 }
@@ -277,12 +281,8 @@ pub struct Stub;
 
 #[async_trait::async_trait]
 impl<P: 'static + Sync + Send> State<P> for Stub {
-    async fn next(
-        self: Box<Self>,
-        _pod_state: &mut P,
-        _pod: &Pod,
-    ) -> anyhow::Result<Transition<P>> {
-        Ok(Transition::Complete(Ok(())))
+    async fn next(self: Box<Self>, _pod_state: &mut P, _pod: &Pod) -> Transition<P> {
+        Transition::Complete(Ok(()))
     }
 
     async fn json_status(
@@ -311,8 +311,8 @@ mod test {
             self: Box<Self>,
             _pod_state: &mut PodState,
             _pod: &Pod,
-        ) -> anyhow::Result<Transition<PodState>> {
-            Ok(Transition::Complete(Ok(())))
+        ) -> Transition<PodState> {
+            Transition::Complete(Ok(()))
         }
 
         async fn json_status(
@@ -337,8 +337,8 @@ mod test {
                 self: Box<Self>,
                 _pod_state: &mut PodState,
                 _pod: &Pod,
-            ) -> anyhow::Result<Transition<PodState>> {
-                Ok(Transition::next(self, ValidState))
+            ) -> Transition<PodState> {
+                Transition::next(self, ValidState)
             }
 
             async fn json_status(
--- a/crates/wascc-provider/src/states.rs
+++ b/crates/wascc-provider/src/states.rs
@@ -30,6 +30,6 @@ macro_rules! fail_fatal {
     ($err:ident) => {{
         let aerr = anyhow::Error::from($err);
         log::error!("{:?}", aerr);
-        return Transition::Fatal(aerr);
+        return Transition::Complete(Err(aerr));
     }};
 }
--- a/crates/wascc-provider/src/states.rs
+++ b/crates/wascc-provider/src/states.rs
@@ -8,3 +8,28 @@ pub(crate) mod running;
 pub(crate) mod starting;
 pub(crate) mod terminated;
 pub(crate) mod volume_mount;
+
+/// When called in a state's `next` function, exits the current state
+/// and transitions to the Error state.
+#[macro_export]
+macro_rules! transition_to_error {
+    ($slf:ident, $err:ident) => {{
+        let aerr = anyhow::Error::from($err);
+        log::error!("{:?}", aerr);
+        let error_state = super::error::Error {
+            message: aerr.to_string(),
+        };
+        return Transition::next($slf, error_state);
+    }};
+}
+
+/// When called in a state's `next` function, exits the state machine
+/// returns a fatal error to the kubelet.
+#[macro_export]
+macro_rules! fail_fatal {
+    ($err:ident) => {{
+        let aerr = anyhow::Error::from($err);
+        log::error!("{:?}", aerr);
+        return Transition::Fatal(aerr);
+    }};
+}
--- a/crates/wascc-provider/src/states/starting.rs
+++ b/crates/wascc-provider/src/states/starting.rs
@@ -142,11 +142,7 @@ pub struct Starting;
 
 #[async_trait::async_trait]
 impl State<PodState> for Starting {
-    async fn next(
-        self: Box<Self>,
-        pod_state: &mut PodState,
-        pod: &Pod,
-    ) -> anyhow::Result<Transition<PodState>> {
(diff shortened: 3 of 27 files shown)


kflansburg:
So, I didn’t want to overload the provider trait with too many associated types, but we could have them define an error state and transition to that when a state handler returns an error. The problem I see with that is that you don’t always want to return to the same error state (such as when the pod is already exiting and does not need to retry)

crates/kubelet/src/state.rs
```diff
@@ -28,8 +28,12 @@ pub struct StateHolder<PodState> {
 pub enum Transition<PodState> {
     /// Transition to new state.
     Next(StateHolder<PodState>),
-    /// This is a terminal node of the state graph.
+    /// Stop executing the state machine and report the result of the execution.
     Complete(anyhow::Result<()>),
+    /// Stop executing the state machine and report that the state machine
+    /// failed.
+    // TODO: is this really any different from Complete(Err(...))?
+    Fatal(anyhow::Error),
```
kflansburg, line 36: I tend to agree that `Complete(Err())` could be used for this.

crates/kubelet/src/state.rs
```diff
@@ -58,8 +62,8 @@ impl<PodState> Transition<PodState> {
     ///         self: Box<Self>,
     ///         _pod_state: &mut PodState,
     ///         _pod: &Pod,
-    ///     ) -> anyhow::Result<Transition<PodState>> {
-    ///         Ok(Transition::next(self, TestState))
+    ///     ) -> Transition<PodState> {
```
kflansburg, line 65: One of the drawbacks to our `compile_fail` doctests are that they need to be chanced here too, but pass because they still fail to compile with this change. I wish you could match on a particular compile error.

crates/wascc-provider/src/states.rs
```diff
@@ -8,3 +8,28 @@ pub(crate) mod running;
 pub(crate) mod starting;
 pub(crate) mod terminated;
 pub(crate) mod volume_mount;
+
+/// When called in a state's `next` function, exits the current state
+/// and transitions to the Error state.
+#[macro_export]
+macro_rules! transition_to_error {
```
kflansburg, line 15: This looks pretty good, maybe a shorter name. 
Could we also introduce something to unwrap `Option` and `Result` easily?

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -118,20 +118,25 @@ impl State<PodState> for Initializing {
...
-                        return Ok(Transition::next(self, Error { message }));
+
+                        let status_json = match serde_json::to_vec(&s) {
+                            Ok(json) => json,
+                            Err(e) => fail_fatal!(e),
+                        };
+
+                        match client
+                            .patch_status(pod.name(), &PatchParams::default(), status_json)
+                            .await
```
kflansburg, line 135: Should this fail? Or just be logged. A control plane outage doesnt necessarily need to stop the Pod.

crates/wasi-provider/src/states/starting.rs
```diff
@@ -139,3 +140,4 @@ impl State<PodState> for Starting {
 }
 
 impl TransitionTo<Running> for Starting {}
+impl TransitionTo<Error> for Starting {}
```
kflansburg, line 143: I like that we had to explicitly define that this state can result in an error, and `Error` is the state we want to handle that.

krustlet-wascc-e2e.stderr.txt
```diff
@@ -0,0 +1,45 @@
+[2020-09-23T20:10:21Z INFO  wascc_host::bus::inproc] Initialized Message Bus (internal)
```
kflansburg, line 1: This seems like it should be added to `.gitignore`.

kflansburg (approved):
I think it looks good. I think it kicked off a good discussion of some macros that we could introduce and include in the `prelude`, but I dont want to delay merging while we work on that.

crates/wascc-provider/src/states.rs
kflansburg: https://github.com/kflansburg/krustlet/commit/b598697c31c5c1628df79372c0212171d20889d6

kflansburg: Passing in `self` is becoming very tiresome :(

itowlson: I started with macros that wrapped `Result` so you could write `let handle = fatal_if_failed!(start_container(...))` but I didn't like it because it buried the lede especially in the error case where you had to write `let foo = error_if_failed!(self, some_complex_expression(...))` and the _thing you were actually doing on the line_ got buried in noise.

I looked at the Try trait so that we could have the modifier in suffix position but that didn't work...

crates/kubelet/src/state.rs
thomastaylor312: I second

crates/wascc-provider/src/states.rs
```diff
@@ -8,3 +8,28 @@ pub(crate) mod running;
...
+    ($slf:ident, $err:ident) => {{
+        let aerr = anyhow::Error::from($err);
+        log::error!("{:?}", aerr);
+        let error_state = super::error::Error {
+            message: aerr.to_string(),
+        };
+        return Transition::next($slf, error_state);
```
thomastaylor312, line 22: Curiosity question here: Can we not just do `return Transition::next(self, error_state)`? This should only be used inside of a state function, so it should know what `self` is already, right?

crates/wascc-provider/src/states/starting.rs
```diff
@@ -176,37 +167,15 @@ impl State<PodState> for Starting {
...
-                    Err(e) => {
-                        // TODO: identify if any of the start_container errors are fatal
-                        // enough that we should return an Err and exit the run loop
-                        error!("{:?}", e);
-                        let error_state = Error {
-                            message: e.to_string(),
-                        };
-                        return Ok(Transition::next(self, error_state));
-                    }
+                    Err(e) => fail_fatal!(e),
```
thomastaylor312, line 170: I think a fatal failure here is probably not correct. In waSCC, there could be a problem with the host or lattice or a capability is not loaded. That problem could be transient or resolved by the time we come from a backoff, so this should probably just transition to an error state

crates/wascc-provider/src/states/terminated.rs
```diff
@@ -7,16 +7,15 @@ pub struct Terminated;
...
-        pod_state: &mut PodState,
-        _pod: &Pod,
-    ) -> anyhow::Result<Transition<PodState>> {
+    async fn next(self: Box<Self>, pod_state: &mut PodState, _pod: &Pod) -> Transition<PodState> {
         let mut lock = pod_state.shared.handles.write().await;
         if let Some(handle) = lock.get_mut(&pod_state.key) {
-            handle.stop().await?;
+            let stop_result = handle.stop().await;
+            if let Err(e) = stop_result {
+                return Transition::Fatal(e);
```
thomastaylor312, line 15: Nit: I _think_ fatal works here, but I also wonder if we should transition back into `Terminated` again to retry. Trying to think of all the cases where an error would occur here and if it is recoverable (i.e. if stop fails, will it succeed if we try again?)

crates/wasi-provider/src/states/initializing.rs
```diff
@@ -87,7 +84,10 @@ impl State<PodState> for Initializing {
             // Each new init container resets the CrashLoopBackoff timer.
             pod_state.crash_loop_backoff_strategy.reset();
 
-            let handle = start_container(pod_state, pod, &init_container).await?;
+            let handle = match start_container(pod_state, pod, &init_container).await {
+                Ok(h) => h,
+                Err(e) => fail_fatal!(e),
```
thomastaylor312, line 89: Same note here as with the wascc provider. I am not sure if this is a fatal failure, especially from a k8s perspective. Even if a docker container is totally misconfigured, it keeps looping over it again after backoff

crates/wasi-provider/src/states/initializing.rs
thomastaylor312: Yeah, I don't think this should be a fatal fail. Definitely log an error though

crates/wasi-provider/src/states/starting.rs
```diff
@@ -111,22 +109,25 @@ impl State<PodState> for Starting {
 
         info!("Starting containers for pod {:?}", pod.name());
         for container in pod.containers() {
-            let container_handle = start_container(pod_state, &pod, &container).await?;
+            let container_handle = match start_container(pod_state, &pod, &container).await {
+                Ok(h) => h,
+                Err(e) => fail_fatal!(e),
```
thomastaylor312, line 114: Same here re: fatal

crates/wasi-provider/src/states/terminated.rs
```diff
@@ -7,16 +7,15 @@ pub struct Terminated;
...
-        pod_state: &mut PodState,
-        _pod: &Pod,
-    ) -> anyhow::Result<Transition<PodState>> {
+    async fn next(self: Box<Self>, pod_state: &mut PodState, _pod: &Pod) -> Transition<PodState> {
         let mut lock = pod_state.shared.handles.write().await;
         if let Some(handle) = lock.get_mut(&pod_state.key) {
-            handle.stop().await?;
+            let stop_result = handle.stop().await;
+            if let Err(e) = stop_result {
+                return Transition::Fatal(e);
```
thomastaylor312, line 15: Same nit here about whether fatal is correct or not

krustlet-wascc-e2e.stderr.txt
thomastaylor312: Yeah, we should add these output files to gitignore for sure

thomastaylor312 (changes requested):
Mostly some discussion items we should resolve before continuing. This looks really good and the questions are more around whether or not some things should be fatal

crates/kubelet/src/state.rs
itowlson: Yep, done

crates/kubelet/src/state.rs
itowlson: Good catch - thanks!

crates/wasi-provider/src/states/initializing.rs
itowlson: This was replicating previous behaviour where the `?` bubbled up to exit the run loop.  I can change it though.

krustlet-wascc-e2e.stderr.txt
itowlson: Yep, sorry about this!

crates/wascc-provider/src/states.rs
itowlson: Unfortunately not, because Rust macros are hygienic - identifiers in the macro body are don't "join up" with identifiers around the calling site.

https://stackoverflow.com/questions/44120455/how-to-call-methods-on-self-in-macros

https://stackoverflow.com/questions/40593165/unable-to-use-self-in-macro-because-the-macro-expansion-ignores-token-self

https://users.rust-lang.org/t/how-to-use-self-or-any-variable-in-macro-generated-function-body/6264

crates/wasi-provider/src/states/starting.rs
itowlson: I originally had this transition to an error state, but found that caused a visible change in behaviour that caused the tests to fail.  I'd like to keep this as primarily a refactoring PR, and create a separate issue to identify and implement better error behaviour.

crates/wascc-provider/src/states/terminated.rs
itowlson: Yeah, I wasn't sure.  But this is the current behaviour and after some unfortunate experiences trying to make more things into state transitions I decided to keep it for now.  It would be good to review all fatal/complete transitions, clarify the expected/desired behaviour, and ensure that the tests are checking for that behaviour.

(I doubt this particular case affected the tests but I definitely had one where changing to an error transition resulted in a pod not being terminated properly!)

crates/wascc-provider/src/states.rs
kflansburg: I had the same question haha.

crates/wascc-provider/src/states/starting.rs
kflansburg: I'm wondering if this could be an opportunity for an alternative state / error message. Like  capability backoff or something.

crates/wascc-provider/src/states/terminated.rs
thomastaylor312: Yeah, that is fine for now. Just a nit as I was reasoning through things

crates/wasi-provider/src/states/initializing.rs
thomastaylor312: I still think this should just be an error transition. I think it doesn't follow the k8s patter to have the pod stop executing entirely if there was a flake with something like file or network access during an init pod.

thomastaylor312 (changes requested):
This is almost ready to go. I am still hesitant to have fatal errors occur when a "container" fails to start (whether that is init or a normal container in either provider). See my comment on one of them for further details

crates/wasi-provider/src/states/initializing.rs
itowlson: @thomastaylor312 I am okay with that, but I want to put it _out of scope for the PR at hand_.  This is a refactoring PR that aims not to (significantly) change observable behaviour.  Changing this to transition to the error state means that pods with failing init containers go into a state where the init container reports itself as terminated but the pod as a whole goes into CrashLoopBackoff.  This currently causes the integration tests to go into an infinite loop waiting for it, and I'm not sure why (possibly it passes the pod ready within 30 seconds test but there isn't a suitable timeout on the completion check).  Obviously I could also fix up the integration tests but it all gets into questions about the correct behaviour and how we test retries and stuff.

So don't get me wrong.  We **should** make this an error transition.  But I feel strongly that we should do so in a feature PR where we have the leisure to discuss the desired behaviour, its ramifications, how we'll test it, and where else we should be making the corresponding changes.  I worry that such a discussion could bog this PR down and I don't want this one to become long-running because I don't want new PRs to have to base themselves on the fallible model.

thomastaylor312: That is fair

crates/wasi-provider/src/states/initializing.rs
bacongobbler: Yeah, @kflansburg and @thomastaylor312 pointed out a bug here where we shouldn't fail. Would you mind changing this so that we only log the error message, @itowlson? I know it's a little more work here, but it could save us a couple of erroneous error cases in the future.

bacongobbler: That being said, the same argument applied [in another comment](https://github.com/deislabs/krustlet/pull/398#discussion_r495621934) could apply here as well. In which case, I'd be okay with that and file a ticket to resolve this in a follow-up.

thomastaylor312 (approved):

Issue #397, opened 2020-09-23
bug: images without a slash in the name should be accepted


bacongobbler:
```
[2020-09-23T21:44:17Z ERROR wasi_provider::states::registered] Failed to parse reference string 'nginx'. Expected at least one slash (/)
```

Names like `nginx` are valid image references.

PR #402, opened 2020-09-25
ref(reference): add custom error types to match docker/distribution

bacongobbler:
This is a starting number of changes to the reference parser. With these changes, the following occurs:

- we have access to concrete string types after parsing
- reference now returns its own custom types. This way, the errors reported from reference.rs are known
- an error is raised if the length exceeds the name total max length

Follow-up PRs will start refactoring the parser to match the one in docker/distribution. That way we can handle cases like #397.


bacongobbler:
For reference (hehe): https://github.com/docker/distribution/blob/master/reference/reference.go

thomastaylor312 (approved):

PR #404, opened 2020-09-27
Address some nits in Pod::queue

kflansburg:
* Dont busy-poll for pod deletion, use `tokio::sync::Notify`.
* Move status patching to single location in `run_to_completion`.
* `run_to_completion` is now infallible.
* Ignore error deleting Pod when it is already deleted. Log other errors.
* Stop sending the Pod initial event twice.

### DIFF
--- a/crates/kubelet/src/state.rs
+++ b/crates/kubelet/src/state.rs
@@ -1,9 +1,9 @@
 //! Used to define a state machine of Pod states.
-use log::debug;
+use log::{debug, error, warn};
 
 pub mod prelude;
 
-use crate::pod::Pod;
+use crate::pod::{Phase, Pod};
 use k8s_openapi::api::core::v1::Pod as KubePod;
 use kube::api::{Api, PatchParams};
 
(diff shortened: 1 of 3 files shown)


crates/kubelet/src/state.rs
```diff
@@ -231,42 +231,84 @@ pub trait State<PodState>: Sync + Send + 'static + std::fmt::Debug {
...
+    }
+}
+
 /// Iteratively evaluate state machine until it returns Complete.
 pub async fn run_to_completion<PodState: Send + Sync + 'static>(
     client: &kube::Client,
     state: impl State<PodState>,
     pod_state: &mut PodState,
     pod: &Pod,
-) -> anyhow::Result<()> {
```
thomastaylor312, line 240: I'm not sure making this infallible gives us much value. We end up having to `unwrap` and we still aren't handling a status patch retry, so it feels like we just moved the problems elsewhere

thomastaylor312 (changes requested):
I like a lot of the cleanup in the queue, but I am not sure making `run_to_completion` infallible is gaining us much

crates/kubelet/src/state.rs
thomastaylor312: However, I did just remember that we are making state handlers infallible in #398, so I think I'm ok with merging this once we merge #398 and rebase this on top

kflansburg: Yes, it's not that I was trying to make this infallible, it's that it will be once that is merged. Also, I felt that consolidating the patch to the one function in this module would make it trivial to add retry. We could go ahead and add that in this pull request if you want.

thomastaylor312: Nope, no problem. Let's land #398 and then rebase and I'll come back and LGTM it

itowlson (approved):

thomastaylor312 (approved):
LGTM pending rebase. e2e is just a flake, so a re-run is in order

thomastaylor312:
There are still a bunch of weird flakes with e2e in the GH actions. I verified the e2e tests pass locally and we'll come back to address the problem with the e2e pipeline

PR #405, opened 2020-09-28
chore(*): Bumps all deps to the latest version

thomastaylor312:
This is just a general hygiene PR to update all dependencies possible to their latest version. There were small code changes to update with the new APIs from some libraries. I was not able to update `hyperx` because our www-authenticate crate depends on a trait from the older version.

I also can't update wasmtime to 0.20 yet due to wascc requiring 0.19. We'll have to stick with it until we split out the crates or wascc upgrades

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,15 +26,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-codec"
+version = "0.3.0"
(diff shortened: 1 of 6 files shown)


bacongobbler (approved):
LGTM. Just needs a rebase to go green

PR #363, opened 2020-08-22
[oci-distribution] Use table-based tests

jlegrone:
This PR refactors the `client` and `reference` modules to use "table-based" tests via a procedural macro, and is related to https://github.com/deislabs/krustlet/pull/349#discussion_r469433569.

This is meant to:
- prevent early failures from masking issues in later test cases
- provide test output that makes it easier to trace failures back to a
   specific test case
- reduce the cost of adding new test cases

`cargo test | sort` before:

```
running 15 tests
test client::test::test_auth ... ok
test client::test::test_fetch_digest ... ok
test client::test::test_pull_image ... ok
test client::test::test_pull_layer ... ok
test client::test::test_pull_manifest ... ok
test client::test::test_to_v2_blob_url ... ok
test client::test::test_to_v2_manifest ... ok
test errors::test::test_deserialize ... ok
test manifest::test::test_manifest ... ok
test reference::test::parse::digest_only ... ok
test reference::test::parse::missing_slash_char ... ok
test reference::test::parse::no_tag_or_digest ... ok
test reference::test::parse::owned_string ... ok
test reference::test::parse::tag_and_digest ... ok
test reference::test::parse::tag_only ... ok
```

`cargo test | sort` after:

```
running 39 tests
test client::test::auth::case_1_no_tag ... ok
test client::test::auth::case_2_tag ... ok
test client::test::auth::case_3_digest ... ok
test client::test::auth::case_4_tag_and_digest ... ok
test client::test::auth::case_5_empty ... ok
test client::test::fetch_digest::case_1_no_tag ... ok
test client::test::fetch_digest::case_2_tag ... ok
test client::test::fetch_digest::case_3_digest ... ok
test client::test::fetch_digest::case_4_tag_and_digest ... ok
test client::test::fetch_digest::case_5_empty ... ok
test client::test::pull_image::case_1_no_tag ... ok
test client::test::pull_image::case_2_tag ... ok
test client::test::pull_image::case_3_digest ... ok
test client::test::pull_image::case_4_tag_and_digest ... ok
test client::test::pull_image::case_5_empty ... ok
test client::test::pull_layer::case_1_no_tag ... ok
test client::test::pull_layer::case_2_tag ... ok
test client::test::pull_layer::case_3_digest ... ok
test client::test::pull_layer::case_4_tag_and_digest ... ok
test client::test::pull_layer::case_5_empty ... ok
test client::test::pull_manifest::case_1_no_tag ... ok
[...]

PR #406, opened 2020-09-28
use test fixtures for parsing references

bacongobbler:
ref: https://github.com/la10736/rstest

This PR asserts the same test cases currently tested, but in a more compact and readable fashion. This makes it easier and more efficient to add additional test cases in the future as we restructure the reference parser.

inspired by #363.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2165,6 +2165,7 @@ dependencies = [
  "hyperx",
  "log",
  "reqwest",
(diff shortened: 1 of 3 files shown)


bacongobbler:
closing in favour of https://github.com/deislabs/krustlet/pull/413, as that contains the test cases from docker/distribution with passing validation.

PR #408, opened 2020-09-28
Use one click tests instead of the manual setup and run

thomastaylor312:


thomastaylor312:
This is ready to go. Looks like dropping it in was as simple as I had hoped thanks to @itowlson's great work on the testing stuff

bacongobbler (approved):
oh this looks wonderful. Love that it would be removing all of this setup.

thomastaylor312:
I'm doing some debugging here still unfortunately. I always forget that GH actions doesn't use the context of this PR to run tests, so I am testing on my fork to make sure this actually works

.github/workflows/build.yml
```diff
@@ -1,101 +0,0 @@
-name: Build and Test
-on:
-  pull_request:
```
thomastaylor312, line 3: This was the exact same as the other workflow file, just with different events. I combined the events and deleted this file

thomastaylor312: And then I remembered that we did this for a reason

thomastaylor312:
This now includes a fix for unintentionally dropped events that was causing issues with the e2e pipeline

bacongobbler (approved):

PR #412, opened 2020-09-28
feat(kubelet): Adds new derive crate for TransitionTo

thomastaylor312:
This is re-exported as an optional feature in the kubelet crate. It will require us to publish the new `kube-derive` crate as well


thomastaylor312:
I still need to write docs for the new crate and add its publishing step to the CI pipeline, so that is why it is marked as WIP. Please feel free to review it from an API and code standpoint

crates/wascc-provider/src/states.rs
```diff
@@ -1,6 +1,5 @@
 pub(crate) mod crash_loop_backoff;
 pub(crate) mod error;
-pub(crate) mod finished;
```
thomastaylor312, line 3: As a note: These finished states in each provider were never used or transitioned to, so I removed them

bacongobbler (commented):
I'm curious on the need to pull this out into a separate crate. As I understand it, a crate should be self-sufficient and provide enough functionality on its own. Are there other crates out there that follow this pattern of `N` and `N-derive`? This is the first time I've seen this pattern before.

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+// TODO: Docs
```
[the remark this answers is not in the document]
thomastaylor312: https://github.com/deislabs/krustlet/pull/412#issuecomment-700336706

thomastaylor312:
@bacongobbler proc macro crates by definition can only publicly export proc macros, so it needs to be its own crate. This pattern of having a "derive" crate follows the exact pattern from Serde, which I borrow the feature pattern from as well. That is why I re-export the derive crate from within the `kubelet` crate so it can appear to be all in one place

crates/kubelet-derive/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "kubelet-derive"
+version = "0.1.0"
+authors = [
```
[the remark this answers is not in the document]
thomastaylor312: Yes, good catch. I did all of these at once so forgot

bacongobbler (approved):
LGTM

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    let input = parse_macro_input!(input as DeriveInput);
+
+    let name = &input.ident;
+    let generics = &input.generics;
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
+    let mut token_stream = TokenStream::new();
+
+    // We need to check if we found at least one attribute. If it isn't set, we need to error
+    let mut found_attr = false;
```
itowlson, line 37: I'd consider splitting this into a find/filter and an expand step a la:

```
let transition_to_attrs = filter_transition_to_attrs(input);
if transition_to_attrs.is_empty() {
  some_helpful_error()
} else {
  impl_tokens_for(transition_to_attrs)
}
```

This makes the high-level flow more self-documenting, and enables you to pull specific logical functions into named functions instead of having the "which inputs are interesting" + "what do I do with the interesting ones" all bound up in one deeply nested block.

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    for attr in input.attrs.into_iter() {
+        if let Meta::List(parsed_attr) = attr.parse_meta().unwrap() {
```
itowlson, line 39: Is the `unwrap` idiomatic?  It won't cause a panic on a malformed attribute (e.g. while typing and the RLS runs the compiler in the background)?

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+                            impl #impl_generics kubelet::state::TransitionTo<#transition_type> for #name #ty_generics #where_clause {}
+                        };
+                        token_stream.extend(TokenStream::from(expanded));
+                    }
+                }
+            }
+        }
+    }
+
+    if !found_attr {
```
itowlson, line 54: Consider flipping the condition so the reader doesn't have to parse the negative.

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        let message = format!(
+            "No `{}` attribute found for `{}`. Please specify at least one type to transition to",
+            ATTRIBUTE_NAME,
+            name.to_string()
+        );
+        TokenStream::from(Error::new(name.span(), message).to_compile_error())
+    } else {
+        token_stream
+    }
+}
```
itowlson, line 64: Does Rust provide a way to unit-test macros?  If so consider some core happy-path tests at least.

itowlson (changes requested):
This is terrific.  But I think the way the macro body tries to do everything in one loop hurts readability.  Unless this restriction is imposed on you by the Rust macro system, I'd strongly recommend splitting it into smaller functions and simplifying the top level, even if it makes you feel like you are going through the list twice!

crates/kubelet-derive/src/lib.rs
thomastaylor312: Based on examples I found it is. But as I break this up, I think I can use a compile error instead

crates/kubelet-derive/src/lib.rs
thomastaylor312: I think it does if I use proc macro 2. I’ll give it a whirl

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+const ATTRIBUTE_NAME: &str = "transition_to";
+
+#[proc_macro_derive(TransitionTo, attributes(transition_to))]
+pub fn derive_transition_to(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+
+    let name = input.ident;
+    let generics = input.generics;
+
+    let mut transitions = get_transitions(input.attrs);
```
itowlson, line 33: No need for this to be mutable here I think?

crates/kubelet-derive/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+    } else if transitions.len() > 1 {
+        let message = format!(
+            "Multiple `{}` attributes found for `{}`. Please specify only one attribute",
+            ATTRIBUTE_NAME,
+            name.to_string()
+        );
+        TokenStream::from(Error::new(name.span(), message).to_compile_error())
+    } else {
+        // We can unwrap here because we already checked length
+        generate_impl(transitions.pop().unwrap(), generics, name)
```
itowlson, line 52: Just index rather than the mutating `pop` perhaps?

itowlson (approved):
LGTM with minor suggestions

crates/kubelet-derive/src/lib.rs
thomastaylor312: It does so I can pop the data out of the vector down the line

crates/kubelet-derive/src/lib.rs
thomastaylor312: It won't let me index because it counts as moving the data, and an index is a shared reference

crates/kubelet-derive/src/lib.rs
thomastaylor312: Looks like proc macro 2 is built for this, but I am going to do it in a follow up

PR #418, opened 2020-10-01
Mark waSCC and WASI Kubelets as NoSchedule

kflansburg:
I believe this is the correct taint, as `NoExecute` applies to Pods that are already running. Probably doesn't hurt to keep `NoExecute` here. Sadly this does  not resolve the `kube-proxy` issue as they seem to apply a wildcard `NoSchedule` toleration.


thomastaylor312 (changes requested):
I think this is the right thing to do. Can you also update all of the docs and demo manifests to reflect this new change?

thomastaylor312:
Oh I think you’ll need to update the e2e tests as well. 

Marking this as breaking so we remember to put something in the release notes

tests/integration_tests.rs
```diff
@@ -81,7 +81,7 @@ async fn verify_wascc_node(node: Node) -> () {
         .expect("node had no taints");
     let taint = taints
         .iter()
-        .find(|t| t.key == "kubernetes.io/arch")
+        .find(|t| (t.key == "kubernetes.io/arch") & (t.effect == "NoExecute"))
```
thomastaylor312, line 84: Any reason you are using bitwise and here over a boolean and?

thomastaylor312 (approved):
This is looking good now. I'd like @bacongobbler or @itowlson to review this before merge as well since it is going to break current users when they upgrade. Want to make sure we are ok with this

bacongobbler (approved):
LGTM! Can you clarify what is the breaking change here though? Looks to me like we're just adding another taint to the nodes. Does that break existing pods if they upgraded Krustlet to 0.6? Trying to determine the severity/impact of the breaking change here. Thanks!

kflansburg:
Yep, as seen with our test cases after my first commit, any pod spec without the NoSchedule toleration will no longer schedule after this change. We should at least note in the release notes that people update their tolerations.

PR #419, opened 2020-10-01
implement more strict manifest checks

bacongobbler:
This will allow us to catch unsupported OCI manifests (like manifest lists) early on in the process before we attempt to parse that data as an OCI manifest.

We'll also exit out early on if the manifest layer content is empty. This is more of a theoretical test case (have you ever heard of a docker image without any layers?), but it's worth catching as an error case to prevent a case where a provider can receive an `ImageData` object with empty content.

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -4,7 +4,7 @@
 //! OCI distribution client in the future.
 
 use crate::errors::*;
-use crate::manifest::OciManifest;
+use crate::manifest::{OciManifest, Versioned, IMAGE_MANIFEST_MEDIA_TYPE};
 use crate::secrets::RegistryAuth;
 use crate::secrets::*;
 use crate::Reference;
@@ -88,6 +88,10 @@ impl Client {
         }
 
         let (manifest, digest) = self.pull_manifest(image).await?;
+
+        if manifest.layers.is_empty() {
+            return Err(anyhow::anyhow!("no layers to pull"));
+        }
 
         let layers = manifest.layers.into_iter().map(|layer| {
             // This avoids moving `self` which is &mut Self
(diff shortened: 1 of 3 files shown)


crates/oci-distribution/src/client.rs
```diff
@@ -89,6 +89,10 @@ impl Client {
 
         let (manifest, digest) = self.pull_manifest(image).await?;
 
+        if manifest.layers.is_empty() {
```
radu-matei, line 92: I think this is the place we should check for the layer's media type to be `WASM_LAYER_MEDIA_TYPE`.
(this, or in `pull_layer`)

bacongobbler: agreed. I was trying to figure out how we could assert this check in a more generic fashion. I was concerned that projects like wasme used a different media type. Some more research suggests I was wrong, so it should be safe to validate against this media type.

https://github.com/solo-io/wasm/blob/68e67db5ccba6b2e67b835b9e53225a992f5462a/tools/wasme/pkg/model/image.go#L44

crates/oci-distribution/src/client.rs
```diff
@@ -89,6 +89,19 @@ impl Client {
...
+            return Err(anyhow::anyhow!("no layers to pull"));
+        }
+
+        for layer in &manifest.layers {
+            if layer.media_type != WASM_LAYER_MEDIA_TYPE {
```
itowlson, line 97: I am vigorously opposed to doing this here.  This crate should be a (moderately) general purpose OCI client - even the kubelet goes to lengths not to be bound to WASM - that is all at the provider level.

I get though that you don't want to go to the trouble of fetching the layers and then have the consumer find that they're the wrong type.  One option would be for the consumer to say what layer types it accepts.  Another would be to break up the flow so that the consumer gets to look at the list of layers and decide whether it wants to proceed, though that may be overkill unless there are other criteria or there is something a consumer can usefully do with individual layers rather than the whole image.

bacongobbler: I completely agree. oras follows a similar approach of the consumer informing oras which media types the client should accept, and validate against those accepted media types. I'm sure we can amend the API to accomodate that workflow. While it may introduce breaking changes to the API, it would allow us to validate incoming content without placing a huge restriction on what content `oci-distribution` can fetch.

bacongobbler: Okay. I managed to pull it out of the function call. The caller now has to pass a `Vec<&str>` of accepted media types to be pulled.

crates/oci-distribution/src/client.rs
```diff
@@ -675,17 +709,27 @@ mod test {
...
+                    &RegistryAuth::Anonymous,
+                    vec![manifest::WASM_LAYER_MEDIA_TYPE],
+                )
                 .await
                 .expect("failed to pull manifest");
 
             assert!(!image_data.layers.is_empty());
             assert!(image_data.digest.is_some());
+
+            // removing layer validation should fail
```
itowlson, line 728: This should be a separate test I think (and there should arguably be a test for if the client does accept certain media types but the image isn't any of those types).

crates/oci-distribution/src/client.rs
```diff
@@ -233,8 +247,27 @@ impl Client {
             reqwest::StatusCode::OK => {
                 let digest = digest_header_value(&res)?;
                 let text = res.text().await?;
+                debug!("validating response: {}", text);
+                let versioned: Versioned = serde_json::from_str(&text).with_context(|| {
```
itowlson, line 251: This change (these three blocks) looks like it could be pulled out into a new `validate_layer` function which would be more intention-revealing, give a clear home to future validation code and reduce the amount the reader has to process here.

itowlson (changes requested):
Pretty much nitpicks but would be nice to address at least the validation breakout one if you have time.

crates/oci-distribution/src/client.rs
bacongobbler: done

itowlson (approved):
Looks great - thanks for being patient with the nitpicks!

bacongobbler:
Any time! Actually your nitpicks helped break up some of the code to make it more portable. Your nitpicking was appreciated. :smile:

PR #421, opened 2020-10-08
Remove Type Length Patch

kflansburg:
Rust `1.47.0` appears to have fixed this issue.


kflansburg:
It appears Actions is not yet installing `1.47.0`. This works locally though.

bacongobbler (approved):

PR #424, opened 2020-10-17
Simplify Container Status Patching

### LINKED ISSUE #410: Make patching container statuses more approachable
Right now, each provider has to manually send patch requests to kubernetes to update their container statuses. This seems ripe for refactoring to make it simpler for each provider to provide status updates.

https://github.com/deislabs/krustlet/blob/0e838ab046b32b70d5da855b59354867e7262f7e/crates/wasi-provider/src/states/running.rs#L45-L56

kflansburg:
Closes #410

### DIFF
--- a/crates/kubelet/src/container/status.rs
+++ b/crates/kubelet/src/container/status.rs
@@ -1,9 +1,12 @@
+use crate::container::Container;
+use crate::pod::Pod;
 use chrono::{DateTime, Utc};
 use k8s_openapi::api::core::v1::{
     ContainerState, ContainerStateRunning, ContainerStateTerminated, ContainerStateWaiting,
-    ContainerStatus as KubeContainerStatus,
+    ContainerStatus as KubeContainerStatus, Pod as KubePod,
 };
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
+use log::{info, warn};
 
 /// Status is a simplified version of the Kubernetes container status
 /// for use in providers. It allows for simple creation of the current status of
@@ -37,7 +40,7 @@ pub enum Status {
 
 impl Status {
     /// Convert the container status to a Kubernetes API compatible type
-    pub fn to_kubernetes(&self, container_name: String) -> KubeContainerStatus {
+    pub fn to_kubernetes(&self, container_name: &str) -> KubeContainerStatus {
         let mut state = ContainerState::default();
         match self {
             Self::Waiting { message, .. } => {
@@ -67,7 +70,7 @@ impl Status {
         let ready = state.running.is_some();
         KubeContainerStatus {
             state: Some(state),
-            name: container_name,
+            name: container_name.to_string(),
             // Right now we don't have a way to probe, so just set to ready if
             // in a running state
             ready,
@@ -81,4 +84,98 @@ impl Status {
             ..Default::default()
         }
     }
+}
+
+/// Patch a single container's status
+pub async fn patch_container_status(
+    client: &kube::Api<KubePod>,
+    pod: &Pod,
+    container_name: &str,
+    status: &Status,
+    init: bool,
+) -> anyhow::Result<()> {
+    let containers: Vec<Container> = if init {
+        pod.init_containers()
+    } else {
+        pod.containers()
+    };
+    match containers
+        .iter()
+        .enumerate()
+        .find(|(_, container)| container.name() == container_name)
+    {
+        Some((container_index, container)) => {
+            let mut patches: Vec<json_patch::PatchOperation> = Vec::with_capacity(3);
+            // patches.push(json_patch::PatchOperation::Replace(
+            //     json_patch::ReplaceOperation {
+            //         path: "/metadata/resourceVersion".to_string(),
+            //         value: serde_json::Value::String("".to_string()),
+            //     },
+            // ));
+            let kube_status = status.to_kubernetes(container.name());
+            let path_prefix = if init {
+                format!("/status/initContainerStatuses/{}", container_index)
+            } else {
+                format!("/status/containerStatuses/{}", container_index)
+            };
+
+            patches.push(json_patch::PatchOperation::Replace(
+                json_patch::ReplaceOperation {
+                    path: "/metadata/resourceVersion".to_string(),
+                    value: serde_json::json!(""),
+                },
+            ));
+
+            patches.push(json_patch::PatchOperation::Replace(
+                json_patch::ReplaceOperation {
+                    path: format!("{}/state", path_prefix),
+                    value: serde_json::json!(kube_status.state.unwrap()),
+                },
+            ));
+            patches.push(json_patch::PatchOperation::Replace(
+                json_patch::ReplaceOperation {
+                    path: format!("{}/ready", path_prefix),
+                    value: serde_json::json!(kube_status.ready),
+                },
+            ));
+            patches.push(json_patch::PatchOperation::Replace(
+                json_patch::ReplaceOperation {
+                    path: format!("{}/started", path_prefix),
+                    value: serde_json::json!(true),
+                },
+            ));
+
+            let patch = json_patch::Patch(patches);
+            info!("{}", serde_json::to_string(&patch).unwrap());
+            let mut params = kube::api::PatchParams::default();
+            params.patch_strategy = kube::api::PatchStrategy::JSON;
+            client
+                .patch_status(pod.name(), &params, serde_json::to_vec(&patch)?)
+                .await?;
+            Ok(())
+        }
+        None => {
+            warn!(
+                "Container status update for unknown container {}.",
+                container_name
+            );
+            Ok(())
+        }
+    }
+}
+
+/// Create inital container status for registering pod.
+pub fn make_initial_container_status(container: &Container) -> KubeContainerStatus {
+    // Create empty patch and update only the fields we want to change.
+    let mut status: KubeContainerStatus = Default::default();
+    let mut state: ContainerState = Default::default();
+    status.name = container.name().to_string();
+    status.ready = false;
+    status.started = Some(false);
+    state.waiting = Some(ContainerStateWaiting {
+        message: Some("Registered".to_string()),
+        reason: Some("Registered".to_string()),
+    });
+    status.state = Some(state);
+    status
 }
--- a/crates/kubelet/src/container/status.rs
+++ b/crates/kubelet/src/container/status.rs
@@ -94,28 +94,45 @@ pub async fn patch_container_status(
     status: &Status,
 ) -> anyhow::Result<()> {
     match pod.find_container(&key) {
-        Some((container_index, container)) => {
+        Some(container) => {
             let kube_status = status.to_kubernetes(container.name());
-            let path_prefix = if key.is_init() {
-                format!("/status/initContainerStatuses/{}", container_index)
-            } else {
-                format!("/status/containerStatuses/{}", container_index)
-            };
 
-            let patches = vec![
-                json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
-                    path: format!("{}/state", path_prefix),
-                    value: serde_json::json!(kube_status.state.unwrap()),
-                }),
-                json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
-                    path: format!("{}/ready", path_prefix),
-                    value: serde_json::json!(kube_status.ready),
-                }),
-                json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
-                    path: format!("{}/started", path_prefix),
-                    value: serde_json::json!(true),
-                }),
-            ];
+            let patches = match pod.container_status_index(&key) {
+                Some(idx) => {
+                    let path_prefix = if key.is_init() {
+                        format!("/status/initContainerStatuses/{}", idx)
+                    } else {
+                        format!("/status/containerStatuses/{}", idx)
+                    };
+
+                    vec![
+                        json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
+                            path: format!("{}/state", path_prefix),
+                            value: serde_json::json!(kube_status.state.unwrap()),
+                        }),
+                        json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
+                            path: format!("{}/ready", path_prefix),
+                            value: serde_json::json!(kube_status.ready),
+                        }),
+                        json_patch::PatchOperation::Replace(json_patch::ReplaceOperation {
+                            path: format!("{}/started", path_prefix),
+                            value: serde_json::json!(true),
+                        }),
+                    ]
+                }
+                None => {
+                    let path = if key.is_init() {
+                        "/status/initContainerStatuses".to_string()
+                    } else {
+                        "/status/containerStatuses".to_string()
+                    };
+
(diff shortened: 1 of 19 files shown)


crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
         }
     }
 }
+
+/// Patch a single container's status
+pub async fn patch_container_status(
```
kflansburg, line 90: This relies on container statuses matching the containers in `Pod.containers` and `Pod.init_containers`, in the same order. I believe that this would be best practice.

crates/kubelet/src/pod/mod.rs
```diff
@@ -164,16 +166,11 @@ impl Pod {
     }
 
     /// Gets all of a pod's containers (init and application)
-    pub fn all_containers(&self) -> Vec<ContainerKey> {
-        let app_containers = self.containers();
-        let app_container_keys = app_containers
-            .iter()
-            .map(|c| ContainerKey::App(c.name().to_owned()));
+    pub fn all_containers(&self) -> Vec<Container> {
```
kflansburg, line 169: This seemed inconsistent with the other two methods `containers` and `init_containers`.

crates/kubelet/src/pod/status.rs
```diff
@@ -1,8 +1,28 @@
 //! Container statuses
 
+use super::Pod;
+use crate::container::{make_initial_container_status, ContainerMap, Status as ContainerStatus};
 use k8s_openapi::api::core::v1::ContainerStatus as KubeContainerStatus;
 
-use crate::container::{ContainerMap, Status as ContainerStatus};
+/// Initialize Pod status.
+pub fn make_registered_status(pod: &Pod) -> anyhow::Result<serde_json::Value> {
```
kflansburg, line 8: This initializes Pod status to include containers in the correct order as expected by `patch_container_status`. It is kind of required, so I'm wondering if this should be done prior to entering the first state handler.

crates/wasi-provider/src/states/running.rs
```diff
@@ -73,17 +27,22 @@ impl State<PodState> for Running {
...
+
             if let Status::Terminated {
                 timestamp: _,
                 message,
                 failed,
             } = status
             {
                 if failed {
-                    return Transition::next(self, Error { message });
+                    // This appears to be required by the test `test_module_exiting_with_error`
```
kflansburg, line 42: I don't agree with `fail_fatal` here or in the init pod state, but it appears to be required by the integration tests. The tests expect the Pod to fail whereas I would expect it to go to `Error`. Incidentally there appears to be a bug when the Pod does go to `Error` and then back to `Registered`, the wasm runtime encounters some error with a channel. We may need to do some clean up in the error state in anticipation of going back to `Registered`.

crates/wasi-provider/src/states/starting.rs
```diff
@@ -109,14 +108,14 @@ impl State<PodState> for Starting {
...
-            };
-            container_handles.insert(
-                ContainerKey::App(container.name().to_string()),
-                container_handle,
-            );
+            match start_container(pod_state, &pod, &container).await {
+                Ok(h) => {
+                    container_handles.insert(ContainerKey::App(container.name().to_string()), h);
+                }
+                // We should log, transition to running, and properly handle container failure.
```
kflansburg, line 115: This is somewhat tied to making all of these handlers infallible, but this one in particular looked like it would drop important container status patches. The same can be said of the `Initializing` state handler.

crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -299,9 +299,20 @@ impl WasiRuntime {
             let func = match export {
                 wasmtime::Extern::Func(f) => f,
                 _ => {
-                    return Err(anyhow::anyhow!(
-                    "_start import was not a function. This is likely a problem with the module"
-                ))
+                    let message = "_start import was not a function. This is likely a problem with the module";
```
kflansburg, line 302: This was the one Err in this method that did not notify the channel.

kflansburg (commented):
This was a first crack at simplifying container status patches. There may be some nits here, but I think this is supporting work for the state machine, so I would expect it to get merged and then the container state machine will likely involve further refactoring.

crates/kubelet/src/pod/status.rs
bacongobbler: As an aside, this comment would be great to add to the doc comments above on line 7.

crates/wasi-provider/src/states/running.rs
bacongobbler: Would it be all right if we pulled this out into a separate PR? That way we can have this discussion separate from the container status patching logic.

crates/wasi-provider/src/wasi_runtime.rs
bacongobbler: Good catch. I swore I updated this to the new notify logic...

bacongobbler:
Looks great. Going to take this for a spin to make sure things work as expected. But the fact that the container status patching logic was removed from the wasi provider is huge!

crates/wasi-provider/src/states/running.rs
itowlson: There was discussion about this on the 'remove panics from handlers' PR - this was to replicate existing behaviour which was relied on by the integration tests.  I think everyone agrees that it is not the desired behaviour, but changing the behaviour should be tackled as a specific work item (which will be non-trivial), not rolled in along with other work.

itowlson: Never mind, I see you already raised an issue to do this as a separate item.  Thanks!

crates/kubelet/src/pod/mod.rs
itowlson: I suspect at the time it was written it was used primarily for getting keys/IDs rather than the full container objects. But it looks like now the only places it is used would work just as well with the full container objects.  There is also a place in `store/mod.rs` that explicitly constructs its own `all_containers` collection which this modified method could supplant.

itowlson (changes requested):
This will be great to have!  Unfortunately I have reservations about relying so absolutely on the order of container statuses in the status section being in sync with the order of specs in the spec section.  It seems like if anything causes this assumption to become violated we have no way to detect and adapt (or even fail informatively) - we will just blindly write statuses to the wrong container.  Are we that confident about the assumption, and if not can we come up with an alternative approach?

crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
...
+    client: &kube::Api<KubePod>,
+    pod: &Pod,
+    container_name: &str,
+    status: &Status,
+    init: bool,
```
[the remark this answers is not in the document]
kflansburg: Did not realize `ContainerKey` included init vs. app. Will change this.

crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
...
+) -> anyhow::Result<()> {
+    let containers: Vec<Container> = if init {
+        pod.init_containers()
+    } else {
+        pod.containers()
+    };
+    match containers
+        .iter()
+        .enumerate()
+        .find(|(_, container)| container.name() == container_name)
```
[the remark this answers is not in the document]
kflansburg: Makes sense.

crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
...
+    {
+        Some((container_index, container)) => {
+            let mut patches: Vec<json_patch::PatchOperation> = Vec::with_capacity(3);
```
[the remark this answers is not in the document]
kflansburg: There are four, I do not know if the resource version is actually necessary, but it was in the original WASI provider code and I added it back when I was debugging. In general I dont think its an issue to use a "magic number" in this case, as long as the number is actually correct. As far as the vec literal, that seems reasonable, although I will probably not create 3+ methods for that, just define the ReplaceOperation inside the vec literal.

crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
...
+            ));
+            patches.push(json_patch::PatchOperation::Replace(
+                json_patch::ReplaceOperation {
+                    path: format!("{}/started", path_prefix),
+                    value: serde_json::json!(true),
+                },
+            ));
+
+            let patch = json_patch::Patch(patches);
+            let mut params = kube::api::PatchParams::default();
```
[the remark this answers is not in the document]
kflansburg: Yes, I believe that syntax works.

crates/kubelet/src/container/status.rs
```diff
@@ -82,3 +85,90 @@ impl Status {
...
+        .find(|(_, container)| container.name() == container_name)
+    {
+        Some((container_index, container)) => {
+            let mut patches: Vec<json_patch::PatchOperation> = Vec::with_capacity(3);
+            let kube_status = status.to_kubernetes(container.name());
+            let path_prefix = if init {
+                format!("/status/initContainerStatuses/{}", container_index)
+            } else {
+                format!("/status/containerStatuses/{}", container_index)
+            };
```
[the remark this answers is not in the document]
kflansburg: I do not think it is guaranteed. I dont think it is reasonable to fetch the Pod each time it needs to be patched. We could store some up-to-date version of the Pod status, and overwrite the entire status when we patch a container. Is there any expectation that things other than than the scheduled kubelet update the status of the Pod under normal operation?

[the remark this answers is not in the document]
thomastaylor312: FWIW, I tried to write this as a JSON patch the first time I was tackling it as I didn't like the idea of fetching the pod each time. This ordering problem was the stickler then too. However, I think we might be able to remove some of the overhead of fetching the Pod each time if we switch to having a reflector. A reflector implementation updates its cache each time an object updates, so then we can just fetch the object from the cache. We could still end up with a case where that object updates after we fetch from the cache, but that should be fairly unlikely as I don't think anything else but the master could update a pod status. What I don't know is how many controllers might try to update pod statuses though.

crates/kubelet/src/container/status.rs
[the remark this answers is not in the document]
thomastaylor312: Yep, the syntax is

```rust
let thing = PatchParams {
    foo: "bar",
    ..Default::default()
};
```

kflansburg:
I have addressed all comments, except for switching to a Reflector. What I think makes the most sense is that `Pod::queue` should update the Pod definition that it is driving the state machine with whenever updates arrive. This would effectively be a reflector, and mean that the `Pod` passed to `json_status` will be the latest definition. I think this is the simplest solution and avoids having some shared state of Pod definitions that `patch_container_status` would need access to.

crates/kubelet/src/state.rs
```diff
@@ -254,44 +257,49 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
...
+    pod: Arc<RwLock<Pod>>,
 ) {
-    let api: Api<KubePod> = Api::namespaced(client.clone(), pod.namespace());
+    let initial_pod = { pod.read().await.clone() };
+    let namespace = initial_pod.namespace().to_string();
+    let name = initial_pod.name().to_string();
+    let api: Api<KubePod> = Api::namespaced(client.clone(), &namespace);
+
+    // Initialize Pod Container Statuses
+    patch_status(&api, &name, make_registered_status(&initial_pod)).await;
```
kflansburg, line 268: I have implemented my suggestion of applying this initial status update prior to applying the Provider's first status. I also removed the use of this function in the Providers to verify that this eliminates their responsibility to do this.

kflansburg (commented):
This last commit implements my suggestion of updating the Pod manifest that is supplied to the State handlers.

crates/kubelet/src/state.rs
```diff
@@ -254,44 +257,49 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
...
+    drop(initial_pod);
```
thomastaylor312, line 269: Nit: consider doing this in a block (along with the `read()` call) so as to avoid the need for a manual `drop`

thomastaylor312 (approved):
I think with the new method of doing this, I am now ok with merging. Let's wait on @itowlson for his final review

crates/kubelet/src/state.rs
```diff
@@ -258,14 +266,62 @@ pub async fn run_to_completion<PodState: Send + Sync + 'static>(
...
+    // few reasons:
+    // * Kubernetes rewrites an empty array to null, preventing us from
+    //   starting with that and appending.
+    // * Pod reflection is not updated within a given state, meaning that
+    //   container status patching cannot be responsible for initializing this
+    //   (this would be a race condition anyway).
+    // I'm not sure if we want to loop forever or handle some sort of failure
+    // condition (if Kubernetes refuses to accept and propagate this
+    // initialization patch.)
+    'main: loop {
```
thomastaylor312, line 288: Ok, think I have a few concerns about this. This is mainly broken down into two things 1) I don't know if this is any better than our old method of doing things and 2) If we do continue doing it this way, I think there could be a better way of doing this. 

1. So I don't know if this provides any benefit on our previous version other than API call reduction. We still have a bunch of waiting and things that we end up doing (no matter how we implement the retry). I think it could actually end up being slower as things step through the state machine
2. If we do want to go this route, I think we should reuse/export the `retry!` macros from the node code at the very minimum. I think a more elegant solution could entail wrapping the pod in a custom struct with a `RwLock` that write locks until it receives a response from the API and then abstract that all away behind `pod.read()` like we already have

thomastaylor312 (changes requested):
I have some concerns with the loop approach in the latest commit. Detailed in the comments below

crates/kubelet/src/state.rs
kflansburg: So regarding performance, I think this is better for two reasons:
* this loop only runs when a Pod is first scheduled, not on each state change.
* in most cases, all it does is wait a second or two for the reflector to update, and we don't have to try to patch multiple times. 
I will address the other nits.

thomastaylor312: > this loop only runs when a Pod is first scheduled, not on each state change.

Ok, so I misread that. I do think we should have a max number of retries (preferably as a `const`) like what you have now and then just error out and let k8s know the pod has errors (this might be a good use of the `Unknown` phase)

kflansburg: Sounds good will do.

crates/kubelet/src/pod/mod.rs
```diff
@@ -153,17 +154,26 @@ impl Pod {
...
 
-    /// Finds the index of the container in the Pod's container status.
+    /// Finds the index of the container in the Pod's container statuses.
     pub fn container_status_index(&self, key: &ContainerKey) -> Option<usize> {
-        let containers: Vec<Container> = if key.is_init() {
-            self.init_containers()
+        let statuses: Vec<KubeContainerStatus> = if key.is_init() {
+            self.kube_pod
+                .status
+                .clone()
```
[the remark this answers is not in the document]
kflansburg: I dont think you can really do this: `.unwrap_or_else(|| &PodStatus::default())`, because the reference doesnt outlive the closure. I will see what I can do though.

thomastaylor312 (approved):

PR #426, opened 2020-10-18
[WIP] Upgrade Tokio to 0.3.0

kflansburg:
This commit updates Krustlet's Tokio version to `0.3.0` and updates Krustlet's
use of Tokio API wherever there were breaking changes. As a result, it compiles.

Krustlet will not run, however, and tests will not pass because futures created
by crates which have not been updated will panic due to the lack of a compatible runtime.

To run these futures, we must identify each one, bring `tokio_compat_02::FutureExt` in scope,
and call `compat()` on the future. This appears to work quite nicely and I have not noticed
any issues with tasks from both version interacting with each other (in limited testing).

The concern I have is that this will run `0.2.x` tasks in a separate, single-threaded runtime.
I believe this may have some performance issues, but this will be made more clear once we compile
a list of the futures in question.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,7 +21,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "log",
- "tokio",
+ "tokio 0.2.22",
  "tokio-util 0.2.0",
 ]
 
@@ -37,7 +37,7 @@ dependencies = [
  "futures-sink",
  "log",
  "pin-project",
- "tokio",
+ "tokio 0.2.22",
  "tokio-util 0.3.1",
 ]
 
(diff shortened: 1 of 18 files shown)


MikailBag:
IIUC (looking at https://github.com/LucioFranco/tokio-compat-02/blob/d88c6f1ce64c6c74340e7e800ba631a9f8a5e8b0/src/lib.rs#L42), 
only tasks created by `tokio_02::spawn` will run on this runtime. Other tasks, even if they use 02 IO types, will run on tokio 03 started in `fn main`.

Additionally this 02 runtime will be responsible for *waking* all tasks using 02 IO types, regardless of what runtime they are running on.

thomastaylor312:
@kflansburg do we get enough benefits from upgrading to 0.3 right now that outweigh the backcompat cruft we are adding in? That would be my worry here. If it isn't adding significant benefits, I think we let other crates catch up (or even open PRs with them to do it) before we move ourselves

kflansburg:
Tracking this work in #480

PR #433, opened 2020-10-21
cross-compile for aarch64

### LINKED ISSUE #358: Pre-compiled ARM binaries
As a user, I would like to be able to download `aarch64` prebuilt binaries so I can run Krustlet on 64 bit ARM devices (like Raspberry Pi).

Other notes:

We may also want to have an "openssl-less" ARM binary to make it easier to just drop in on an embedded device

bacongobbler:
closes #358

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ jobs:
   build:
     name: build release assets
     runs-on: ${{ matrix.config.os }}
+    env: ${{ matrix.config.env }}
     strategy:
       matrix:
         config:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -18,6 +18,14 @@ jobs:
               args: "--release",
               extension: "",
             }
+          - {
+              os: "ubuntu-latest",
+              arch: "aarch64",
+              args: "--release --target aarch64-unknown-linux-gnu",
+              extension: ".exe",
```
thomastaylor312, line 25: This extension should be empty

.github/workflows/release.yml
```diff
@@ -57,7 +65,24 @@ jobs:
           rustup component add clippy --toolchain stable-x86_64-apple-darwin
           rustup update stable
 
+      - name: setup for cross-compile builds
+        if: runner.os == 'linux'
```
thomastaylor312, line 69: We should make this `if` statement use something in the matrix config so we don't end up compiling OpenSSL during a normal linux build

thomastaylor312 (changes requested):
Excited to have this. Just two small fixes needed

.github/workflows/release.yml
bacongobbler: good catch

.github/workflows/release.yml
bacongobbler: hmm... good idea. perhaps `matrix.config.arch == "aarch64"` would do

thomastaylor312 (approved):

thomastaylor312:
This is good to go from the code/configuration side. Looks like something broke with the pipeline though

bacongobbler:
all tests passed. Github actions is holding onto the old matrix, but that should fix itself once this is merged.

PR #434, opened 2020-10-21
fix(kubelet): Make sure we actually box

thomastaylor312:
We weren't actually boxing our fused futures so there was still a bunch
of stuff pinned on the stack. This was causing stack overflow problems
on Windows


itowlson (approved):
Good catch - thanks!

bacongobbler (approved):

PR #435, opened 2020-10-22
pass all environment variables to each capability

### LINKED ISSUE #338: investigate environment variables for waSCC
From a comment in #331:

> Reading the source code more closely, this seems intentional from a design standpoint. Each call to `bind_actor` contains a `HashMap<String, String>` that defines the configuration used between the actor and the capability. As far as I'm aware of, there is no concept of an environment variable in waSCC.
>
> In other words, right now we're binding every environment variable in the pod manifest as configuration for the http server capability, and right now the only value the http server capability cares about is PORT.
>
> Another way to put this is that any environment variables you add to the pod manifest are currently ignored or unused.

bacongobbler:
Prior versions only passed a shortlist of known configuration to each capability. This PR passes the entire environment variable tree to each waSCC capabililty.

This effectively does nothing at the moment, but if we are to implement a way for users to provide their own capability providers, then they can populate those capabilities with their pod's environment variables.

closes #338

### DIFF
--- a/crates/wascc-provider/src/lib.rs
+++ b/crates/wascc-provider/src/lib.rs
@@ -342,19 +342,14 @@ impl kubelet::log::HandleFactory<tokio::fs::File> for LogHandleFactory {
 fn wascc_run(
     host: Arc<Mutex<Host>>,
     data: Vec<u8>,
-    mut env: EnvVars,
+    env: EnvVars,
     volumes: Vec<VolumeBinding>,
     log_path: &Path,
     port_assigned: u16,
 ) -> anyhow::Result<ContainerHandle<ActorHandle, LogHandleFactory>> {
     let mut capabilities: Vec<Capability> = Vec::new();
     info!("sending actor to wascc host");
-    let log_output = NamedTempFile::new_in(log_path)?;
-    let mut logenv: HashMap<String, String> = HashMap::new();
-    logenv.insert(
-        LOG_PATH_KEY.to_string(),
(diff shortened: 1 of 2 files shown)


crates/wascc-provider/src/lib.rs
```diff
@@ -371,11 +371,12 @@ fn wascc_run(
     }
 
     if actor_caps.contains(&HTTP_CAPABILITY.to_owned()) {
-        env.insert("PORT".to_string(), port_assigned.to_string());
+        let mut httpenv = env.clone();
+        httpenv.insert("PORT".to_string(), port_assigned.to_string());
```
thomastaylor312, line 375: My only concern here (and with log path) is that if a user specifies an env variable of `PORT` it will screw up the port configuration

thomastaylor312 (approved):
One minor concern. Nothing blocking though

crates/wascc-provider/src/lib.rs
bacongobbler: Right. I was considering that too. However, the internal port number the http server is listening on is set up internally, so the user would not know what port to specify.

Log path is the same way: these are internal temporary ephemeral paths that the users should not be able to specify. It would be simple for a user to set a LOG_PATH that could read another actor's logs, leaking sensitive information to an attacker.

thomastaylor312: And we are overwriting, so I think it should be ok

Issue #182, opened 2020-04-09
How should Krustlet handle the Container Storage Interface?


thomastaylor312:
Many of the cloud provider volumes have Container Storage Interface (CSI) plugin implementations to handle mounting volumes to containers. Right now, the attachdetach-controller will attach cloud provider volumes to VMs but will not actually mount the volume. So a few of questions for discussion:

1. Should we implement CSI calls inside of krustlet itself (or at least something similar)?
2. Will it be possible to reuse the current Go plugins or are they too container-centric?
3. Would it be better to leave volume mounting up to individual implementations?
4. Would a good middle ground just be to implement drivers for open source file mounting (NFS, SMB, etc)?

For reference, the CSI docs can be found here: https://kubernetes-csi.github.io/docs/

thomastaylor312:
One other question I have discovered while implementing CSI: Many of the current cloud providers use the "legacy" cloud controller manager. After some initial digging, I am still not sure how those disks get attached to a node in an AKS/EKS/GKE cluster, but it may Just Work™. We should do some investigation/testing with that to see what happens

bacongobbler:
I agree. It's not totally clear how the sausage is made.

As far as I understand the system, most storage class provisioners today are bundled with the cluster. Those "in-tree provisioners" are being deprecated in favour of "external provisioners". The Kubelet is built with the understanding on how to make calls to the correct external provisioner given the provided storage class name.

https://kubernetes.io/docs/concepts/storage/volumes/

With #445, CSI plugins can now register themselves with Krustlet. So we have support right now to register external provisioners like the [azure disk CSI driver](https://github.com/kubernetes-sigs/azuredisk-csi-driver). The trick is that most documentation recommends running them on every node via a DaemonSet... As a container, naturally.

So there are two calls to action we need to address to support volume mounts via the CSI:

1. Since CSIs are exposed to workloads through a Storage Class, they are referenced through a PersistentVolumeClaim. As I understand it, we currently do not have PVC mount support. We will need to add PVC mount support, which may mean the providers will need to issue CSI calls to provision claims (more research is needed)
2. We will need to document *how* to run a storage class provisioner alongside Krustlet, like the azure disk CSI driver.

To answer your other questions:

> * Will it be possible to reuse the current Go plugins or are they too container-centric?

I believe it will be possible to reuse the current CSI drivers. They were initially written to be run as a DaemonSet on Kubernetes, but the drivers themselves are just a bunch of binaries with a large swath of feature flags.

[...]

PR #445, opened 2020-11-02
Adds plugin support to Krustlet

thomastaylor312:
BEHOLD THE HACKERY! Ok, so yeah. This PR is a hack and a half, so let's start with what this is adding.

## What this adds
This PR adds the equivalent of the plugin watcher and registration system from Kubelet as part of the work for #182. This does not actually implement CSI providers, but provides all the scaffolding for us to be able to do so. The plugin watcher is running, but not actually passed into providers or anything else yet. One other thing to note is that we only support CSI plugins with this interface as of now, so device plugins are not yet supported.

### Notify and MacOS
The "API" expected for plugins in Kubelet is a watched directory where Unix domain sockets are opened and discovered using `inotify` or other system equivalent. The `notify` crate is a handy cross-platform option for notifying on file changes. However, on Mac, I observed that newly created sockets were not observed by the underlying FSNotify library. It would detect metadata changes (such as those caused by `touch`), but not the create event, which meant that plugins were never discovered. I was able to reproduce the bug using the raw C API and opened a bug report with Apple (FB8830541, non-public insofar as I can tell). If we are being honest with ourselves, this means a bug fix is at least a year out, if not longer.

Because of this I had to write a really ugly polling loop that lists files in the directory and checks for differences. It isn't nice, but it is functional. Windows and other Unix style platforms use their own built in libraries under the hoodand are not affected by this hack.

### Unix Domain Socket support for Windows
Unix domain sockets landed in Windows 10 a few years ago, however, there is still not support in the Rust standard library for them due to their compatibility guarantees with older versions of Windows (see [this issue](https://github.com/rust-lang/rust/issues/56533)). There is a [handy library](https://github.com/Azure/mio-uds-windows) (in the Azure org no less) with mio support, but it is not published. There is a fork of this repo that has been updated recently, but it removed mio support. There is also a tokio crate from the same author, but it was built against 0.1 and the current version right now is 0.3. The mio-uds-windows crate, however, works with tokio 0.2 (our current version) and should be easier to update to work with mio 0.7 and tokio 0.3 when we complete that upgrade.

Because we cannot publish a crate with a git dependency, I copied all of the code into a module within the `kubelet` crate. This module is the exact same code modified to be a module instead of its own crate (and cleaning up use of deprecated functions/macros). All other code is untouched. This module and its dependencies are only included on Windows builds. As soon as there is mainline support for Unix Domain Sockets on Windows in tokio/mio, we should remove this code.

Due to this custom code, I had to wrap the logic for creating a gRPC server listening on a socket in another new module that conditionally includes the correct library. The server implementation is currently only used for testing purposes, but the client half is used in the normal code path.

## Other Questions

**Why is this so large?**
Because I copied in a whole other project and had to write a bunch of wrapper code and hacks, this PR is pretty big. So there isn't really an easy way to break this up into smaller chunks that make sense. I could have done each part (copying the code, adding the watch hack, etc) as its own PR, but it would have been completely out of context. Anyway, I apologize for the size, and please read the next question before reviewing

**What do I need to review?**
Please DO NOT spend time reviewing the code in the `mio_uds_windows` module. This code is essentially unchanged from where I copied it. It also has its own tests that run. Please do review everything else though and make sure I didn't miss any major things I should have when writing those hacks.


crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+        }
+        Ok(())
+    }
+
+    async fn handle_delete(&self, event: Event) {
+        let mut lock = self.plugins.write().await;
+        for deleted_plugin in event
+            .paths
+            .into_iter()
+            .filter(|p| !p.is_dir() && p.extension().unwrap_or_default() == SOCKET_EXTENSION)
```
itowlson, line 180: Deduplicate into e.g. `plugin_paths(event)` or `socketlike_paths(event)`

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+        {
+            let key = match lock.iter().find(|(_, v)| *v.plugin_path == deleted_plugin) {
```
itowlson, line 182: `lock` is not a helpful name here - consider e.g. `plugins` or `registered_plugins` or, if you want to focus on its role as an accessor, `plugins_writer` or even `plugins_lock`.

itowlson: To be honest my difficulty with the word `lock` is greatly exacerbated by it being right next to the word `key`.

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+                // Take ownership of the key to avoid an immutable borrow
+                Some((key, _)) => key.to_owned(),
+                // If for some reason it is already gone, no need to error
+                None => continue,
+            };
+            lock.remove(&key);
+        }
```
itowlson, line 189: Consider pulling this block out into a helper function `remove_plugin(plugins /* or lock */, deleted_plugin)`.  There is a lot of Rustish ceremony about `iter` and `find` and `to_owned` here that obscures the very simple intent.

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+    async fn validate(&self, info: &PluginInfo, discovered_path: &PathBuf) -> anyhow::Result<()> {
+        trace!(
+            "Starting validation for plugin {:?} discovered at path {}",
+            info,
+            discovered_path.display()
+        );
+        // Step 1: Check for valid type and if it is a CSIPlugin
+        let plugin_type = PluginType::try_from(info.r#type.as_str())?;
+        trace!("Type validation complete for plugin {:?}", info);
+        if matches!(plugin_type, PluginType::DevicePlugin) {
```
itowlson, line 209: Consider allow-listing rather than deny-listing.  Also consider pulling out a helper constant `ALLOWED_PLUGIN_TYPES` and a helper function `is_allowed_plugin_type(&PluginInfo)`.

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+            warn!("DevicePlugins are not currently supported");
+            return Err(anyhow::anyhow!("DevicePlugins are not currently supported"));
+        }
+
+        // Step 2: Check if we support this version
+        trace!("Checking supported versions for plugin {:?}", info);
```
itowlson, line 215: Again consider pulling out a helper function.  Generally consider refactoring this big function with `Step N` comments into something like:

```
self.validate_is_allowed_plugin_type(info)?;
self.validate_is_allowed_api_version(info)?;
self.validate_is_unique_name(info)?;
Ok(())
```

This makes the list of validations extremely clear, and makes it possible to inspect and review the logic of each validation in isolation under an intent-expressing name.

crates/kubelet/src/fs_watch/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+            .await
+            .map(|entries| {
+                entries
+                    .into_iter()
+                    .map(|e| e.path())
+                    .collect::<HashSet<PathBuf>>()
+            })
+    }
+
+    fn handle_creates(
```
itowlson, line 141: This isn't so much 'handling' as 'sending' I feel...

crates/kubelet/src/fs_watch/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+            // At this point there isn't much we can do as the channel is closed. So just log an
+            // error
+            error!(
+                "Unable to send event {:?} due to the channel being closed",
+                e.0
+            );
+        }
+    }
+
+    fn handle_deletes(
```
itowlson, line 165: This implementation seems almost identical to `handle_creates`.  Consider pulling out a `handle_changes` (or `send_changes`) that takes an `EventKind` as a parameter and putting all the common logic there.

itowlson (commented):
I confused the heck out of myself with the paths stuff in the plugin watcher (sockets? files? endpoints?) so you may need to do a lot of filtering of misguided comments there I'm afraid.  Perhaps a little more documentation and clearer naming - but mostly this is my own ignorance and perhaps that's not your problem to solve!

Other than that... naming and breaking up, really.  It all looks like it makes sense but pulling out some small blocks and giving them names would really help readability.  So functionally fine but just some style stuff.  And congratulations on landing this - what an epic!

kflansburg (commented):
I think this looks pretty good, nothing in particular jumped out at me regarding code quality. Can you share a link to some documentation on this registration mechanism in Kubernetes? I found this:

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md

That document describes something a little different where it looks like plugins register themselves with Kubelet, and it does not have to watch the filesystem.

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+use tonic::Request;
+
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::path::{Path, PathBuf};
+
+#[cfg(target_family = "unix")]
+const DEFAULT_PLUGIN_PATH: &str = "/var/lib/kubelet/plugins_registry/";
+#[cfg(target_family = "windows")]
+const DEFAULT_PLUGIN_PATH: &str = "c:\\ProgramFiles\\kubelet\\plugins_registry";
```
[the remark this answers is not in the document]
thomastaylor312: I figured you'd have comments here! Yeah, I forgot a space, but also related to your other comment about this, I am not sure what a good default is. The default path for unix-like things is from the k8s docs, but I couldn't really figure out where the equivalent would be on windows. Any ideas here?

I am also curious on your thoughts around whether or not we should use the defaults we have in `Config` (which follow the rest of our config stuff in `$HOME/.krustlet/<somewhere>`)

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+    }
+
+    /// Gets the endpoint for the given plugin name, returning `None` if it doesn't exist
+    pub async fn get(&self, plugin_name: &str) -> Option<PathBuf> {
+        let lock = self.plugins.read().await;
+        lock.get(plugin_name)
+            .map(|v| v.endpoint.as_ref().unwrap_or(&v.plugin_path).to_owned())
+    }
+
+    /// Starts the plugin registrar and runs all automatic plugin discovery and registration loops
```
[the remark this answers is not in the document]
thomastaylor312: I can add some more detail here. I just made the bad assumption that it being called `run` + being async would clue people in

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+    let req = Request::new(RegistrationStatus {
+        plugin_registered: error.is_none(),
+        error: error.unwrap_or_else(String::new),
+    });
+
+    trace!("Calling NotifyRegistrationStatus at {:?}", path);
+    client
+        .notify_registration_status(req)
+        .await
+        .map(|_| ())
```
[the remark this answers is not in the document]
thomastaylor312: Oh that is a good idea. Good catch on this

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+impl Default for PluginRegistrar {
+    fn default() -> Self {
+        PluginRegistrar {
+            plugin_dir: PathBuf::from(DEFAULT_PLUGIN_PATH),
+            plugins: RwLock::new(HashMap::new()),
+        }
+    }
+}
+
+impl PluginRegistrar {
```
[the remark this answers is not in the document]
thomastaylor312: I can change it to `PluginRegistry`

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+            .await?;
+
+        // Manually assemble an event and call handle_event to reconfigure sockets properly on restart
+        self.handle_create(Event {
+            paths: dir_entries,
+            ..Default::default()
+        })
+        .await?;
+
+        let mut event_stream = FileSystemWatcher::new(&self.plugin_dir)?;
```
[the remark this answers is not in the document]
thomastaylor312: Yes there definitely is a possible race condition here (as seen by my delay in the tests). Honestly not quite sure how to handle that situation here without rescanning the directory one more time before dropping into the loop

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+        lock.get(plugin_name)
+            .map(|v| v.endpoint.as_ref().unwrap_or(&v.plugin_path).to_owned())
+    }
+
+    /// Starts the plugin registrar and runs all automatic plugin discovery and registration loops
+    pub async fn run(&self) -> anyhow::Result<()> {
+        // Create plugin directory if it doesn't exist
+        create_dir_all(&self.plugin_dir).await?;
+
+        // Walk the plugin dir before hand and process any currently existing sockets
```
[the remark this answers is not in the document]
thomastaylor312: Ok, so this is confusing. Perhaps I can add some documentation to clarify how Kubelet does this (because this is mostly from reverse engineering and gluing together disparate k8s docs). You basically have it right. The directory that kubelet is watching is where running plugins (which could be running in a container in normal k8s, but is more likely a separate service in Krustlet's case) place their sockets for Kubelet to communicate with. However, the actual specification of a `PluginInfo` supports two types of plugins: `CSIPlugin` and `DevicePlugin`. CSI plugins work by opening a `csi.sock` in another directory outside of the one watched by Kubelet. So there ends up being two endpoints: one for the plugin itself and another for the CSI stuff. DevicePlugins, which we don't yet support, only have the single path.

I'll add documentation to clarify this

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+    /// Returns a new plugin registrar configured with the given plugin directory path
+    pub fn new<P: AsRef<Path>>(plugin_dir: P) -> Self {
+        PluginRegistrar {
+            plugin_dir: PathBuf::from(plugin_dir.as_ref()),
+            ..Default::default()
+        }
+    }
+
+    /// Gets the endpoint for the given plugin name, returning `None` if it doesn't exist
+    pub async fn get(&self, plugin_name: &str) -> Option<PathBuf> {
```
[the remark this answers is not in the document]
thomastaylor312: Unfortunately the plugin file path _can_ be an endpoint. Not right now, as we only support CSI plugins. But DevicePlugins do not have `endpoint` set as I mentioned in my [other comment](https://github.com/deislabs/krustlet/pull/445#discussion_r516838528). I do think renaming this to `get_endpoint` is a good idea though. But basically PathBuf _does_ accurately represent the detected file path (which is a socket and an "endpoint" if another endpoint is configured) and a possibly configured "endpoint" (which is also a socket)

crates/kubelet/src/plugin_watcher/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+        let mut event_stream = FileSystemWatcher::new(&self.plugin_dir)?;
+
+        while let Some(res) = event_stream.next().await {
+            match res {
+                Ok(event) if event.kind.is_create() => {
+                    if let Err(e) = self.handle_create(event).await {
+                        error!("An error occurred while processing a new plugin: {:?}", e);
+                    }
+                }
+                Ok(event) if event.kind.is_remove() => self.handle_delete(event).await,
```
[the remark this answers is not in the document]
thomastaylor312: I am not entirely sure what you are envisioning here. Could you clarify a bit? Both `handle_*` functions are `&self` so they can validate properly, so I am not sure what adding another function would do other than breaking it out (which is also fine, but doesn't seem to be what you are implying)

crates/kubelet/src/plugin_watcher/mod.rs
thomastaylor312: Good call

crates/kubelet/src/plugin_watcher/mod.rs
thomastaylor312: Since this is the only type we have to check for and it may end up being removed if we add support in the future, is there any benefit in doing this? Doesn't feel like it adds much benefit in this case

thomastaylor312:
@kflansburg Yeah, I am going to add some documentation. That was the first plugin system I found, but it turns out there is _another_ plugin system that CSI plugins use

crates/kubelet/src/plugin_watcher/mod.rs
itowlson: The benefit of allow-listing instead of deny-listing is that we _opt into_ plugin types when we are ready to support them.  The current test says "hey, we support everything except device plugins!"  Whereas actually we only support CSI plugins.  It _so happens_ that you know that these _so happen_ to amount to the same thing.  But I feel that it matters _for the understanding of others_ (and to a lesser extent for future-proofing) how the code expresses it.

The benefit of pulling out a helper function is _naming_.  Don't make me deliver the rant about how functions are primarily there for naming and only incidentally for reuse.  You wouldn't like me when I'm delivering the rant about etc. etc.

crates/kubelet/src/plugin_watcher/mod.rs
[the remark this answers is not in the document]
itowlson: Could you handle it by having the watcher raise an event for the initial set of files instead of only for changes?  Or do the built-in ones not support that behaviour?

Otherwise I guess you need to start the watcher, then do the scan, and handle the possibility of a duplicate (a file that is created immediately after the watcher starts, so gets a Create event, but is also picked up in the subsequent "initial state" scan).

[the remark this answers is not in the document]
thomastaylor312: I don't think the built in ones support that behavior, definitely not without escaping down to OS specific code. But even then I don't think they do

crates/kubelet/src/plugin_watcher/mod.rs
thomastaylor312: I still don't think this really adds much in way of clarity given the use of an enum and feels much more like ceremoney, but I'll roll with with it for now as I want to see if I like it as much as you do 😄

thomastaylor312: After implementing it, I personally think that it just adds more code with a zero to possibly negative readability increase. If you feel strongly about it, I'll keep it

crates/kubelet/src/grpc_sock/windows/mod.rs
```diff
@@ -0,0 +1,91 @@
+// This is a modified version of: https://github.com/hyperium/tonic/blob/f1275b611e38ec5fe992b2f10552bf95e8448b17/examples/src/uds/server.rs
```
bacongobbler, line 1: Do we need to include a LICENSE or explain what changes we made to the upstream implementation?

Are the changes made here able to be patched upstream?

Basically I'm trying to determine if we need to take ownership over this codebase or if it's just a temporary fork.

crates/kubelet/src/fs_watch/mod.rs
```diff
@@ -0,0 +1,306 @@
+//! A simple abstraction layer over OS specific details on watching a filesystem. Due to a bug in
+//! MacOS with sending an event on socket creation, we need to implement our own hacky watcher. To
+//! keep it as clean as possible, this module abstracts those details away behind a `Stream`
+//! implementation. A bug has been filed with Apple and we can remove this if/when the bug is fixed
```
bacongobbler, line 4: Would you mind adding a tracking ID to determine when we can remove this file?

crates/kubelet/src/mio_uds_windows/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "mio-uds-windows"
+version = "0.1.0"
+license = "MIT"
+authors = ["Azure IoT Edge Devs"]
+description = "Unix Domain Sockets for mio...on Windows!"
+repository = "https://github.com/Azure/mio-uds-windows"
```
bacongobbler, line 7: if this is a permanent fork, should we update the crate info?

bacongobbler (commented):
The code itself looks good to me. I'm mostly curious on how we should handle the vendored code as well as how we will be packaging these crates moving forward.

crates/kubelet/src/mio_uds_windows/Cargo.toml
bacongobbler: I'm also wondering if this crate (along with any others introduced here) should be pulled out of Krustlet and maintained as a separate crate. They appear to help in some regards (file watching, Windows UDS, gRPC sockets...) but they're semi-orthogonal to Krustlet.

What do you think about pulling these crates out to be maintained separately?

crates/kubelet/src/grpc_sock/windows/mod.rs
thomastaylor312: This is our own code and not a fork. Specifically, this is copied from their examples directory and tweaked to work with Windows. In these kinds of cases, a link back to the code feels fine

crates/kubelet/src/fs_watch/mod.rs
thomastaylor312: It is in the PR description, but it is also non-public (couldn't find a setting for it). Do you still want it here?

crates/kubelet/src/mio_uds_windows/Cargo.toml
thomastaylor312: Oh! That cargo.toml file shouldn't even be there any more. I'll remove it

As for the separate crate question, ff we put them out into another crate, we'd have to publish them in order to publish krustlet as a crate. I don't want to accidentally put yet another fork of this out into the crates ecosystem when it is purely for use by us

crates/kubelet/src/fs_watch/mod.rs
bacongobbler: Yes please! That way it makes it easier to track down the line. I've found that historical references being stored in code makes it easier to track several years down the line in case it's never fixed upstream. Just something like "Tracking ID FB8830541 - @thomastaylor312 can look up the status of the bug".

bacongobbler (approved):
LGTM!

itowlson (approved):
One minor suggestion, otherwise LGTM

PR #451, opened 2020-11-10
References were formatted incorrectly on Windows

### LINKED ISSUE #446: Windows tests failing with OCI paths
There are two components to this bug:

1. Windows tests pipelines are still showing green even when there are failures. This likely has something to do with my lack of Powershell skills
2. There are currently failures with the OCI crate:

```
failures:
    store::oci::file::test::file_module_store_always_pulls_if_tag_and_policy_omitted
    store::oci::file::test::file_module_store_always_pulls_if_tag_latest_and_policy_omitted
    store::oci::file::test::file_module_store_can_pull_if_policy_always
    store::oci::file::test::file_module_store_can_pull_if_policy_if_not_present
    store::oci::file::test::file_module_store_can_pull_if_tag_given_but_policy_omitted
    store::oci::file::test::file_module_store_can_reuse_cached_if_policy_never
    store::oci::file::test::file_module_store_copes_with_no_tag
    store::oci::file::test::file_module_store_gets_updates_if_policy_always
    store::oci::file::test::file_module_store_ignores_updates_if_policy_if_not_present

test result: FAILED. 25 passed; 9 failed; 0 ignored; 0 measured; 0 filtered out
```

This is causing problems with other parts of the code as well: 
```
Test executable failed (exit code 101).

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"docker.io/library/hello-world:latest"`,
 right: `"docker.io\\library/hello-world:latest"`', src\reference.rs:8:1
```

You can see an example run and output [here](https://github.com/deislabs/krustlet/runs/1316193432?check_suite_focus=true#step:4:1277)

itowlson:
Fixes #446

### DIFF
--- a/crates/oci-distribution/src/reference.rs
+++ b/crates/oci-distribution/src/reference.rs
@@ -1,7 +1,6 @@
 use std::convert::{Into, TryFrom};
 use std::error::Error;
 use std::fmt;
-use std::path::PathBuf;
 use std::str::FromStr;
 
 use crate::regexp;
@@ -90,10 +89,11 @@ impl Reference {
 
     /// full_name returns the full repository name and path.
     fn full_name(&self) -> String {
-        let mut path = PathBuf::new();
-        path.push(self.registry());
-        path.push(self.repository());
-        path.to_str().unwrap_or("").to_owned()
+        if self.registry() == "" {
+            format!("{}", self.repository())
+        } else {
+            format!("{}/{}", self.registry(), self.repository())
+        }
     }
 
     /// whole returns the whole reference.
@@ -273,6 +273,7 @@ mod test {
             assert_eq!(repository, reference.repository());
             assert_eq!(tag, reference.tag());
             assert_eq!(digest, reference.digest());
+            assert_eq!(input, reference.whole());
         }
 
         #[rstest(input, err,


crates/oci-distribution/src/reference.rs
```diff
@@ -90,10 +89,7 @@ impl Reference {
 
     /// full_name returns the full repository name and path.
     fn full_name(&self) -> String {
-        let mut path = PathBuf::new();
-        path.push(self.registry());
-        path.push(self.repository());
-        path.to_str().unwrap_or("").to_owned()
+        format!("{}/{}", self.registry(), self.repository())
```
bacongobbler, line 92: It is possible that `self.registry()` may return an empty string, which will result in a string like `/foo`. The previous code caught this scenario because an empty string would return `foo`.

One way to fix might be to check if `self.registry()` returns an empty string, and only join the two together if it exists.

It might be worth writing a unit test just to make sure I'm not totally wrong that the old code did indeed work as I described...

itowlson: Good catch - thanks!

itowlson: Fixed and test added - please take another look when convenient

bacongobbler (approved):

PR #457, opened 2020-11-25
fix workflow syntax

bacongobbler:
Arguments need to use the context and expression syntax rather than bash variables.

https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#contexts


kflansburg (approved):
LGTM

PR #460, opened 2020-11-26
Fix release action

kflansburg:
GitHub context needed to be used for env vars in a few other places.


bacongobbler:
Those areas should not be necessary as they’re running in a bash terminal. But it doesn’t hurt!

bacongobbler (approved):

bacongobbler:
Ahh... looking at the latest run it does look like this is necessary. Good catch.

Issue #447, opened 2020-11-10
Running Krustlet on Docker Desktop Mac OS X


david6983:
For my project, I would like to:
- run krustlet on Docker Desktop on mac os x without using minikube, microk8s or kind.
- Then, my goal is to run krustlet on a kubernetes cluster provided by rancher 2.0.

Any idea, how can easily install krustlet ?

bacongobbler:
Hi @david6983!

Have you had the chance to look through the HOWTO tutorials on running Krustlet using KinD, minikube, etc? Several of us are running and testing Krustlet on a macOS development environment and would be happy to help you out if you have any issues.

https://github.com/deislabs/krustlet/tree/master/docs

bacongobbler:
With regards to Rancher 2.0, I do not believe we've tested that scenario yet. I would suggest trying out one of the HOWTO guides first before attempting to venture off the beaten path.

david6983:
I am wondering why not using just Docker Desktop instead of all the solutions like microk8s, KinD, minikube and so on.

Docker Desktop is actually running a kubernetes cluster, why not just using it instead of using minikube for example ?

bacongobbler:
Docker Desktop should work as well.

If you're having issues with a particular section of the docs, please feel free to [reach out on Slack](https://github.com/deislabs/krustlet#community-discussion-contribution-and-support). Happy to help you out there and figure out what parts of the guide are unclear or need further clarification.

david6983:
in the krustlet-on-wsl2.md for example which look like the closest to a Docker Desktop Mac Os installation, I don't understand what are the .crt file (--cert-file) in this command

```
Since you are running locally, this step is important. Otherwise krustlet will pick up on your
# local config and not be able to update the node status properly
$ export KUBECONFIG=~/.krustlet/config/kubeconfig
$ krustlet-wasi --node-ip $mainIP --node-name krustlet --cert-file=~/.krustlet/config/krustlet.crt --private-key-file=~/.krustlet/config/krustlet.key --bootstrap-file=~/.krustlet/config/bootstrap.conf
```

I ran ./bootstrap.sh from krstulet directory but I don't have any .crt files

bacongobbler:
Okay, I think I see what you're pointing at now. I think that's a stale section of the documentation that was accidentally missed when we introduced https://github.com/deislabs/krustlet/pull/275. Feel free to disregard those flags and use the following instead:

```
krustlet-wasi --node-ip $mainIP --node-name krustlet
[...]

PR #461, opened 2020-11-27
Remove unused flags from howto guides

jiayihu:
Just finishing the removal started in #447 because I had the same issue because of the flags

### DIFF
--- a/docs/howto/krustlet-on-gke.md
+++ b/docs/howto/krustlet-on-gke.md
@@ -113,8 +113,6 @@ $ KUBECONFIG=${PWD}/kubeconfig krustlet-wasi \
 --hostname="krustlet" \
 --node-ip=${IP} \
 --node-name="krustlet" \
---cert-file=./krustlet.crt \
---private-key-file=./krustlet.key \
 --bootstrap-file=./bootstrap.conf
 ```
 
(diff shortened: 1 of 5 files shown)


itowlson:
@thomastaylor312 We still invoke using these flags in the justfile - should they stay in the docs or go as this PR suggests?

kflansburg:
This still controls *where* the certs are written by Krustlet during bootstrapping. The user was confused because the bootstrap script does not create these files and so they were not there prior to the first run of Krustlet. I agree that this is probably a little confusing to have in the beginner documentation. FWIW, I still use these all the time, as they are key if you plan on running bot wasi and wascc at the same time.

jiayihu:
For me it wasn't just confusing, it actually gave errors when launching the command as shown by another user in https://github.com/deislabs/krustlet/issues/447#issuecomment-724946954 and for a quick-start documentation it's not helpful at all. If it is still useful it should be left as note maybe.

kflansburg:
The error that you linked to seems to be due to it not finding bootstrap.conf, and not the certificates. It isnt clear to me why this happened.

thomastaylor312:
I think that taking these out of the beginner docs is good, but we should leave examples of them documented in other places

bacongobbler:
agree on ^^. Let's go ahead and merge this and follow up as needed. A topic page dedicated to krustlet's feature flags would be a good start.

PR #462, opened 2020-11-30
PVC mount support

### LINKED ISSUE #182: How should Krustlet handle the Container Storage Interface?

bacongobbler:
Adds support for mounting PersistentVolumeClaims to a Pod via the Container Storage Interface.

Closes #182

### DIFF
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -50,6 +50,7 @@ tokio  = { version = "0.2", features = ["fs", "stream", "macros", "signal", "uds
 kube = { version = "0.42", default-features = false }
 kube-runtime = { version= "0.42", default-features = false }
 k8s-openapi = { version = "0.9", default-features = false, features = ["v1_18"] }
+k8s-csi = "0.2"
 chrono = { version = "0.4", features = ["serde"] }
 structopt = { version = "0.3", features = ["wrap_help"], optional = true }
 hostname = "0.3"
@@ -63,6 +64,7 @@ rcgen = "0.8"
 uuid = { version = "0.8.1", features = ["v4"] }
 krator = { path = "../krator", version = "0.1", default-features = false }
 json-patch = "0.2"
+tempdir = "0.3"
 tonic = "0.3"
 # prost is needed for the files built by the protobuf
 prost = "0.6"
@@ -84,7 +86,6 @@ ws2_32-sys = "0.2.1"
 bytes = "0.3"
 env_logger = { version = "0.4", default-features = false }
 slab = "0.4"
-tempdir = "0.3"
 version-sync = "0.5"
 
 [dev-dependencies]
--- a/crates/kubelet/src/provider/mod.rs
+++ b/crates/kubelet/src/provider/mod.rs
@@ -5,11 +5,13 @@ use async_trait::async_trait;
 use k8s_openapi::api::core::v1::{ConfigMap, EnvVarSource, Secret};
 use kube::api::Api;
 use log::{error, info};
+use std::sync::Arc;
 use thiserror::Error;
 
 use crate::container::Container;
 use crate::log::Sender;
 use crate::node::Builder;
+use crate::plugin_watcher::PluginRegistry;
 use crate::pod::Pod;
 use crate::pod::Status as PodStatus;
 use krator::{ObjectState, State};
@@ -30,6 +32,7 @@ use krator::{ObjectState, State};
 /// # Example
 /// ```rust
 /// use async_trait::async_trait;
+/// use kubelet::plugin_watcher::PluginRegistry;
 /// use kubelet::pod::{Pod, Status};
 /// use kubelet::provider::Provider;
 /// use kubelet::pod::state::Stub;
@@ -58,9 +61,13 @@ use krator::{ObjectState, State};
 ///     const ARCH: &'static str = "my-arch";
 ///
 ///     type PodState = PodState;
-///    
+///
 ///     fn provider_state(&self) -> SharedState<ProviderState> {
 ///         Arc::new(RwLock::new(ProviderState {}))
+///     }
+///
+///     fn plugin_registry(&self) -> Option<Arc<PluginRegistry>> {
+///         Some(Arc::new(Default::default()))
 ///     }
 ///
 ///     async fn initialize_pod_state(&self, _pod: &Pod) -> anyhow::Result<Self::PodState> {
@@ -118,6 +125,11 @@ pub trait Provider: Sized + Send + Sync + 'static {
     /// not available. Override this only when there is an implementation.
     async fn exec(&self, _pod: Pod, _command: String) -> anyhow::Result<Vec<String>> {
         Err(NotImplementedError.into())
+    }
+
+    /// Fetch the CSI driver plugin registry.
+    fn plugin_registry(&self) -> Option<Arc<PluginRegistry>> {
+        None
     }
 
     /// Resolve the environment variables for a container.
--- a/crates/kubelet/src/volume/configmap.rs
+++ b/crates/kubelet/src/volume/configmap.rs
@@ -1,0 +1,42 @@
+use std::path::PathBuf;
+
+use k8s_openapi::api::core::v1::{ConfigMap, KeyToPath};
+
+use super::*;
+
+pub(crate) async fn populate(
+    config_map: ConfigMap,
+    path: &PathBuf,
+    items: &Option<Vec<KeyToPath>>,
+) -> anyhow::Result<VolumeType> {
+    tokio::fs::create_dir_all(path).await?;
+    let binary_data = config_map.binary_data.unwrap_or_default();
+    let binary_data = binary_data.into_iter().map(|(key, data)| async move {
+        match mount_setting_for(&key, items) {
+            ItemMount::MountAt(mount_path) => {
+                let file_path = path.join(mount_path);
+                tokio::fs::write(file_path, &data.0).await
+            }
+            ItemMount::DoNotMount => Ok(()),
+        }
+    });
+    let binary_data = futures::future::join_all(binary_data);
+    let data = config_map.data.unwrap_or_default();
+    let data = data.into_iter().map(|(key, data)| async move {
+        match mount_setting_for(&key, items) {
+            ItemMount::MountAt(mount_path) => {
+                let file_path = path.join(mount_path);
+                tokio::fs::write(file_path, data).await
+            }
+            ItemMount::DoNotMount => Ok(()),
+        }
+    });
+    let data = futures::future::join_all(data);
+    let (binary_data, data) = futures::future::join(binary_data, data).await;
+    binary_data
+        .into_iter()
+        .chain(data)
+        .collect::<tokio::io::Result<_>>()?;
+
+    Ok(VolumeType::ConfigMap)
+}
--- a/crates/kubelet/src/volume/mod.rs
+++ b/crates/kubelet/src/volume/mod.rs
@@ -1,41 +1,59 @@
-//! A module for use in managing volumes in providers. Use of this module is not mandatory to create
-//! a Provider, but it does provide common implementation logic for supported volume providers.
+//! A module for use in managing volumes in providers. Use of this module is not
+//! mandatory to create a Provider, but it does provide common implementation
+//! logic for supported volume providers.
 use std::collections::HashMap;
 use std::ops::Deref;
 use std::path::PathBuf;
+use std::sync::Arc;
 
-use k8s_openapi::api::core::v1::Volume as KubeVolume;
-use k8s_openapi::api::core::v1::{ConfigMap, KeyToPath, Secret};
-use k8s_openapi::ByteString;
+use tokio::task;
+
+use k8s_openapi::api::core::v1::KeyToPath;
+use k8s_openapi::api::core::v1::{ConfigMap, PersistentVolumeClaim, Secret, Volume as KubeVolume};
 use kube::api::Api;
 use log::{debug, error};
 
+use crate::plugin_watcher::PluginRegistry;
 use crate::pod::Pod;
 
+mod configmap;
+mod hostpath;
+mod persistentvolumeclaim;
+mod secret;
+
+/// type of volume
 #[derive(Debug)]
-enum Type {
+pub enum VolumeType {
+    /// configmap volume
     ConfigMap,
+    /// secret volume
     Secret,
+    /// PVC volume
+    PersistentVolumeClaim,
+    /// hostpath volume
     HostPath,
 }
 
-/// A smart wrapper around the location of a volume on the host system. If this is a ConfigMap or
-/// Secret volume, dropping this reference will clean up the temporary volume. [AsRef] and
-/// [std::ops::Deref] are implemented for this type so you can still use it like a normal PathBuf
+/// A smart wrapper around the location of a volume on the host system. If this
+/// is a ConfigMap or Secret volume, dropping this reference will clean up the
+/// temporary volume. [AsRef] and [std::ops::Deref] are implemented for this
+/// type so you can still use it like a normal PathBuf
 #[derive(Debug)]
 pub struct Ref {
     host_path: PathBuf,
-    volume_type: Type,
+    volume_type: VolumeType,
 }
 
 impl Ref {
-    /// Resolves the volumes for a pod, including preparing temporary directories containing the
-    /// contents of secrets and configmaps. Returns a HashMap of volume names to a PathBuf for the
-    /// directory where the volume is mounted
+    /// Resolves the volumes for a pod, including preparing temporary
+    /// directories containing the contents of secrets and configmaps. Returns a
+    /// HashMap of volume names to a PathBuf for the directory where the volume
+    /// is mounted
     pub async fn volumes_from_pod(
         volume_dir: &PathBuf,
         pod: &Pod,
         client: &kube::Client,
+        plugin_registry: Option<Arc<PluginRegistry>>,
     ) -> anyhow::Result<HashMap<String, Self>> {
         let base_path = volume_dir.join(pod_dir_name(pod));
         tokio::fs::create_dir_all(&base_path).await?;
@@ -43,12 +61,14 @@ impl Ref {
             let volumes = vols.iter().map(|v| {
                 let mut host_path = base_path.clone();
                 host_path.push(&v.name);
+                let pr = plugin_registry.clone();
                 async move {
-                    let volume_type = configure(v, pod.namespace(), client, &host_path).await?;
+                    let volume_type = configure(v, pod.namespace(), client, pr, &host_path).await?;
                     Ok((
                         v.name.to_owned(),
-                        // Every other volume type should mount to the given host_path except for a
-                        // hostpath volume type. So we need to handle that special case here
+                        // Every other volume type should mount to the given
+                        // host_path except for a hostpath volume type. So we
+                        // need to handle that special case here
                         match &v.host_path {
                             Some(hostpath) => Ref {
                                 host_path: PathBuf::from(&hostpath.path),
@@ -70,6 +90,33 @@ impl Ref {
             Ok(HashMap::default())
         }
     }
+
+    /// Unmounts any volumes mounted to the pod. Usually called when dropping
+    /// the pod out of scope.
+    pub async fn unmount_volumes_from_pod(
+        volume_dir: &PathBuf,
+        pod: &Pod,
+        client: &kube::Client,
+        plugin_registry: Option<Arc<PluginRegistry>>,
+    ) -> anyhow::Result<()> {
+        if let Some(vols) = pod.volumes() {
+            let base_path = volume_dir.join(pod_dir_name(pod));
+            for vol in vols {
+                if let Some(pvc_source) = &vol.persistent_volume_claim {
+                    let vol_path = base_path.join(&vol.name);
+                    persistentvolumeclaim::unpopulate(
+                        pvc_source,
+                        client,
+                        pod.namespace(),
+                        plugin_registry.clone(),
+                        &vol_path,
+                    )
+                    .await?;
+                }
+            }
+        }
+        Ok(())
+    }
 }
 
 impl AsRef<PathBuf> for Ref {
@@ -88,8 +135,9 @@ impl Deref for Ref {
 
 impl Drop for Ref {
     fn drop(&mut self) {
-        if matches!(self.volume_type, Type::ConfigMap | Type::Secret) {
-            // TODO: Currently there is no way to do this async (though there is an async destructors proposal)
+        if matches!(self.volume_type, VolumeType::ConfigMap | VolumeType::Secret) {
+            // TODO: Currently there is no way to do this async (though there is
+            // an async destructors proposal)
             debug!(
                 "deleting {:?} directory {:?}",
                 self.volume_type, self.host_path
@@ -102,116 +150,6 @@ impl Drop for Ref {
             });
         }
     }
-}
-
-/// This is a gnarly function to check all of the supported data members of the Volume struct.
-/// Because it isn't a HashMap, we need to check all fields individually
-async fn configure(
-    vol: &KubeVolume,
-    namespace: &str,
-    client: &kube::Client,
-    path: &PathBuf,
-) -> anyhow::Result<Type> {
-    if let Some(cm) = &vol.config_map {
-        populate_from_config_map(
-            &cm.name
-                .as_ref()
-                .ok_or_else(|| anyhow::anyhow!("no configmap name was given"))?,
-            namespace,
-            client,
-            path,
-            &cm.items,
-        )
-        .await
-    } else if let Some(s) = &vol.secret {
-        populate_from_secret(
-            &s.secret_name
-                .as_ref()
-                .ok_or_else(|| anyhow::anyhow!("no secret name was given"))?,
-            namespace,
-            client,
-            path,
-            &s.items,
-        )
-        .await
-    } else if let Some(hostpath) = &vol.host_path {
-        // Check the the directory exists on the host
-        tokio::fs::metadata(&hostpath.path).await?;
-        Ok(Type::HostPath)
-    } else {
-        Err(anyhow::anyhow!(
-            "Unsupported volume type. Currently supported types: ConfigMap, Secret, and HostPath"
-        ))
-    }
-}
-
-async fn populate_from_secret(
-    name: &str,
-    namespace: &str,
-    client: &kube::Client,
-    path: &PathBuf,
-    items: &Option<Vec<KeyToPath>>,
-) -> anyhow::Result<Type> {
-    tokio::fs::create_dir_all(path).await?;
-    let secret_client: Api<Secret> = Api::namespaced(client.clone(), namespace);
-    let secret = secret_client.get(name).await?;
-    let data = secret.data.unwrap_or_default();
-    let data = data.iter().map(|(key, ByteString(data))| async move {
-        match mount_setting_for(key, items) {
-            ItemMount::MountAt(mount_path) => {
-                let file_path = path.join(mount_path);
-                tokio::fs::write(file_path, &data).await
-            }
-            ItemMount::DoNotMount => Ok(()),
-        }
-    });
-    futures::future::join_all(data)
-        .await
-        .into_iter()
-        .collect::<tokio::io::Result<_>>()?;
-
-    Ok(Type::Secret)
-}
-
-async fn populate_from_config_map(
-    name: &str,
-    namespace: &str,
-    client: &kube::Client,
-    path: &PathBuf,
-    items: &Option<Vec<KeyToPath>>,
-) -> anyhow::Result<Type> {
-    tokio::fs::create_dir_all(path).await?;
-    let cm_client: Api<ConfigMap> = Api::namespaced(client.clone(), namespace);
-    let config_map = cm_client.get(name).await?;
-    let binary_data = config_map.binary_data.unwrap_or_default();
-    let binary_data = binary_data.iter().map(|(key, data)| async move {
-        match mount_setting_for(key, items) {
-            ItemMount::MountAt(mount_path) => {
-                let file_path = path.join(mount_path);
-                tokio::fs::write(file_path, &data.0).await
-            }
-            ItemMount::DoNotMount => Ok(()),
-        }
-    });
-    let binary_data = futures::future::join_all(binary_data);
-    let data = config_map.data.unwrap_or_default();
-    let data = data.iter().map(|(key, data)| async move {
-        match mount_setting_for(key, items) {
-            ItemMount::MountAt(mount_path) => {
-                let file_path = path.join(mount_path);
-                tokio::fs::write(file_path, data).await
-            }
-            ItemMount::DoNotMount => Ok(()),
-        }
-    });
-    let data = futures::future::join_all(data);
-    let (binary_data, data) = futures::future::join(binary_data, data).await;
-    binary_data
-        .into_iter()
-        .chain(data)
-        .collect::<tokio::io::Result<_>>()?;
-
-    Ok(Type::ConfigMap)
 }
 
 fn pod_dir_name(pod: &Pod) -> String {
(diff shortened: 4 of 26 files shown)


bacongobbler:
I got everything working here according to the CSI examples. I DO NOT expect anyone else to be able to reproduce the same environment, but if you want to try giving it a whirl... Here goes:

1. compile the following projects from HEAD, and move them into your PATH: 
   a. https://github.com/kubernetes-csi/csi-driver-host-path
   b. https://github.com/kubernetes-csi/node-driver-registrar
   c. https://github.com/kubernetes-csi/external-provisioner

2. In separate terminals spawn the following:

```
# hostpathplugin will need root permissions to execute 'mount'
./hostpathplugin --nodeid ultralisk --endpoint unix://csi/csi.sock -v 5
./csi-node-driver-registrar --csi-address /csi/csi.sock --kubelet-registration-path /csi/csi.sock --plugin-registration-path ~/.krustlet/plugins
./csi-provisioner --csi-address /csi/csi.sock --kubeconfig ~/.kube/config
```

3. Start Krustlet. Approve the bootstrap TLS certificate.
4. Create the storage class: `kubectl create -f https://raw.githubusercontent.com/kubernetes-csi/csi-driver-host-path/master/examples/csi-storageclass.yaml`
5. Create the PVC: `kubectl create -f https://raw.githubusercontent.com/kubernetes-csi/csi-driver-host-path/master/examples/csi-pvc.yaml`
6. Create the modified `k8s.yaml` example in `demos/wasi`

bacongobbler:
Note: this is blocked until https://github.com/kflansburg/k8s-csi/pull/1 is merged and a new version of k8s-csi is released, as it will not compile until it's been re-compiled against tonic 0.3. I've pointed Krustlet at a temporary fork for the time being.

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+        // TODO: determine the correct access mode and mount flags from the volume
+        // https://github.com/kubernetes/kubernetes/blob/734889ed822d1a60c6dd61ccd8f1ed0e8ab31ea5/pkg/volume/csi/csi_attacher.go#L325-L333
+        readonly: false,
+        secrets: Default::default(),
+        publish_context: Default::default(),
+        volume_context: Default::default(),
+    };
+
+    if stage_unstage_volume {
+        debug!("stage/unstage supported");
```
thomastaylor312, line 290: Can you add more context here?

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+                secrets: Default::default(),
+                publish_context: Default::default(),
+                volume_context: Default::default(),
+            })
+            .await?;
+    }
+
+    csi_client.node_publish_volume(req).await?;
+
+    debug!("done");
```
thomastaylor312, line 317: Same with more context here

crates/kubelet/src/volume/secret.rs
```diff
@@ -0,0 +1,30 @@
...
+use super::*;
+
+pub(crate) async fn populate(
+    secret: Secret,
+    path: &PathBuf,
+    items: &Option<Vec<KeyToPath>>,
+) -> anyhow::Result<Type> {
+    tokio::fs::create_dir_all(path).await?;
+    let data = secret.data.unwrap_or_default();
+    let data = data.iter().map(|(key, ByteString(data))| async move {
```
thomastaylor312, line 15: Nit: I know this is copied code, but if you feel so inclined, could you change this to `into_iter`? We are consuming the data anyway

demos/wasi/hello-world-rust/k8s.yaml
```diff
@@ -28,18 +13,13 @@ spec:
...
-          valueFrom:
-            configMapKeyRef:
-              key: myval
-              name: hello-world-wasi-rust
       volumeMounts:
         - name: storage
           mountPath: /mnt/storage
   volumes:
     - name: storage
-      configMap:
```
thomastaylor312, line 41: We should probably do a separate example using the same hello world but with PVC

thomastaylor312 (changes requested):
This is looking pretty good! We definitely should document all the steps required (basically at least the components needed) and a bug "THIS IS VERY ALPHA" label somewhere

crates/kubelet/src/provider/mod.rs
```diff
@@ -121,6 +128,9 @@ pub trait Provider: Sized + Send + Sync + 'static {
         Err(NotImplementedError.into())
     }
 
+    /// Fetch the CSI driver plugin registry.
+    fn plugin_registry(&self) -> Arc<PluginRegistry>;
```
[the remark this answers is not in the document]
thomastaylor312: After looking at the rest of the code, I'm downgrading this to a nit

crates/kubelet/src/volume/mod.rs
```diff
@@ -243,3 +158,78 @@ impl From<Option<String>> for ItemMount {
...
+    client: &kube::Client,
+    plugin_registry: Arc<PluginRegistry>,
+    path: &PathBuf,
+) -> anyhow::Result<Type> {
+    if let Some(cm) = &vol.config_map {
+        let name = &cm
+            .name
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("no configmap name was given"))?;
+        let cm_client: Api<ConfigMap> = Api::namespaced(client.clone(), namespace);
```
kflansburg, line 177: Should fetching things from the API be moved into the respective modules?

crates/kubelet/src/volume/mod.rs
```diff
@@ -243,3 +158,78 @@ impl From<Option<String>> for ItemMount {
...
+            // revisit this assumption (omission of this field implies the
+            // default).
+            //
+            // https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1
+            .get(spec.storage_class_name.as_ref().unwrap())
+            .await?;
+        debug!("found a storage class");
+        // TODO(bacongobbler): use something a bit more gentle than `.expect`
+        let volume_name = spec.volume_name.as_ref().expect("volume name must exist");
+        // TODO(bacongobbler): When a PVC specifies a selector in addition to
```
kflansburg, line 219: Isnt this handled by k8s when the PVC is bound?

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+    type Err = BadVolumeMode;
+
+    // defines what type of volume is required by the claim. The "filesystem"
+    // mode is implied when not included in the claim spec.
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Block" => Ok(VolumeMode::Block),
+            "Filesystem" => Ok(VolumeMode::Filesystem),
+            "" => Ok(VolumeMode::Filesystem),
+            _ => Err(BadVolumeMode),
```
kflansburg, line 79: Is it possible for these to be created with the other values? Is there value in creating these new types rather than just doing `anyhow::bail` with some context?

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+enum AccessMode {
+    ReadOnlyMany,
+    ReadWriteMany,
+    ReadWriteOnce,
+}
+
+impl FromStr for AccessMode {
+    type Err = BadAccessMode;
+
+    // defines what type of reclaim policy is requested. The "delete" mode is
```
kflansburg, line 147: This looks like copypasta

kflansburg (commented):
I don't understand this enough to approve but had some comments.

itowlson (changes requested):
I have basically no idea whether this is right or wrong, but I sure do have a lot of opinions about how to organise code I don't understand!

crates/kubelet/src/volume/mod.rs
```diff
@@ -243,3 +158,78 @@ impl From<Option<String>> for ItemMount {
...
+    } else if let Some(s) = &vol.secret {
+        let name = &s
+            .secret_name
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("no secret name was given"))?;
+        let secret_client: Api<Secret> = Api::namespaced(client.clone(), namespace);
+        let secret = secret_client.get(name).await?;
+        secret::populate(secret, path, &s.items).await
+    } else if let Some(pvc_source) = &vol.persistent_volume_claim {
+        debug!("found a PVC source");
```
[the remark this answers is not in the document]
thomastaylor312: Most likely, but that could be captured as another issue (we should improve logging in a couple places

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    if let Some(selector) = &spec.selector {
+        validate_label_selector(selector)?;
+    }
+
+    match &spec.storage_class_name {
+        // no storage class name specified or set to the empty string
+        // means the user requested the "default" storage class. there
+        // is no "default" storage class for Krustlet at this time, so
```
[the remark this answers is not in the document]
thomastaylor312: You have to specify a default storage class at the cluster level, so the admission controller will handle it. But I actually think we'd need this error in the case we do get an empty storage class name

crates/kubelet/src/provider/mod.rs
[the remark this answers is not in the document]
bacongobbler: I like that idea @kflansburg.

I'm curious how we should handle cases where a PVC is requested but the provider doesn't support CSI plugins though. Should we error out and say that the pod cannot be scheduled? Trying to determine what is the core criteria for being a Provider here.

crates/kubelet/src/volume/mod.rs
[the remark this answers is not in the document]
bacongobbler: Oh these are just your typical garden variety debug messages. I'll remove them

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+            "ReadWriteMany" => Ok(AccessMode::ReadWriteMany),
+            "ReadWriteOnce" => Ok(AccessMode::ReadWriteOnce),
+            _ => Err(BadAccessMode),
+        }
+    }
+}
+
+// Validates a PersistentVolumeClaimSpec.
+// https://github.com/kubernetes/kubernetes/blob/c970a46bc1bcc100bbbfabd5c12bd4c5d87f8aea/pkg/apis/core/validation/validation.go#L1965
+pub(crate) fn validate(spec: &PersistentVolumeClaimSpec) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
bacongobbler: Yes. This validation is performed by the Kubernetes API, so we need to validate and sanitize this metadata as well.

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+pub(crate) async fn populate(
+    pv: PersistentVolume,
+    storage_class: StorageClass,
+    plugin_registry: Arc<PluginRegistry>,
+    path: &PathBuf,
+) -> anyhow::Result<Type> {
+    let endpoint = plugin_registry
+        .get_endpoint(&storage_class.provisioner)
+        .await
+        .expect("could not get CSI plugin endpoint");
```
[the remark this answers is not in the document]
bacongobbler: Yes absolutely.

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+    let csi = pv.spec.unwrap().csi.unwrap();
+    let data = serde_json::json!({
+        "volumeHandle": csi.volume_handle,
+        "driverName": csi.driver,
+    });
+
+    // save volume data to path
+    // https://github.com/kubernetes/kubernetes/blob/734889ed822d1a60c6dd61ccd8f1ed0e8ab31ea5/pkg/volume/csi/csi_attacher.go#L294-L305
+    let mut vol_data_filepath = path.clone();
+    vol_data_filepath.push("../".to_owned() + VOLUME_DATA_FILE_NAME);
```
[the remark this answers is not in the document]
bacongobbler: Somehow I missed this but yeah doing a `.join` would be muuuuuch better.

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+                }
+            }
+        }
+    }
+
+    // we keep this around even if the driver does not support STAGE_UNSTAGE_VOLUME. unmount() still needs it.
+    tokio::fs::create_dir_all(path).await?;
+
+    // https://github.com/kubernetes/kubernetes/blob/734889ed822d1a60c6dd61ccd8f1ed0e8ab31ea5/pkg/volume/csi/csi_attacher.go#L295-L298
+    let csi = pv.spec.unwrap().csi.unwrap();
```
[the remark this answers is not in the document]
bacongobbler: Not really, no. We could if we implemented the validation logic as above but at this point the call to `unwrap()` is totally unsafe. I'll see what I can do about this.

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+    });
+
+    // save volume data to path
+    // https://github.com/kubernetes/kubernetes/blob/734889ed822d1a60c6dd61ccd8f1ed0e8ab31ea5/pkg/volume/csi/csi_attacher.go#L294-L305
+    let mut vol_data_filepath = path.clone();
+    vol_data_filepath.push("../".to_owned() + VOLUME_DATA_FILE_NAME);
+    tokio::fs::write(&vol_data_filepath, serde_json::to_string(&data)?).await?;
+
+    // https://github.com/kflansburg/k8s-csi/blob/85a1e81757332724f5e23b452fdb2da92f0c1923/proto/csi/v1.3.0/api.proto#L1189
+    let staging_path = path.parent().unwrap().join("_staging");
```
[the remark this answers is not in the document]
bacongobbler: We need this for the call to `publish` as well for some reason... Dunno why but it's a required field in the gRPC call.

crates/kubelet/src/volume/persistentvolumeclaim.rs
bacongobbler: garden variety debug statement; will remove

crates/kubelet/src/volume/persistentvolumeclaim.rs
bacongobbler: garden variety debug statement; will remove

crates/kubelet/src/volume/secret.rs
bacongobbler: yep!

demos/wasi/hello-world-rust/k8s.yaml
bacongobbler: agreed. Might need to point to the documentation on setting up the example too, but yes.

crates/kubelet/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio  = { version = "0.2", features = ["fs", "stream", "macros", "signal"] }
 kube = { version = "0.42", default-features = false }
 kube-runtime = { version= "0.42", default-features = false }
 k8s-openapi = { version = "0.9", default-features = false, features = ["v1_18"] }
+k8s-csi = { git = "https://github.com/bacongobbler/k8s-csi" }
```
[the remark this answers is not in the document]
bacongobbler: yes!

crates/kubelet/src/volume/mod.rs
bacongobbler: If I can recall from memory... The volume manager code selects and determines the correct storage class. This is performed at runtime. Kubernetes does not handle this for the Kubelet.

I'll see if I can find the relevant code bits again and drop a doc comment to give some more context around this.

crates/kubelet/src/volume/mod.rs
```diff
@@ -1,54 +1,75 @@
...
 
+mod configmap;
+mod hostpath;
+mod persistentvolumeclaim;
+mod secret;
+
+/// type of volume
 #[derive(Debug)]
-enum Type {
+pub enum Type {
```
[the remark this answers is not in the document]
bacongobbler: oh I totally agree here.

crates/kubelet/src/provider/mod.rs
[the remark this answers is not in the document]
thomastaylor312: Yeah, I think we'd just error the pod in a case where pvcs are not supported and one is requested

crates/kubelet/src/volume/mod.rs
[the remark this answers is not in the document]
bacongobbler: Though to be fair in the future I am planning on refactoring all of this out as part of #482. I still think this is a great idea in the interim however

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,319 @@
...
+            "Block" => Ok(VolumeMode::Block),
+            "Filesystem" => Ok(VolumeMode::Filesystem),
+            "" => Ok(VolumeMode::Filesystem),
+            _ => Err(BadVolumeMode),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct BadReclaimPolicy;
```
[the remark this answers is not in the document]
bacongobbler: Is there a good example of this somewhere I can use as a reference guide?

crates/kubelet/src/provider/mod.rs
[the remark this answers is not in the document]
kflansburg: Yeah, I believe there are some well known errors / behavior here, you get a reason `FailedMount` and an error like `Failed to mount x for pod: CSI driver support not implemented.`. The only tricky part I would think is how to interrupt the state machine from where this is called. Not sure about that.

crates/kubelet/src/volume/persistentvolumeclaim.rs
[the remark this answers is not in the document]
thomastaylor312: Take a look at the [bindle errors](https://github.com/deislabs/bindle/blob/b51ae14f26be2a647ceffc38ace26f3290dfacae/src/provider/mod.rs#L119) for an example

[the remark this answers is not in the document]
bacongobbler: Thank you! I'll have a look at this today.

crates/kubelet/src/volume/mod.rs
bacongobbler: I'd prefer that the functions deal with the concrete type rather than all the kube client fetching and whatnot. That way the transition from this design to #482 will be MUCH easier.

That being said let's go forward with this design for now and see if we need to revisit it after we start tackling #482. Either way this is all private functions so it doesn't really affect the public API at this point :)

crates/kubelet/src/volume/persistentvolumeclaim.rs
bacongobbler: Is there something I should address here?

crates/kubelet/src/volume/configmap.rs
```diff
@@ -0,0 +1,42 @@
...
+    let data = data.iter().map(|(key, data)| async move {
+        match mount_setting_for(key, items) {
+            ItemMount::MountAt(mount_path) => {
+                let file_path = path.join(mount_path);
+                tokio::fs::write(file_path, data).await
+            }
+            ItemMount::DoNotMount => Ok(()),
+        }
+    });
+    let data = futures::future::join_all(data);
```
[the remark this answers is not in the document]
bacongobbler: I do agree with your comment here. While I didn't write this code as part of the PR, I'll see what I can do to improve the code's readability.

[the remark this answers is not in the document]
bacongobbler: I tried to tackle this, but my failure to understand generics fails me once again. `configmap.binary_data` returns a `BTreeMap<String, ByteString>` whereas `configmap.data` returns a `BTreeMap<String, String>`. So a function signature like `async fn write_all(data: BTreeMap<String, String>) -> anyhow::Result<()>` will not work in this context.

Any ideas?

crates/kubelet/src/volume/mod.rs
```diff
@@ -243,3 +158,78 @@ impl From<Option<String>> for ItemMount {
...
+        let volume_name = spec.volume_name.as_ref().expect("volume name must exist");
+        // TODO(bacongobbler): When a PVC specifies a selector in addition to
+        // requesting a StorageClass, the requirements are ANDed together: only
+        // a PV of the requested class and with the requested labels may be
+        // bound to the PVC.
+        // https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1
+        let pv = pv_client.get(&volume_name).await?;
+        debug!("found a PV");
+
+        persistentvolumeclaim::populate(pv, storage_class, plugin_registry, path).await
```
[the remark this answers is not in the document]
bacongobbler: hmm... @kflansburg raised this concern in another comment as well.

I do agree with both your feedback. However, I do intend to refactor this some more in the next release (#482 should help a lot with this). Is it all right if we move forward here without addressing this feedback at this time, or would you prefer that I address it now before we merge?

crates/kubelet/src/volume/persistentvolumeclaim.rs
bacongobbler: For context, yes this is copypasta from the rust docs. Reference implementation: https://doc.rust-lang.org/std/str/trait.FromStr.html

crates/kubelet/src/volume/configmap.rs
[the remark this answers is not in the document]
itowlson: `async fn write_all<T,C: AsRef<[u8]> + Unpin>(data: BTreeMap<String, T>, accessor: (&T -> C)) -> anyhow::Result<()>` maybe?  The `C` is the input type to `tokio::fs::write`; the `T` is the type of the k8s configmap entry values; and `acccessor` is the extractor function that takes you from T to C (the identity for String, the `.0` accessor for ByteString).

If it doesn't fall out fairly easily then I would certainly punt on it - it's not a biggie.

crates/kubelet/src/volume/mod.rs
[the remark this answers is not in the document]
itowlson: Ship it as is :shipit: and I'll just watch #482 like a hawk 🐦 and yes I am aware that looks more like a parrot

crates/kubelet/src/volume/persistentvolumeclaim.rs
itowlson: I think the issue might be `s/reclaim policy/access mode`?

kflansburg: Yep, it seemed like it was from another struct.

bacongobbler: :headsmack: I see that now. Will address tomorrow :)

crates/kubelet/src/volume/configmap.rs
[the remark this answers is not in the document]
thomastaylor312: I actually think you could just do `async fn write_all<T: AsRef<[u8]>>(data: BTreeMap<String, T>)` as both `String` and `ByteString` implement `AsRef<[u8]>`

crates/kubelet/src/volume/configmap.rs
```diff
@@ -0,0 +1,42 @@
...
+use super::*;
+
+pub(crate) async fn populate(
+    config_map: ConfigMap,
+    path: &PathBuf,
+    items: &Option<Vec<KeyToPath>>,
+) -> anyhow::Result<VolumeType> {
+    tokio::fs::create_dir_all(path).await?;
+    let binary_data = config_map.binary_data.unwrap_or_default();
+    let binary_data = binary_data.iter().map(|(key, data)| async move {
```
thomastaylor312, line 14: Nit: We are consuming (and moving) the data here, so `into_iter` is a better fit

crates/kubelet/src/volume/mod.rs
```diff
@@ -88,8 +109,12 @@ impl Deref for Ref {
 
 impl Drop for Ref {
     fn drop(&mut self) {
-        if matches!(self.volume_type, Type::ConfigMap | Type::Secret) {
-            // TODO: Currently there is no way to do this async (though there is an async destructors proposal)
+        if matches!(
+            self.volume_type,
+            VolumeType::ConfigMap | VolumeType::Secret | VolumeType::PersistentVolumeClaim
```
thomastaylor312, line 114: I missed this the first time. I actually don't think we can let this clean up the persistent volume because otherwise it will be removing the data from the PV. The PV needs to be unmounted before we clean up the directory. I am guessing the volume mounter takes care of this cleanup and even if it doesn't, we can make it do it later when we try to port it to Rust. For now I am fine leaving an orphaned directory if that is the case

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+async fn publish_volume(
+    csi_client: &mut NodeClient<tonic::transport::Channel>,
+    csi: &CSIPersistentVolumeSource,
+    staging_path: &Path,
+    stage_unstage_volume: bool,
+    path: &PathBuf,
+) -> anyhow::Result<()> {
+    let mut req = NodePublishVolumeRequest {
```
thomastaylor312, line 295: Are we supposed to be handling the `NodeUnpublishVolume` request on pod termination? I don't think we are doing that in the providers and so we are leaving volumes published to a node when a pod is no longer there, which will cause errors when they get reused

docs/howto/csi.md
```diff
@@ -0,0 +1,101 @@
...
+available for use:
+
+- [node-driver-registrar](https://github.com/kubernetes-csi/node-driver-registrar)
+- [external-provisioner](https://github.com/kubernetes-csi/external-provisioner)
+
+Do keep in mind that these projects rely on linux-specific command line tooling
+like `mount`, so these tools may only work on Linux. MacOS and Windows support
+is not guaranteed.
+
+These projects will eventually be merged into the Krustlet project in future
```
thomastaylor312, line 23: Only external provisioner will be I think. The registrar is something that should be built into CSI implementations or just leverage the existing project

thomastaylor312 (changes requested):
Docs look really good. I realized we were missing some things this second go around

crates/kubelet/src/volume/mod.rs
bacongobbler: Fair enough. And I think leaving an orphaned directory is fine since it's a `mount`/`umount` anyways

docs/howto/csi.md
thomastaylor312: So after investigating more, I don't think we have any immediate plans to port these over. Looks like external-provisioner is not linux specific. The node-driver-registrar is OS specific and we probably need to contribute back to it so it will work on Mac (which is a really easy fix)

thomastaylor312: Opened a PR to allow for building on Darwin: https://github.com/kubernetes-csi/node-driver-registrar/pull/133

crates/kubelet/src/volume/configmap.rs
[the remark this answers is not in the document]
bacongobbler: I'm having a really tough time cranking this one out in a timely period. The function signatures are good, but it's the return type on the closure that's throwing me a curve ball.

This is what it looks like right now:

```
async fn write_all<T,F>(data: BTreeMap<String, T>, map_fn: F) -> anyhow::Result<()>
where
    T: AsRef<[u8]> + Unpin,
    F: FnMut(String, T) -> Map<Iter<String, T>, ???>,
{}
```

Reason being is that it's a closure consuming `items` and `path`, so you have to pass in a closure. And I need to take a deeper dive at how `map` actually works to determine the correct return type for F.

I'll file an issue for now so we can track this for later.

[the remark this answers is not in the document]
bacongobbler: https://github.com/deislabs/krustlet/issues/503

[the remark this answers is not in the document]
itowlson: No worries.  Sorry for sending you down a dead end.

crates/kubelet/src/volume/persistentvolumeclaim.rs
bacongobbler: Yeah... I think we need that call. Where would be the best place to implement that call?

bacongobbler: Looking into implementing this in the `Terminated` state, just before calling `state_reader.stop()`.

thomastaylor312: @bacongobbler I was actually thinking it might be better to hook something in when the state machine terminates or to introduce a new common state called `volume cleanup`

kflansburg: IMO we should try not to add this to the state machine. It would be preferable that providers which do not use the common states do not have to know / remember to re-implement this.

bacongobbler: I hooked this into the `Terminated` state for the time being. Love you both but I cannot tell which direction y'all want me to take 😂 

Can you two please converse and let me know what your decision is after working through the design?

The nice thing is that I was able to refactor a bit of the volume package. Most of it should easily hook into whichever system y'all want me to write.

thomastaylor312: After talking this over offline, we are going to stick with having the cleanup be in the `Terminated` state for now. However, we realized that Krator needs some sort of a deregistration hook and that has been captured in #504

bacongobbler: Okay! I added a note in the code along with a link to the ticket. Thanks everyone.

thomastaylor312:
Looks like the doc tests are failing @bacongobbler. I also am seeing a lot of clippy errors, but I can fix those later

bacongobbler:
Oh yeah I see that. Never run the doc tests locally... Will fix those right up!

Thanks for catching the clippy errors as well.

thomastaylor312:
One last doc add I just thought of. In some clusters, the `system:nodes` group doesn't have access to read `StorageClass` objects. So if the Pod is erroring saying that the kubelet doesn't have access, you'll have to create the following cluster role and role binding to allow the nodes group to access them:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: storageclass-reader
rules:
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses"]
    verbs: ["get", "watch", "list"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: node-storageclass-reader
subjects:
  - kind: Group
    name: system:nodes
    apiGroup: rbac.authorization.k8s.io
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: storageclass-reader
```

bacongobbler:
Good call. I just added it to the HOWTO doc as an addendum. Let me know if you'd like me to make further changes/tweaks there.

thomastaylor312 (approved):

bacongobbler:
Thanks @thomastaylor312!

@itowlson are there any further changes you'd like me to address here?

itowlson (approved):
LGTM

Issue #469, opened 2020-12-29
Cannot pull from Google Artifact Registry and no update on pull state


jiayihu:
I have the hello-wasi example working if the image is pulled from GCP Registry but not from the new Artifact Registry. From logs, I can see that it panics here at line 152 with unexpected `None`:

https://github.com/deislabs/krustlet/blob/0a10d3f1e0c6ae9a2e00f61271db8c819321e465/crates/oci-distribution/src/client.rs#L143-L152

The pod YAML is just:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: hello-wasi
spec:
  containers:
    - name: hello-wasi
      image: europe-west1-docker.pkg.dev/fedra-thesis/images/hello-wasi:latest
    # (working)  image: eu.gcr.io/fedra-thesis/hello-wasi:latest
  nodeSelector:
    kubernetes.io/arch: "wasm32-wasi"
  tolerations:
    - key: "kubernetes.io/arch"
      operator: "Equal"
      value: "wasm32-wasi"
      effect: "NoExecute"
    - key: "kubernetes.io/arch"
      operator: "Equal"
      value: "wasm32-wasi"
      effect: "NoSchedule"
```

You can try it yourself to see that the pod remains in `Pending` status. I'll leave the image public, let me know if I delete it by mistake.  FYI `wasm-to-oci` is able to pull the image without issues.

Krustlet should also post updates about the image pull status, it wasn't intuitive debugging the cause of the eternal pending. The `describe` command only stated the following.

```
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  19m   default-scheduler  Successfully assigned default/hello-wasi to rasp-4
```

The usual kubelet is much richer with pull events as you know:

```
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  10s               default-scheduler  Successfully assigned default/hello-world to rasp-2
  Normal   Pulled     6s                kubelet            Successfully pulled image "hello-world" in 4.094783929s
  Normal   Pulling    5s (x2 over 10s)  kubelet            Pulling image "hello-world"
  Normal   Pulled     3s                kubelet            Successfully pulled image "hello-world" in 1.714372192s
  Normal   Created    3s (x2 over 6s)   kubelet            Created container hello-world
  Normal   Started    3s (x2 over 5s)   kubelet            Started container hello-world
[...]

PR #474, opened 2021-01-07
Check if service field is provided by auth challenge

jiayihu:
This solves #469 It should be a no brainer.

Trying to request the image from Google Artifacts returns `Bearer realm="https://europe-west1-docker.pkg.dev/v2/token"` without the `service` field. The [Distribution Spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md) doesn't tell anything about authentication AFAIK.

I found about the service field in the [Docker Registry docs](https://docs.docker.com/registry/configuration/#silly), so it's probably not needed anymore in Artifacts?

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -266,15 +266,21 @@ impl Client {
 
         let challenge = &challenge_opt[0];
         let realm = challenge.realm.as_ref().unwrap();
-        let service = challenge.service.as_ref().unwrap();
+        let service = challenge.service.as_ref();
+        let mut query = vec![("scope", &scope)];
+
+        if service.is_some() {
+            query.push(("service", service.unwrap()))
+        }
 
         // TODO: At some point in the future, we should support sending a secret to the
         // server for auth. This particular workflow is for read-only public auth.
         debug!("Making authentication call to {}", realm);
+
         let auth_res = self
             .client
             .get(realm)
-            .query(&[("service", service), ("scope", &scope)])
+            .query(&query)
             .apply_authentication(authentication)
             .send()
             .await?;


bacongobbler:
Hi @jiayihu! Would you mind writing a unit test to cover this test case? That way we can ensure this scenario is tested and accounted for in future iterations. Thanks!

jiayihu:
Sure, but I think that would need me to add an image url here: 

https://github.com/deislabs/krustlet/blob/592bb8d20cf729fdd1e2036f5fe1389d58a27336/crates/oci-distribution/src/client.rs#L823-L835

I think would be better if one of you maintainers could upload the `hello-wasm` image to Google Artifacts. Then I'll add the test 🤔

thomastaylor312:
@bacongobbler I'm not sure if we can easily test this due to the need for a google artifacts registry, unless there is one sitting around someone will let us use. However, that would also make us dependent on whoever owns that registry

bacongobbler:
I figured there should be a way to set up a mock authentication server where we could introspect the request data, but that seems out of scope for this PR.

thomastaylor312 (approved):

thomastaylor312:
Thanks for the fix on this @jiayihu!

crates/oci-distribution/src/client.rs
```diff
@@ -266,15 +266,21 @@ impl Client {
 
         let challenge = &challenge_opt[0];
         let realm = challenge.realm.as_ref().unwrap();
-        let service = challenge.service.as_ref().unwrap();
+        let service = challenge.service.as_ref();
+        let mut query = vec![("scope", &scope)];
+
+        if service.is_some() {
```
bacongobbler, line 272: In the future, a more idiomatic approach would be:

```
if let Some(s) = service {
    query.push("service", s);
}
```

Using `if let` removes the extra `unwrap` call and cleans up the code to be more readable.

https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

thomastaylor312: Haha @bacongobbler I caught that after I merged and already did it locally. PR coming soon

jiayihu: You're right @bacongobbler, I had a feeling it felt wrong but clippy wasn't saying anything. Sorry 😅

PR #475, opened 2021-01-08
ref(oci): Fixes nit I missed in inital review

thomastaylor312:
I missed this nit when I was reviewing #474. Also contains a `Cargo.lock` update that happened when I built 

cc @bacongobbler


bacongobbler:
Would you mind pulling out the Cargo.lock changes? Those are handled by both #468 and #473; the `cfg-if` changes will conflict with #468.

thomastaylor312:
Ok, it is taken out

bacongobbler (approved):

PR #478, opened 2021-01-10
Clippy 1.49.0

kflansburg:
Addresses new Clippy lints. 

Ignores one type of lint in two locations due to https://github.com/rust-lang/rust-clippy/issues/6559


thomastaylor312 (approved):
Thanks for catching all these

PR #479, opened 2021-01-12
Krator Example

kflansburg:


crates/krator/src/lib.rs
```diff
@@ -11,3 +11,4 @@ pub mod state;
 pub use object::{ObjectState, ObjectStatus};
 pub use operator::Operator;
 pub use runtime::OperatorRuntime;
+pub use state::{State, Transition, TransitionTo};
```
thomastaylor312, line 14: Should we put most of these (including the ones above) in a `prelude`?

thomastaylor312 (approved):

bacongobbler (approved):
The moose approves of this PR.

picture of a moose

crates/krator/src/lib.rs
kflansburg: I started doing this actually. It got lost in the rebase. I will add later as the API stabilizes.

PR #487, opened 2021-01-17
Krator Support for Modified Events

### LINKED ISSUE #481: Extend Krator framework to enable it to react to changes to the watched object
**Proposed Enhancement**
I've looked the Krator crate to see whether we could use it as basis for our Kubernetes operators.

One limitation that I noticed is that the framework currently does not provide a method to allow reacting to changes to a watched object. In the current implementation of the state machine when the watched object in the apiserver changes it is updated in the shared state, but this in only exposed to client code after the next state transition. And to notice the implementation would have to keep a copy of the object and compare on every state transition.

This works well with pods, which are not usually updated, but rather deleted and recreated. However for operators the watched objects are normally patched, not recreated, which means that the state machine would need to be notified of these changes. I think this will be a frequent requirement and is worth offering as an optional function.

**Possible Implementations**
There are a couple of potential solutions that come to mind, I'll just list them here with no claim to completeness  or preference whatsoever:

- add an informer to the podstate that can be used to listen for changes - this is a workaround that wouldn't require changes to the framework
- copy the way that deletions are handled and notify state objects, I've played around a little and created a [test implementation](https://github.com/soenkeliebau/krustlet/commits/apply_notify) (not against Krator, but back when the code was still part of the Kubelet)
- allow stating a "reconcile-state" that is transitioned to automatically (Note: we may need to think about allowing to delay this transition in case the operator is currently doing something that shouldn't be interrupted)

I'm sure there are plenty more possible solutions, I'll leave it at that for now and see if other people think this is a legitimate need.

**Prior discussion:**

We've discussed this on the [slack channel](https://kubernetes.slack.com/archives/C015X72FV54/p1605741144159700) a little while ago, which I'm linking here for reference - I couldn't find a way to export the conversation, if anybody know of something for that please let me know and I'll attach an export to this ticket.

kflansburg:
First attempt at handling modified events for Krator. I replaced most references to the object manifest with a `tokio::sync::watch::Receiver<Manifest>`. This should allow you to fetch the latest manifest, and optionally await updates to that manifest. The current issue I'm facing is that once `recv` is called, subsequent calls will await new updates, even for clones of the receiver. This means that most of the uses here will block on their first use until the Pod is updated. We may have to implement a custom alternative without this behavior. 

cc @soenkeliebau
closes #481

### DIFF
--- a/crates/krator/src/manifest.rs
+++ b/crates/krator/src/manifest.rs
@@ -1,8 +1,11 @@
+use core::pin::Pin;
+use core::task::{Context, Poll};
+use tokio::stream::Stream;
 use tokio::sync::watch::{channel, Receiver, Sender};
 
-#[derive(Clone)]
 /// Wrapper for `ObjectState::Manifest` type which reflects
 /// the latest version of the object's manifest.
+#[derive(Clone)]
 pub struct Manifest<T: Clone> {
     rx: Receiver<T>,
 }
@@ -20,13 +23,10 @@ impl<T: Clone> Manifest<T> {
     }
 }
 
-impl<T: Clone> tokio::stream::Stream for Manifest<T> {
+impl<T: Clone> Stream for Manifest<T> {
     type Item = T;
 
-    fn poll_next(
-        mut self: core::pin::Pin<&mut Self>,
-        cx: &mut core::task::Context,
-    ) -> core::task::Poll<Option<Self::Item>> {
-        core::pin::Pin::new(&mut self.rx).poll_next(cx)
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        Pin::new(&mut self.rx).poll_next(cx)
     }
 }
--- a/crates/krator/src/manifest.rs
+++ b/crates/krator/src/manifest.rs
@@ -1,0 +1,32 @@
+use tokio::sync::watch::{channel, Receiver, Sender};
+
+#[derive(Clone)]
+/// Wrapper for `ObjectState::Manifest` type which reflects
+/// the latest version of the object's manifest.
+pub struct Manifest<T: Clone> {
+    rx: Receiver<T>,
+}
+
+impl<T: Clone> Manifest<T> {
+    /// Create a new Manifest wrapper from the initial object manifest.
+    pub fn new(inner: T) -> (Sender<T>, Self) {
+        let (tx, rx) = channel(inner);
+        (tx, Manifest { rx })
+    }
+
+    /// Obtain a clone of the latest object manifest.
+    pub fn latest(&self) -> T {
+        self.rx.borrow().clone()
+    }
+}
+
+impl<T: Clone> tokio::stream::Stream for Manifest<T> {
+    type Item = T;
+
+    fn poll_next(
+        mut self: core::pin::Pin<&mut Self>,
+        cx: &mut core::task::Context,
+    ) -> core::task::Poll<Option<Self::Item>> {
+        core::pin::Pin::new(&mut self.rx).poll_next(cx)
+    }
+}
--- a/crates/krator/src/runtime.rs
+++ b/crates/krator/src/runtime.rs
@@ -5,8 +5,8 @@ use std::sync::Arc;
 use futures::{StreamExt, TryStreamExt};
 use log::{debug, error, info, warn};
 use tokio::sync::mpsc::Sender;
+use tokio::sync::watch::{channel, Receiver};
 use tokio::sync::Notify;
-use tokio::sync::RwLock;
 
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use k8s_openapi::Metadata;
@@ -123,16 +123,15 @@ impl<O: Operator> OperatorRuntime<O> {
 
         let deleted = Arc::new(Notify::new());
 
-        let (shared_manifest, object_state) = match initial_event {
+        let (manifest, object_state) = match initial_event {
             Event::Applied(manifest) => {
                 let object_state = self.operator.initialize_object_state(&manifest).await?;
-                let manifest = Arc::new(RwLock::new(manifest));
                 (manifest, object_state)
             }
             _ => return Err(anyhow::anyhow!("Got non-apply event when starting pod")),
         };
 
-        let reflector_manifest = Arc::clone(&shared_manifest);
+        let (manifest_tx, manifest_rx) = channel(manifest);
         let reflector_deleted = Arc::clone(&deleted);
 
         // Two tasks are spawned for each resource. The first updates shared state (manifest and
@@ -154,7 +153,13 @@ impl<O: Operator> OperatorRuntime<O> {
                         if meta.deletion_timestamp.is_some() {
                             reflector_deleted.notify();
                         }
-                        *reflector_manifest.write().await = manifest;
+                        match manifest_tx.broadcast(manifest) {
+                            Ok(()) => (),
+                            Err(e) => {
+                                warn!("Unable to broadcast manifest update: {:?}", e);
+                                return;
+                            }
+                        }
                     }
                     Event::Deleted(manifest) => {
                         // I'm not sure if this matters, we get notified of pod deletion with a
@@ -167,6 +172,13 @@ impl<O: Operator> OperatorRuntime<O> {
                             manifest.namespace()
                         );
                         reflector_deleted.notify();
+                        match manifest_tx.broadcast(manifest) {
+                            Ok(()) => (),
+                            Err(e) => {
+                                warn!("Unable to broadcast manifest update: {:?}", e);
+                                return;
+                            }
+                        }
                         break;
                     }
                     _ => warn!("Resource got unexpected event, ignoring: {:?}", &event),
@@ -176,7 +188,7 @@ impl<O: Operator> OperatorRuntime<O> {
 
         tokio::spawn(run_object_task::<O>(
             self.client.clone(),
-            shared_manifest,
+            manifest_rx,
             self.operator.shared_state().await,
             object_state,
             deleted,
@@ -249,33 +261,38 @@ impl<O: Operator> OperatorRuntime<O> {
 
 async fn run_object_task<O: Operator>(
     client: Client,
-    manifest: Arc<RwLock<O::Manifest>>,
+    mut manifest: Receiver<O::Manifest>,
     shared: SharedState<<O::ObjectState as ObjectState>::SharedState>,
     mut object_state: O::ObjectState,
     deleted: Arc<Notify>,
     operator: Arc<O>,
 ) {
     debug!("Running registration hook.");
-    match operator.registration_hook(Arc::clone(&manifest)).await {
-        Ok(()) => debug!("Running hook complete."),
-        Err(e) => {
-            error!("Operator registration hook failed: {:?}", e);
-            return;
-        }
-    }
-
     let state: O::InitialState = Default::default();
     let (namespace, name) = {
-        let m = manifest.read().await;
+        let m = match manifest.recv().await {
+            Some(manifest) => manifest,
+            None => {
+                warn!("Manifest sender dropped.");
+                return;
+            }
+        };
+        match operator.registration_hook(manifest.clone()).await {
+            Ok(()) => debug!("Running hook complete."),
+            Err(e) => {
+                error!("Operator registration hook failed: {:?}", e);
+                return;
+            }
+        }
         (m.namespace(), m.name())
     };
 
     tokio::select! {
-        _ = run_to_completion(&client, state, shared.clone(), &mut object_state, Arc::clone(&manifest)) => (),
+        _ = run_to_completion(&client, state, shared.clone(), &mut object_state, manifest.clone()) => (),
         _ = deleted.notified() => {
             let state: O::DeletedState = Default::default();
             debug!("Object {} in namespace {:?} terminated. Jumping to state {:?}.", name, &namespace, state);
-            run_to_completion(&client, state, shared.clone(), &mut object_state, Arc::clone(&manifest)).await;
+            run_to_completion(&client, state, shared.clone(), &mut object_state, manifest.clone()).await;
         }
     }
 
--- a/crates/krator/src/runtime.rs
+++ b/crates/krator/src/runtime.rs
@@ -5,7 +5,6 @@ use std::sync::Arc;
 use futures::{StreamExt, TryStreamExt};
 use log::{debug, error, info, warn};
 use tokio::sync::mpsc::Sender;
-use tokio::sync::watch::{channel, Receiver};
 use tokio::sync::Notify;
 
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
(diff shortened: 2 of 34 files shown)


kflansburg:
Looking more closely, the Tokio `1.0` has clearer semantics, we dont have to wait for Tokio `1.0` though. Basically, `recv` should be called to `await` new changes, `borrow` is what we should call to access a reference to the value right away. We just need to be sure not to hold that lock. This may be something we implement a wrapper type for given the complexity there. I will take another crack tomorrow.

kflansburg:
Ok, I've moved to using a wrapper type, which cleaned up a lot of stuff. 

If you just want the latest version of the object, you can call `latest()` which clones the object and does not require error checking, which is nice. 

If you want to monitor for changes, the wrapper implements `Stream` which just passes to the underlying `watch::Receiver`. I think here is where we may want to improve the ergonomics, because I think it will return right away with the latest revision in some cases, and not in others, so we may need to add some logic to make this more predictable.

We can probably also eventually extend this wrapper to easily access Status and Metadata information.

crates/kubelet/src/container/state.rs
```diff
@@ -20,26 +23,41 @@ pub async fn run_to_completion<S: ResourceState<Manifest = Container, Status = S
...
 
     let mut state: Box<dyn State<S>> = Box::new(initial_state);
 
+    // Forward pod updates as container updates.
+    let initial_container = initial_pod.find_container(&container_name).unwrap();
+    let (container_tx, container_rx) = Manifest::new(initial_container);
+    let mut task_pod = pod.clone();
+    let task_container_name = container_name.clone();
+    tokio::spawn(async move {
+        while let Some(latest_pod) = task_pod.next().await {
```
kflansburg, line 42: Here is the only example right now of using the `Stream` API. I use it to detect pod updates and forward changes to individual containers to the `Manifest<Container>` used in the container state machine.

crates/krator/src/manifest.rs
```diff
@@ -0,0 +1,32 @@
+use tokio::sync::watch::{channel, Receiver, Sender};
+
+#[derive(Clone)]
```
thomastaylor312, line 3: SUPER NIT: Normally we have the `derive` under the comment

crates/krator/src/manifest.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl<T: Clone> Manifest<T> {
+    /// Create a new Manifest wrapper from the initial object manifest.
+    pub fn new(inner: T) -> (Sender<T>, Self) {
+        let (tx, rx) = channel(inner);
+        (tx, Manifest { rx })
+    }
+
+    /// Obtain a clone of the latest object manifest.
+    pub fn latest(&self) -> T {
```
thomastaylor312, line 18: Nit: Should this just return a borrow and allow the user to clone? I know it might not be worth it

crates/krator/src/manifest.rs
```diff
@@ -0,0 +1,32 @@
...
+        self.rx.borrow().clone()
+    }
+}
+
+impl<T: Clone> tokio::stream::Stream for Manifest<T> {
+    type Item = T;
+
+    fn poll_next(
+        mut self: core::pin::Pin<&mut Self>,
```
thomastaylor312, line 27: I think the types in this function should be imported from `std`. Also, isn't `Pin` in scope by default?

crates/krator/src/manifest.rs
```diff
@@ -0,0 +1,32 @@
...
+        (tx, Manifest { rx })
+    }
+
+    /// Obtain a clone of the latest object manifest.
+    pub fn latest(&self) -> T {
+        self.rx.borrow().clone()
+    }
+}
+
+impl<T: Clone> tokio::stream::Stream for Manifest<T> {
```
thomastaylor312, line 23: Good call on implementing `Stream` here

crates/krator/src/runtime.rs
```diff
@@ -167,6 +172,13 @@ impl<O: Operator> OperatorRuntime<O> {
                             manifest.namespace()
                         );
                         reflector_deleted.notify();
+                        match manifest_tx.broadcast(manifest) {
+                            Ok(()) => (),
+                            Err(e) => {
+                                warn!("Unable to broadcast manifest update: {:?}", e);
```
thomastaylor312, line 178: Nit: I think that in a deleted case, this might be better classified as an error as it is kinda important that it notifies it is going to be gone. But I could be wrong here

crates/kubelet/src/container/state.rs
```diff
@@ -20,26 +23,41 @@ pub async fn run_to_completion<S: ResourceState<Manifest = Container, Status = S
...
-    };
+    let initial_pod = pod.latest();
+    let namespace = initial_pod.namespace().to_string();
+    let pod_name = initial_pod.name().to_string();
+    let api: Api<KubePod> = Api::namespaced(client.clone(), &namespace);
 
     let mut state: Box<dyn State<S>> = Box::new(initial_state);
 
+    // Forward pod updates as container updates.
+    let initial_container = initial_pod.find_container(&container_name).unwrap();
```
thomastaylor312, line 37: This `unwrap` should be removed and the error handled here and on line 43

crates/wascc-provider/src/states/pod/starting.rs
```diff
@@ -42,7 +42,7 @@ impl State<PodState> for Starting {
                 Arc::clone(&pod_state.run_context),
             );
             let task_provider = Arc::clone(&provider_state);
-            let task_pod = Arc::clone(&arc_pod);
+            let task_pod = pod_rx.clone();
```
thomastaylor312, line 45: Nit: This could be a lot of allocations/clones in the case of a lot of containers. Perhaps an arc is still useful here?

crates/wasi-provider/src/states/pod/initializing.rs
```diff
@@ -58,7 +60,7 @@ impl State<PodState> for Initializing {
                 // TODO: I think everything should be a SharedState to the same pod in the reflector.
                 Arc::clone(&provider_state),
                 container_state,
-                Arc::new(RwLock::new(pod.clone())),
+                pod_rx.clone(),
```
thomastaylor312, line 63: Nit: Same thing here with cloning

thomastaylor312 (commented):
Just a few comments to discuss, this is mostly ready to go. Though I would hope @soenkeliebau can take a look at this as well to give us another point of view

crates/krator/src/manifest.rs
soenkeliebau: I think this probably returns a .clone() because borrows [hold a read lock](https://docs.rs/tokio/1.0.2/tokio/sync/watch/struct.Receiver.html#method.borrow) and could cause the sender to block, so returning a borrow here would force the user to know this and keep the borrow short lived.

kflansburg: This is correct, it was one of the reasons for wrapping here to protect the user from this easy misuse of the reference.

soenkeliebau:
I've had this on my todo list for today, but somehow got sidetracked. I've started looking at it, but may have had one glass of wine too many at this point - I'll  finish up tomorrow and post my review.

First impression though, looks great!

crates/krator/src/runtime.rs
kflansburg: I think this error would only happen if all `Receivers` are dropped. That would mean that `run_to_completion` has completed or panicked, so I don't think we can really do anything. I can bump the log level to `ERROR` if you wish.

crates/kubelet/src/container/state.rs
kflansburg: Should this error be reported to Kubernetes? If the Container is missing from the Pod manifest, I would expect `patch_container_status` to behave poorly.

kflansburg: For line 43, I will log an error, but I dont think it makes sense to stop trying (in case a subsequent Pod does have the container in it).

crates/wascc-provider/src/states/pod/starting.rs
kflansburg: Hmmm, `Manifest` requires a `&mut` to use it as a `Stream`. That prevents us from using just an `Arc`, and any other sort of lock would prevent us from streaming concurrently. I think these (`tokio::watch::Receiver`) are fairly lightweight?

crates/kubelet/src/container/state.rs
```diff
@@ -20,26 +23,60 @@ pub async fn run_to_completion<S: ResourceState<Manifest = Container, Status = S
...
+        while let Some(latest_pod) = task_pod.next().await {
+            let latest_container = match latest_pod.find_container(&task_container_name) {
+                Some(container) => container,
+                None => {
+                    error!(
+                        "Unable to locate container {} in pod {} manifest.",
+                        &task_container_name,
+                        latest_pod.name()
+                    );
+                    continue;
```
soenkeliebau, line 59: Should this trigger some sort of error handling? I don't think it is legal in Kubernetes to patch a pod and add a container, which might have triggered this, what other scenarios are there that might cause this to fail?

soenkeliebau (commented):
I've looked through the code and left a few comments - more questions really to be honest.

crates/krator/src/runtime.rs
```diff
@@ -154,7 +153,13 @@ impl<O: Operator> OperatorRuntime<O> {
                         if meta.deletion_timestamp.is_some() {
                             reflector_deleted.notify();
                         }
-                        *reflector_manifest.write().await = manifest;
+                        match manifest_tx.broadcast(manifest) {
+                            Ok(()) => (),
+                            Err(e) => {
+                                warn!("Unable to broadcast manifest update: {:?}", e);
+                                return;
```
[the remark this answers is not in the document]
kflansburg: I don't think it makes sense to keep trying with this task if this fails. We also haven't settled on a policy of automatically restarting the state machine in the event of an unhandled error. I'm not convinced that this would be desirable behavior. 

I could see an argument for calling `reflector_deleted.notify()` here, to signal to the other task that it needs to exit.

crates/krator/src/runtime.rs
kflansburg:
```
 it has to be possible for the watched objects to persist when the Krustlet is shut down, and in case it is deleted in the meantime, we need that notification.
```
I'm not sure I follow what you are saying here. If Krator shuts down (SIGINT), it actually doesnt "evict" any objects. This is functionality implemented in Krustlet. Right now Krator just becomes unschedulable essentially. It actually wont shut down. I'm not sure what the desired behavior here would be. As far as getting a "notification" that an object has been deleted (I'm assuming when Krator starts up again?), the deletion timestamp should be set so I believe this will work fine. 

Regarding finalizers:

1. If an object which Krator has a finalizer registered for is deleted while Krator is shut down, I think the object would get stuck, but this is expected behavior with finalizers (at least I experience it pretty often). 
2. If Krator is operating an object with Krator finalizers, I think this should all work, just do the finalizers in the `Deleted` state and then exit. The only thing we might need here is a way to remove the finalizer once it has run. 
3. If Krator is operating an object with external finalizers, there will probably be an error logged when it tries to delete the object if those have not completed. I'm not sure what is responsible for retrying that delete or if k8s will delete the object once the finalizers have run.

crates/kubelet/src/container/state.rs
kflansburg: I don't think it is legal, but I don't think it makes sense to do anything beyond logging an error and then continuing to pass on subsequent updates if they are valid. I suppose we could abort the whole Pod but that seems drastic.

crates/krator/src/runtime.rs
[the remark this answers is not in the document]
soenkeliebau: I might be misunderstanding fundamentally what the operator framework is trying to achieve :)

I'll briefly describe our use case and what we would be trying to achieve with it - maybe that helps illustrate where I am coming from..

We are looking to write for example a ZookeeperOperator, this operator would watch CRDs of the type zookeepercluster and for every object of that type create some config maps and pods, which are picked up and acted upon by our Krustlet-Provider implementation. So we would have one operator and many Krustlets (PodOperator - not sure what the correct term would be).

When the operator receives a notification that the zookeepercluster object has been deleted it will delete all _dependent_ pods, causing the Krustlets to stop all services on the worker servers. So for us a delete event on the object watched by the operator is a fairly significant thing and should not be triggered by a _malfunction_ in the operator.

Not sure if that makes any sense, I am probably too close to what we are doing to explain it well. Happy to elaboate further or jump on a quick call if you think that helps!

crates/krator/src/runtime.rs
soenkeliebau: My understanding of finalizers is that an object will not be fully deleted from Kubernetes as long as at least one finalizer is attached to it. Basically if Krator adds a finalizer to an object it is saying "do not under any circumstances delete this before I remove my finalizer, I will be back to do so!"
Without a finalizer it is my understanding that when deleting an object a deletion timestamp will be set, but the object could vanish before Krator comes back (crashed over the weekend, restarted monday, gone for a few days) - this scenario could be solved by using a finalizer.

You are right in that I can simply set the finalizer in my code and remove it in the deleted state, this is not necessarily something which needs to be part of Krator - I honestly hadn't even thought of that :)

crates/kubelet/src/container/state.rs
thomastaylor312: Logging is completely fine for now (as we don't know if the error should be reported to k8s). Definitely don't want a panic though, which the log takes care of

crates/wascc-provider/src/states/pod/starting.rs
thomastaylor312: My worry isn't about the channel so much as the embedded manifest (the inner `T`) which is a pretty large struct

thomastaylor312: Oh, I just realized that it is just the embedded receiver. This is fine then

crates/krator/src/runtime.rs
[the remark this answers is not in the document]
thomastaylor312: Not sure if I follow here or not, I don't think we are triggering a delete in the case of a return. It just stops listening for events on that state machine. I think we could maybe improve the behavior that happens in that case, but I don't think it triggers a delete based on my reread of the code

crates/kubelet/src/container/state.rs
thomastaylor312: Yeah, this was an explicit decision if I remember correctly. Aborting the whole pod felt like overkill for something that could fail on a transient network failure

crates/krator/src/runtime.rs
kflansburg: I see, it does sound like a finalizer would be a good way to ensure that Krator is able to do that cleanup. 

FWIW, my suggestion of setting / removing the finalizer is how I think they are intended to be used (each controller detects the deletion timestamp, runs its finalizer, and then patches the Pod to remove the finalizer.

crates/krator/src/runtime.rs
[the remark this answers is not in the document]
kflansburg: Yeah, I'm not really sure what to do here. I can think of situations where you would not want to restart the state machine automatically (a naive implementation would recreate all the ZK pods right?). 

One thing I can suggest if the operator is only responsible for create / delete, is to mark all of the Pods, etc. as child objects of the ZookeeperCluster. K8s will then handle the garbage collection for you. I've done this with Spark.

thomastaylor312 (approved):

PR #488, opened 2021-01-17
chore: don't use relative path in tsconfig

willemneal:


bacongobbler:
hi @willemneal. Would you mind describing what issue this fixes? I can't tell what I'm supposed to be testing for here. Thanks!

willemneal:
You should use node resolution for finding the types info.  This way if you have a project with a workspace (see [asbuild](https://github.com/AssemblyScript/asbuild)), then nested directories won't need to update their paths' to be relative to the top level directory.  Also this is example code so it's good to give a proper example.

bacongobbler (approved):

PR #490, opened 2021-01-18
Updated|Corrected Microk8s instructions

DazWilkin:
Reran this document after a hiatus.

+ II[U|R]C setting `KUBECONFIG=foo ./krustlet-... ---bootstrap-file=bar` uses `bar` to authenticate to the cluster so `foo` is redundant
+ Corrected a typo in the `Step 3: test`
+ Updated the `hello-world-wasi-rust` output to reflect current state

Now to review the GKE instructions ;-)

PR #494, opened 2021-01-25
Lint docs for all PRs and branches

### LINKED ISSUE #493: Update tests to run markdown lint for docs path changes.

kflansburg:
It doesn't appear to be possible to skip some jobs based on what paths were updated, so I split lints out into a new Action that runs for everything. 

cc #490 
Closes #493

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,11 +3,7 @@ on:
   push:
     branches:
       - master
-    paths-ignore:
-      - "docs/**"
-  pull_request:
-    paths-ignore:
-      - "docs/**"
+  pull_request: {}
 jobs:
   build:
     runs-on: ${{ matrix.config.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,22 +9,6 @@ on:
     paths-ignore:
       - "docs/**"
 jobs:
-  lint_docs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


bacongobbler:
Right now we have a `paths-ignore` rule in build.yaml so we don't run tests on doc changes. Should we opt to remove that and instead add a `paths` rule in lint.yaml for any doc changes?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

bacongobbler:
image

kflansburg:
Ah, I did not see the paths directive, that's exactly what I was looking for! I will give that a shot.

kflansburg:
Ok @bacongobbler I think this is ready.

thomastaylor312:
This is looking good to me, I just have one question. So as it stands this looks like what happens:

1. Any change to code (including docs changes) will trigger a full test run
2. Any change to stuff within `docs/` will trigger 1 additional lint pass

So are we going to require PRs to have passing doc lint tests? If that is the case, then when a PR does not include docs changes, it will end up with one "hanging" test every time. How do we want to handle that?

kflansburg:
Can we merge this soon? #514 introduced more failing documentation.

kflansburg:
> So are we going to require PRs to have passing doc lint tests? If that is the case, then when a PR does not include docs changes, it will end up with one "hanging" test every time. How do we want to handle that?

My hope is that these tests do not show in the UI as hanging. If it does, then I say we just run everything for all PRs.

thomastaylor312 (approved):

PR #499, opened 2021-01-28
fix(derive): Fixes kubelet-derive and renames it

thomastaylor312:
This moves the kubelet-derive crate to now be the krator-derive crate. As part
of this, it can now handle more complex types properly and has been reexported
within the `kubelet` crate. I updated all of the current providers to use the
derive option instead of a manual implementation as well.

Note that I also found that we weren't turning on a tokio feature we should have
used. I honestly don't know why the grpc stuff was functioning without it, but
now it is on just for safety sake


kflansburg (approved):
LGTM

PR #506, opened 2021-01-29
fix(krator): Removes `derive` feature from default

thomastaylor312:
I put it there for testing and forgot to remove it. I also made sure it
was listed as a required feature for the moose example so people get a
nice error message if they don't pass the feature flag


kflansburg (approved):
LGTM

PR #507, opened 2021-01-31
Update wasmtime to 0.24

### LINKED ISSUE #466: Support aarch64-unknown-linux-musl
Not a request, I'm opening an issue to help anyone having the same issue in future. I've able to run krustlet with k3s on my raspberry pi 3B+ cluster without issues when I was using Ubuntu Server.

However, I decided to migrate to Alpine Linux to use less resources, but Alpine uses `musl` instead of `glibc`. This resulted in the arm64 binary not being runnable because it's compiled to be dynamically linked with `/lib/ld-linux-aarch64.so.1`. I had no luck with the compat packages at https://pkgs.alpinelinux.org/contents?file=ld-linux-*&path=&name=&branch=&repo=main&arch=

So I decided to build krustlet myself. To do so, you must wait for https://github.com/bytecodealliance/wasmtime/issues/2133 to be solved or patch `libc` yourself, like I did with the help of @lu-zero. Of course I assume one wants `krustlet-wasi`, `krustlet-wascc` had zero issues.

Then you cross-compile krustlet with `aarch64-unknown-linux-musl` target. On MacOS I used [cross](https://github.com/rust-embedded/cross) along with a custom image because they removed openssl support a while ago https://github.com/rust-embedded/cross/issues/229. You can find the Dockerfile I've used at https://github.com/jiayihu/fedra-thesis/blob/master/packages/fedra-compute/docker/Dockerfile. Or you can just use the image I've pushed configuring `Cross.toml`:

```toml
[target.aarch64-unknown-linux-musl]
image = "jiayihu/krustlet-musl:latest"
```

I was also successful by compiling to current target `aarch64-unknown-linux-gnu` with static linking. I've tried only the debug build and with rust-tls, but there shouldn't be surprises for release and openssl. Maybe that's the easiest solution to be applied in this repo. The only change needed is the compiler flag `-C target-feature=+crt-static`:

```
[target.aarch64-unknown-linux-gnu]
linker = "aarch64-linux-gnu-gcc"
rustflags = [ "-C", "target-feature=+crt-static"]
```

I can open a PR for it, what do you think?

jiayihu:
This PR updates to latest wasmtime version for `wasi-provider`. It also closes #466 
To be precise, one still needs to remove wascc from `Cargo.toml` otherwise it's compiled even with `--bin krustlet-wasi` and `wascc-host` still depends on `wasmtime 0.19`.

Updating `wascc-host` seems too much for me because I'm not very familiar with it. I used it only once several months ago. It also has changed name to https://github.com/wasmCloud/wasmCloud, so I don't know what else could have changed.


crates/wasi-provider/src/wasi_runtime.rs
```diff
@@ -168,12 +168,12 @@ impl WasiRuntime {
                 .envs(&data.env)
                 .stdout(wasi_common::OsFile::try_from(output_write.try_clone()?)?)
                 .stderr(wasi_common::OsFile::try_from(output_write.try_clone()?)?);
-            let mut ctx_builder_unstable = wasi_common::old::snapshot_0::WasiCtxBuilder::new();
+            let mut ctx_builder_unstable = wasi_common::WasiCtxBuilder::new();
```
jiayihu, line 171: I'm not very familiar with WASI in wasmtime. I guess `old__snapshot_0` is the first experimental implementation of WASI. The change to use `wasi_common::WasiCtxBuilder` is taken from the PR removing `old::snapshot_0`: https://github.com/bytecodealliance/wasmtime/pull/2444/files#diff-2218893006e42ddf80cedd8282a70b44af7fbb2b1423527dfcb1666a895ed211L366

bacongobbler:
It appears that CI is failing to compile the project due to the update.

> the package `wasi-common` links to the native library `wasi-common-19`, but it conflicts with a previous package which links to `wasi-common-19` as well:
package `wasi-common v0.22.0`
    ... which is depended on by `wasi-provider v0.5.0 (/home/runner/work/krustlet/krustlet/crates/wasi-provider)`
    ... which is depended on by `krustlet v0.5.0 (/home/runner/work/krustlet/krustlet)`

Would you mind looking into this?

jiayihu:
Sure, it's probably because I haven't committed the Cargo.lock changes yet. I'll have a look asap

thomastaylor312:
We might want to hold off on this until we move the wascc-provider over after the 0.6 release. It will make the dependencies a bit easier to manage

jiayihu:
After having a look at it, I believe it's not possible to fix the conflict without updating `wascc-provider`: https://users.rust-lang.org/t/cargo-fails-to-select-version-for-crate/20971/3

No problem in holding until 0.6 is release though

thomastaylor312:
@jiayihu We cut 0.6 yesterday! I'll be moving the wascc-provider tomorrow to its new home

crates/wasi-provider/src/wasi_runtime.rs
radu-matei: Yeah, `old_snapshot_0` was explicitly referencing the [snapshot phase](https://github.com/WebAssembly/WASI/tree/main/phases) of WASI.

jiayihu:
I opened a PR to update `wasmtime` in `wasi-provider` https://github.com/wapc/wasmtime-provider/pull/6

thomastaylor312:
I just finished moving the wascc provider over today: https://github.com/wasmCloud/krustlet-wasmcloud-provider. I will be pulling it out of the main krustlet repo soon

thomastaylor312:
@jiayihu Forgot to follow up here, you should be good to go for updating wasmtime now without any issues

radu-matei (approved):
I started updating WAGI to 0.24 on Friday, and all the changes here LGTM.

Thanks a lot, @jiayihu and @kflansburg.

kflansburg:
Thanks @radu-matei , like @jiayihu , I'm not very familiar with `wasmtime`, so I wasn't sure if this was super hacky. But I need at least `wasmtime` 0.23 for this to compile on Apple Silicon, so I figured I would try to push this over the line. @thomastaylor312 , do you think you could  review it as well before I merge?

jiayihu:
Sorry guys, I have been busy the later days and forgot to reply. Thanks @kflansburg for updating to even 0.25! In my case, I had a look at the commit on the wasmtime PR itself to understand how to migrate these experimental features.

thomastaylor312 (approved):
This looks good to me. Thanks @jiayihu and @kflansburg.

Also, a note for us to work on in a future PR: The latest version of wasmtime has `async` support (not within wasm itself, but the actual functions to spin it up). This means we can probably move away from spawn blocking

PR #511, opened 2021-02-02
fix(derive): Fixes circular dependency

thomastaylor312:
The dev dependency on krator led to a circular dependency issue that prevented
publishing of the crate. These changes were actually published with krator 0.1
so we could unblock things.


kflansburg (approved):
LGTM

PR #516, opened 2021-02-04
chore(*): Removes wascc from the repo

thomastaylor312:
Now that we have moved the wascc provider to its new home, we can remove
the dependency from here. We need to make sure we include an announcement
about this in the 0.7 release notes


docs/community/developers.md
```diff
@@ -41,8 +41,8 @@ implementation):
 $ just build --no-default-features --features rustls-tls
 ```
 
-The same flags can be passed to `just run-wasi` and `just run-wascc` if you want
-to just [run](#running) the project instead.
+The same flags can be passed to `just run`. if you want to just [run](#running)
```
bacongobbler, line 44: looks like a period was added here by mistake?

docs/community/developers.md
```diff
@@ -86,12 +86,7 @@ things to be aware of. See the [caveats](#caveats) section for more details
 
 ## Running
 
-There are two different runtimes available for Krustlet: `wascc` or `wasi`.
-
-The `wascc` runtime is a secure WebAssembly host runtime, connecting "actors"
-and "capability providers" together to connect your WebAssembly runtime to
-cloud-native services like message brokers, databases, or other external
-services normally unavailable to the WebAssembly runtime.
+There default included runtime with Krustlet is `wasi`.
```
bacongobbler, line 89: The*

bacongobbler (approved):
code LGTM

docs/community/developers.md
thomastaylor312: Yes it was

docs/community/developers.md
thomastaylor312: Good catch

PR #518, opened 2021-02-07
Use LC_ALL to set all locale categories

vasu1124:
On macOS 11.2, `LC_CTYPE=C` was not sufficient:
```
tr: Illegal byte sequence
tr: Illegal byte sequence
error: error validating "STDIN": error validating data: [unknown object type "nil" in Secret.stringData.token-id, unknown object type "nil" in Secret.stringData.token-secret]; if you choose to ignore these errors, turn validation off with --validate=false
```
I needed to set `LC_ALL=C`.

### DIFF
--- a/docs/howto/assets/bootstrap.sh
+++ b/docs/howto/assets/bootstrap.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Nice catch on this! Thank you for the fix

PR #519, opened 2021-02-14
Introduce deregistration hook in Krator Operator

### LINKED ISSUE #504: Create deregistration hook in Krator
As we have been implementing the CSI feature, we realized there isn't a way for a state machine to call some sort of deregistration logic that can do things like cleaning up volumes. Having a deregistration hook will allow state machines to hook common cleanup steps into the process after a state machine finishes. This hook should always run, even on state machine errors

kflansburg:
* Update Krustlet to use this hook to clean up volume mounts, rather
  than `Terminated` state.

Closes #504

### DIFF
--- a/crates/kubelet/src/provider/mod.rs
+++ b/crates/kubelet/src/provider/mod.rs
@@ -127,6 +127,11 @@ pub trait Provider: Sized + Send + Sync + 'static {
         Err(NotImplementedError.into())
     }
 
+    /// Gets the path at which to construct temporary directories for volumes.
+    fn volume_path(&self) -> Option<std::path::PathBuf> {
+        None
+    }
+
     /// Fetch the CSI driver plugin registry.
     fn plugin_registry(&self) -> Option<Arc<PluginRegistry>> {
         None
--- a/crates/krator/src/operator.rs
+++ b/crates/krator/src/operator.rs
@@ -58,4 +58,12 @@ pub trait Operator: 'static + Sync + Send {
         &self,
         manifest: Self::Manifest,
     ) -> crate::admission::AdmissionResult<Self::Manifest>;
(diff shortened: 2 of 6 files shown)


crates/kubelet/src/provider/mod.rs
```diff
@@ -127,6 +127,11 @@ pub trait Provider: Sized + Send + Sync + 'static {
         Err(NotImplementedError.into())
     }
 
+    /// Gets the path at which to construct temporary directories for volumes.
```
kflansburg, line 130: @bacongobbler I don't really like this pattern (also used before for `plugin_registry`). It would be too easy to forget to implement these methods, and probably super hard to debug given that there are only two places that make use of the method on the provider, and many that use the method on the state. IMO it is caused by volume functionality being _partially_ tied to `GenericProviderState`. I think it might make more sense for these methods to exist in only one place (Provider state), and for all providers, not just ones implementing `GenericProvider`. This could be enforced with a trait bound and provide the same default implementation that we see here. Thoughts?

kflansburg (commented):
We will need to update the Wascc provider as well.

thomastaylor312 (commented):
This is looking good. Gonna leave approval for @bacongobbler

crates/kubelet/src/provider/mod.rs
bacongobbler: I agree. Satisfying a Trait instead of overriding a function seems like the more idiomatic approach. Assuming there's a way to determine if a Provider satisfies a given trait, that would also give us a clean abstraction to determine if certain features are supported or not.

bacongobbler (approved):
I like the approach taken here way better. LGTM!

PR #520, opened 2021-02-15
Krator support for Admission Webhook

kflansburg:
This allows the Operator to validate, mutate, and/or reject incoming objects. This seems to be an uncommon pattern, but it makes a lot of sense to me that an Operator gets to validate the objects that it acts on. This is completely optional behind a feature flag (disabled for Krustlet). 

I've updated the moose example for (optionally, based on feature) testing this functionality, and to demonstrate provisioning certificates for mTLS and configuring Kubernetes to use the endpoint.

### DIFF
--- a/crates/krator/examples/assets/deploy.sh
+++ b/crates/krator/examples/assets/deploy.sh
@@ -30,6 +30,4 @@ cat webhook.yaml | envsubst | kubectl apply -f -
 rm server.req
 rm ca.srl
 rm ca.key
-#rm server.crt
-#rm server.key
 rm ca.crt
--- a/crates/krator/examples/assets/deploy.sh
+++ b/crates/krator/examples/assets/deploy.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Create CA Key
+openssl genrsa -out ca.key 2048
+
+# Create CA Cert
+openssl req -new -key ca.key -x509 -out ca.crt -days 3650 -subj "/CN=ca"
+
+# Create Server Key and Signing Request
+openssl req -new -nodes -newkey rsa:2048 -keyout server.key -out server.req -batch -subj "/CN=moose-admission-webhook.default.svc" 
+
+# Create Signed Server Cert
+openssl x509 -req -in server.req -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 3650 -sha256
+
+case `uname` in
+        Darwin)
+            b64_opts='-b=0'
+            export ENDPOINT=$(ipconfig getifaddr en0)
+            ;;
+        *)
+            b64_opts='--wrap=0'
+            export ENDPOINT="172.17.0.1"
+esac
+
+export CA_BUNDLE=$(cat ca.crt | base64 ${b64_opts})
+
+# Configure Admission Webhook
+cat webhook.yaml | envsubst | kubectl apply -f -
+
+rm server.req
+rm ca.srl
+rm ca.key
+#rm server.crt
+#rm server.key
+rm ca.crt
--- a/crates/krator/src/admission.rs
+++ b/crates/krator/src/admission.rs
@@ -97,13 +97,11 @@ pub(crate) async fn endpoint<O: Operator>(operator: Arc<O>) {
                         patch_type: None,
                     },
                 };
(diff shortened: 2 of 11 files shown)


crates/krator/examples/assets/deploy.sh
```diff
@@ -0,0 +1,35 @@
...
+
+export CA_BUNDLE=$(cat ca.crt | base64 ${b64_opts})
+
+# Configure Admission Webhook
+cat webhook.yaml | envsubst | kubectl apply -f -
+
+rm server.req
+rm ca.srl
+rm ca.key
+#rm server.crt
```
thomastaylor312, line 33: Nit: I assume you can remove this?

crates/krator/src/admission.rs
```diff
@@ -0,0 +1,115 @@
...
+    status: Option<Status>,
+    /// The patch body. Currently we only support "JSONPatch" which implements RFC 6902.
+    patch: Option<json_patch::Patch>,
+    /// The type of Patch. Currently we only allow "JSONPatch".
+    patch_type: Option<String>,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct AdmissionReviewRequest<T> {
```
thomastaylor312, line 50: Interesting, does `T` not need to be `Deserialize`?

crates/krator/src/admission.rs
```diff
@@ -0,0 +1,115 @@
...
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Status;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+/// Result of admission hook.
+#[allow(clippy::large_enum_variant)]
+pub enum AdmissionResult<T> {
+    /// Permit the request. Pass the object (with possible mutations) back.
+    /// JSON Patch of any changes will automatically be created.
+    Allow(T),
```
thomastaylor312, line 12: Does `T` need to be `Serializable` as it is going into a JSON patch?

crates/krator/src/admission.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// Deny the request. Pass a Status object to provide information about the error.
+    Deny(Status),
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// AdmissionRequest describes the admission.Attributes for the admission request.
+struct AdmissionRequest<T> {
```
thomastaylor312, line 20: Should all of these types be available in the `k8s_openapi` crate or are they something technically outside of the normal API spec?

crates/krator/src/admission.rs
```diff
@@ -0,0 +1,115 @@
...
+                    }
+                    AdmissionResult::Deny(status) => AdmissionResponse {
+                        uid: request.request.uid,
+                        allowed: false,
+                        status: Some(status),
+                        patch: None,
+                        patch_type: None,
+                    },
+                };
+                Ok::<warp::reply::Json, std::convert::Infallible>(warp::reply::json(
```
thomastaylor312, line 100: Nit: I think you can shorten this to `Ok::<_, std::convert::Infallible>`

thomastaylor312 (approved):
LGTM. Just a few questions I had that should have no impact on actually merging

crates/krator/src/admission.rs
kflansburg: Yes, but creating the actual patch only happens in the `endpoint` method so I think it makes sense to just rely on the fact that `T` comes from `Operator::Manifest`. This type isn't really intended to be used outside of the context of our API.

crates/krator/src/admission.rs
kflansburg: I'm not totally sure, they are defined [here](https://github.com/kubernetes/api/blob/master/admission/v1/generated.proto) which is in the main API source tree, but I searched for a while and could not find anything in `k8s_openapi` that implemented this. As far as I can tell it may be because there isn't an OpenAPI spec published for this by the Kubernetes API server.

crates/krator/src/admission.rs
kflansburg: I think if you tried to use a `T` that isn't `Deserialize`, the `derive` would fail to compile. But this is a private type that is only used in `endpoint` so I don't think this is a concern.

PR #521, opened 2021-02-15
oci-distribution: expose client builder

flavio:
# What this PR does

Allow users to provide a `reqwest::ClientBuilder` instance when the `oci-distribution::Client` is created.

# Why this is useful

This can be useful for users who want to fine tune the client used to access the remote registry. For example, to add a root certificate to be used when connecting to a registry secured with a self-signed certificate.

# Open questions

This is a quick prototype, what I don't like about this solution is:

  * Break the API of `oci-distrubution::Client`. The PR fixes the broken code, but what about other users of this crate?
  * Expose the `reqwest::ClientBuilder`. Disclaimer, I'm new to Rust, in Go I would propose to use instead an interface as parameter. In Rust I would say we could request a series of traits to be provided by the `ClientBuilder`, but this would require changes inside of `reqwest::ClientBuilder`.

I would be interested to know what you think about these drawbacks.

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -120,8 +120,17 @@ impl TryFrom<ClientConfig> for Client {
 
     fn try_from(config: ClientConfig) -> Result<Self, Self::Error> {
         let mut client_builder = reqwest::Client::builder()
-            .danger_accept_invalid_certs(config.accept_invalid_certificates)
-            .danger_accept_invalid_hostnames(config.accept_invalid_hostnames);
+            .danger_accept_invalid_certs(config.accept_invalid_certificates);
+
+        client_builder = match () {
+            #[cfg(feature = "native-tls")]
+            () => client_builder.danger_accept_invalid_hostnames(config.accept_invalid_hostnames),
+            #[cfg(not(feature = "native-tls"))]
+            () => {
+                warn!("Cannot change value of `accept_invalid_hostnames`: missing 'native-tls' feature");
+                client_builder
+            }
+        };
 
         for c in &config.extra_root_certificates {
             let cert = match c.encoding {
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -19,8 +19,9 @@ use hyperx::header::Header;
 use reqwest::header::HeaderMap;
 use sha2::Digest;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
-use tracing::debug;
+use tracing::{debug, warn};
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
 /// The data for an image or module.
@@ -114,14 +115,42 @@ pub trait ClientConfigSource {
     fn client_config(&self) -> ClientConfig;
 }
 
+impl TryFrom<ClientConfig> for Client {
+    type Error = anyhow::Error;
+
+    fn try_from(config: ClientConfig) -> Result<Self, Self::Error> {
+        let mut client_builder = reqwest::Client::builder()
+            .danger_accept_invalid_certs(config.accept_invalid_certificates)
+            .danger_accept_invalid_hostnames(config.accept_invalid_hostnames);
+
+        for c in &config.extra_root_certificates {
+            let cert = match c.encoding {
+                CertificateEncoding::Der => reqwest::Certificate::from_der(c.data.as_slice())?,
+                CertificateEncoding::Pem => reqwest::Certificate::from_pem(c.data.as_slice())?,
+            };
+            client_builder = client_builder.add_root_certificate(cert);
+        }
+
+        Ok(Self {
+            config,
+            tokens: HashMap::new(),
+            client: client_builder.build()?,
+        })
+    }
+}
+
 impl Client {
     /// Create a new client with the supplied config
     pub fn new(config: ClientConfig) -> Self {
-        Self {
-            config,
-            tokens: HashMap::new(),
-            client: reqwest::Client::new(),
-        }
+        Client::try_from(config.clone()).unwrap_or_else(|err| {
+            warn!("Cannot create OCI client from config: {:?}", err);
+            warn!("Creating client with default configuration");
+            Self {
+                config,
+                tokens: HashMap::new(),
+                client: reqwest::Client::new(),
+            }
+        })
     }
 
     /// Create a new client with the supplied config
@@ -710,11 +739,40 @@ impl Client {
     }
 }
 
+/// The encoding of the certificate
+#[derive(Debug, Clone)]
+pub enum CertificateEncoding {
+    #[allow(missing_docs)]
+    Der,
+    #[allow(missing_docs)]
+    Pem,
+}
+
+/// A x509 certificate
+#[derive(Debug, Clone)]
+pub struct Certificate {
+    /// Which encoding is used by the certificate
+    pub encoding: CertificateEncoding,
+
+    /// Actual certificate
+    pub data: Vec<u8>,
+}
+
 /// A client configuration
 #[derive(Debug, Clone, Default)]
 pub struct ClientConfig {
     /// Which protocol the client should use
     pub protocol: ClientProtocol,
+
+    /// Accept invalid hostname. Defaults to false
+    pub accept_invalid_hostnames: bool,
+
+    /// Accept invalid certificates. Defaults to false
+    pub accept_invalid_certificates: bool,
+
+    /// A list of extra root certificate to trust. This can be used to connect
+    /// to servers using self-signed certificates
+    pub extra_root_certificates: Vec<Certificate>,
 }
 
 /// The protocol that the client should use to connect
@@ -882,6 +940,7 @@ mod test {
     fn manifest_url_generation_respects_http_protocol() {
         let c = Client::new(ClientConfig {
             protocol: ClientProtocol::Http,
+            ..Default::default()
         });
         let reference = Reference::try_from("webassembly.azurecr.io/hello:v1".to_owned())
             .expect("Could not parse reference");
@@ -895,6 +954,7 @@ mod test {
     fn blob_url_generation_respects_http_protocol() {
         let c = Client::new(ClientConfig {
             protocol: ClientProtocol::Http,
+            ..Default::default()
         });
         let reference = Reference::try_from("webassembly.azurecr.io/hello@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff".to_owned())
             .expect("Could not parse reference");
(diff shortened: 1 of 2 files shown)


kflansburg:
I think this is a good idea. I've had to maintain a separate branch for testing against internal registries without cert validation. 

Regarding your questions: 
1. Could we create a new method, called `new_with_builder` or something similar. This would prevent the API from breaking, and the user would not have to pass a `None` everywhere when they don't need customization.
2. I'm not completely sure what your concern is here, but I think it is reasonable to re-export `reqwest::ClientBuilde`. This is a  pretty typical pattern in Rust and I think it is better than trying to maintain an interface that has feature parity with `reqwest::ClientBuilder`.

crates/oci-distribution/src/client.rs
```diff
@@ -880,9 +887,13 @@ mod test {
...
-        let c = Client::new(ClientConfig {
-            protocol: ClientProtocol::Http,
-        });
+        let c = Client::new(
+            ClientConfig {
+                protocol: ClientProtocol::Http,
+            },
+            None,
+        )
+        .expect("Could not crate Client");
```
bacongobbler, line 896: typo: create*

crates/oci-distribution/src/client.rs
```diff
@@ -893,9 +904,13 @@ mod test {
...
```
bacongobbler, line 913: typo

crates/oci-distribution/src/client.rs
```diff
@@ -912,7 +927,7 @@ mod test {
     fn manifest_url_generation_uses_https_if_not_on_exception_list() {
         let insecure_registries = vec!["localhost".to_owned(), "oci.registry.local".to_owned()];
         let protocol = ClientProtocol::HttpsExcept(insecure_registries);
-        let c = Client::new(ClientConfig { protocol });
+        let c = Client::new(ClientConfig { protocol }, None).expect("Could not crate Client");
```
bacongobbler, line 930: typo

crates/oci-distribution/src/client.rs
```diff
@@ -925,7 +940,7 @@ mod test {
     fn manifest_url_generation_uses_http_if_on_exception_list() {
         let insecure_registries = vec!["localhost".to_owned(), "oci.registry.local".to_owned()];
         let protocol = ClientProtocol::HttpsExcept(insecure_registries);
-        let c = Client::new(ClientConfig { protocol });
+        let c = Client::new(ClientConfig { protocol }, None).expect("Could not crate Client");
```
bacongobbler, line 943: typo

crates/oci-distribution/src/client.rs
```diff
@@ -938,7 +953,7 @@ mod test {
     fn blob_url_generation_uses_https_if_not_on_exception_list() {
         let insecure_registries = vec!["localhost".to_owned(), "oci.registry.local".to_owned()];
         let protocol = ClientProtocol::HttpsExcept(insecure_registries);
-        let c = Client::new(ClientConfig { protocol });
+        let c = Client::new(ClientConfig { protocol }, None).expect("Could not crate Client");
```
bacongobbler, line 956: typo

crates/oci-distribution/src/client.rs
```diff
@@ -955,7 +970,7 @@ mod test {
     fn blob_url_generation_uses_http_if_on_exception_list() {
         let insecure_registries = vec!["localhost".to_owned(), "oci.registry.local".to_owned()];
         let protocol = ClientProtocol::HttpsExcept(insecure_registries);
-        let c = Client::new(ClientConfig { protocol });
+        let c = Client::new(ClientConfig { protocol }, None).expect("Could not crate Client");
```
bacongobbler, line 973: typo

crates/oci-distribution/src/client.rs
```diff
@@ -1191,9 +1206,14 @@ mod test {
...
-        let mut c = Client::new(ClientConfig {
-            protocol: ClientProtocol::Http,
-        });
+        let mut c = Client::new(
+            ClientConfig {
+                protocol: ClientProtocol::Http,
+            },
+            None,
+        )
+        .expect("Could not crate Client");
```
bacongobbler, line 1215: typo

crates/oci-distribution/src/client.rs
```diff
@@ -1232,9 +1252,13 @@ mod test {
...
```
bacongobbler, line 1261: typo

crates/oci-distribution/src/client.rs
```diff
@@ -1293,9 +1317,13 @@ mod test {
...
-        let mut c = Client::new(ClientConfig {
-            protocol: ClientProtocol::HttpsExcept(vec!["oci.registry.local".to_string()]),
-        });
+        let mut c = Client::new(
+            ClientConfig {
+                protocol: ClientProtocol::HttpsExcept(vec!["oci.registry.local".to_string()]),
+            },
+            None,
+        )
+        .expect("Could not crate Client");
```
bacongobbler, line 1326: typo

bacongobbler (changes requested):
I agree with Kevin's points. Found a few typos in the PR while reviewing.

crates/oci-distribution/src/client.rs
flavio: Now resolved, that's what happens when you do copy and paste ;)

flavio:
>     1. Could we create a new method, called `new_with_builder` or something similar. This would prevent the API from breaking, and the user would not have to pass a `None` everywhere when they don't need customization.

I think this is actually a good idea. I've updated the PR to reflect that.

>     2. I'm not completely sure what your concern is here, but I think it is reasonable to re-export `reqwest::ClientBuilde`. This is a  pretty typical pattern in Rust and I think it is better than trying to maintain an interface that has feature parity with `reqwest::ClientBuilder`.

I just didn't want to expose an internal dependency of this crate to the outside. It's a matter of taste, some people dislike things like that :shrug: 

Thanks for the feedback

kflansburg (approved):
LGTM

bacongobbler (approved):
I would prefer that we somehow remove the hard reliance on `reqwest`. We managed to accomplish this via the `ClientConfig` object and the `ClientConfigSource` trait. It would allow consumers to use oci-distribution without relying on re-implementing `reqwest`'s internals. It also gives us some leeway on how we handle backwards compatibility. Especially since reqwest still hasn't hit 1.0 yet

That being said, the library has a fairly hard dependency on `reqwest` at this time, so I'm fine with supporting this.

flavio:
@bacongobbler I share your concerns, that was what I tried to bring up with my second question.

Thanks for the hint, I think I now know how I could do something clean and idiomatic with rust. I'll come back with an updated PR

thomastaylor312:
@flavio Just wanted to make sure I wasn't missing something here. Were you going to update your PR here or did you want us to merge this for now?

flavio:
@thomastaylor312 sorry, I've been pretty busy recently. I'll work on this PR tomorrow

thomastaylor312:
No problem! Just wanted to make sure we weren't leaving you hanging

flavio:
I turned the PR into a draft because I need your feedback on the latest set of changes I submitted.

I took inspiration from what @bacongobbler suggested and I looked into `ClientConfig` and `ClientConfigSource`. I added a bunch of generic data types to the `ClientConfig` struct to handle client security options. I thought that, in the end, all these knobs are related with the Client configuration.

I felt pretty happy about the idea, until I went to change `Client::new` and I realized I had a problem. The `Reqwest::ClientBuilder` is returning a `Result<Reqwest::Client>`, not a `Reqwest::Client` I **temporarily** solved the issue by calling `unwrap()` in a bunch of places :(

I see two possible solutions:
  * We break the `Client::new` API by changing the return value to be a `Result` :unamused: 
  * I move all these extra fields into a new struct (let's say `ClientSecurityConfig`) and then I create a `Client::new_with_security_opts(config: ClientConfig, securityConfig: ClientSecurityConfig) -> anyhow::Result<Self>`. I don't like to end up having two different "client configuration" structs, but this won't at least break the API

Do you have other ideas, how do you feel about these changes and options?

flavio:
ping @thomastaylor312 :)

thomastaylor312:
I'll let @kflansburg and @bacongobbler take a look at this again. Looks like you have some missing fields though that are causing the build failures

flavio:
Yeah, I didn't bother to fix the tests because I would like to know which path I should follow to implement the change. The code is also callling a couple of `unwrap()` for the same reason.

I change the PR to a `draft` one because of that :)

flavio:
Ping, can someone provide me the information to move forward with this PR?

kflansburg (approved):
@flavio Sorry this got lost, can you mark it as ready for review? 

I believe this should be merged. I think you've found a good balance between the two patterns that we discussed.

kflansburg:
@flavio It looks like there is one remaining `Default` that you need to add at `crates/kubelet/src/config_interpreter.rs:13:9`.

bacongobbler (approved):

bacongobbler:
Looks good! In the future, please try to refrain from adding more than what's absolutely necessary to get the work done. I don't recall the TLS certificate code ever being part of the original plan.

That being said, this all looks like a good starting point to move forward. Please mark it as ready for review and we can look into the test failures. Thanks!

flavio:
@bacongobbler, @kflansburg : I removed the "draft" status from the PR, **but** I still need your feedback about what I wrote into [this comment](https://github.com/deislabs/krustlet/pull/521#issuecomment-789942705) :pray: 

The tests are currently broken because I didn't bother to fix them, I will do that once we agree on I can get rid of the temporary `unwrap()` calls I left into the code.

kflansburg:
`new` does not _have_ to return `Self`, although it is convention. A more idiomatic approach might be to implement `TryFrom<ClientConfig>` for `Client`, which can return a `Result`. 

https://doc.rust-lang.org/std/convert/trait.TryFrom.html

flavio:
I've updated the PR to reflect the latest feedback:

* I've implemented the `TryFrom` trait for `Client`. This can be used by users who want to be 100% sure the reqwest client will obey to their configuration. The function returns a `Result` object -> users can handle `Err` returns as they want
* The `Client::new` function still returns a `Self` object. Internally it attempts to create the `Client` instance using `TryFrom` but, when the conversion fails, it will produce a warning message and instantiate the Reqwest client using the default infallible settings.

Since the `Client::new` API isn't changed, the code using this crate will keep building without changes.

With future PRs we can change the existing Krustlet code to allocate the `Client` instance using `TryFrom`, so that we can pull contents from registries secured with self-signed certificates.

kflansburg (approved):
LGTM, Thanks @flavio !

flavio:
The windows build is failing, I'll take a look at that during the weekend

flavio:
I fixed the build issue on windows. Can someone allow the CI to pick the new code?

cc @kflansburg

flavio:
I'm sorry, the build on windows failed again because of `cargo fmt` warnings. I fixed everything and did a force push.

Can you please trigger another build?

flavio:
The CI system seems to be happy with the changes, plus we got two approvals. Can we merge it? :smiley:

PR #525, opened 2021-02-19
Allow duplicate token fields in OCI auth JSON

### LINKED ISSUE #495: [OCI Distribution] Serde fails to decode auth response from Docker Hub.
It looks like Docker response with _both_ `token` and `access_token`, and `serde` complains:
```
Caused by:                                                                                                                                                                                                                                                     │✗
    duplicate field `token` at line 1 column 2160
```

jdolitsky:
When requesting an authorization token from Docker Hub, the JSON response contains both a 'token' and 'access_token'
field (with identical values). Since 'access_token' was used as an alias field, this was causing the serde deserialize to fail due to "duplicate field \`token\` at line 1 column X".

This could have been fixed by adding both fields to the struct, marking both as optional. However, this method would have meant that a successful deserialization using serde would not necessarily indicate that a token was found in either field, and would require additional logic in the caller.

To solve this, a custom deserialize implementation has been added for RegistryToken, which allows for both fields. In the case that both fields are present (i.e. Docker Hub), then the 'token' field takes precedence and 'access_token' is ignored.

Fixes #495.

P.S. I'm fairly new to Rust! Let me know if there are more idiomatic ways to do this, especially the `visit_map` function.

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -18,8 +18,10 @@ use futures_util::stream::StreamExt;
 use hyperx::header::Header;
 use log::debug;
 use reqwest::header::HeaderMap;
+use serde::de;
 use sha2::Digest;
 use std::collections::HashMap;
+use std::fmt;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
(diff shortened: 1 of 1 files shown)


kflansburg:
This is great. It looks like you are expecting the values to be ints as well as strings. Is that the case for some registries?

jdolitsky:
@kflansburg - for the 'token' and 'access_token' fields, they should always be strings. The tests using integer values for these fields is just to ensure that parsing fails if they are not strictly strings (at least one field must be a string).

Other fields, like 'expires_in' (number of seconds until token expires) should actually be integers, but they are basically ignored for now. If they are present, they do not cause the deserialization to fail.

crates/oci-distribution/src/client.rs
```diff
@@ -987,6 +1065,94 @@ mod test {
...
+        let rt = res.unwrap();
+        assert_eq!(rt.token, "abc");
+
+        // non-string fields do not break parsing
+        let text = r#"{"aaa": 300, "access_token": "xyz", "token": "abc", "zzz": 600}"#;
+        let res: Result<RegistryToken, serde_json::Error> = serde_json::from_str(&text);
+        assert!(res.is_ok());
+
+        // Note: tokens should always be strings. The next two tests ensure that if one field
+        // is invalid (integer), then parse can still succeed if the other field is a string.
```
jdolitsky, line 1104: @kflansburg - added this comment to clarify the intention

kflansburg:
Hey @jdolitsky, sorry for the delay. I think I'm still a little concerned about the complexity here. When I noticed this I originally thought that this would be a good use case for [serde's untagged enums](https://serde.rs/enum-representations.html#untagged). Did you attempt to use this and / or run into any limitations there?

jdolitsky:
@kflansburg - I did not attempt that. I'm unfamiliar with it, and it's unclear to me how that works based on the link you shared... happy to take a pass with a little more information.

Another option I considered was to make 2 fields, `token` and `access_token`, then add a helper function to parse it vs. calling `serde_json::from_str` directly. The only issue I take with that is if some of these structs are eventually made public.

kflansburg:
@jdolitsky, I think something like this would work:

```rust
#[derive(Deserialize, Default)]
#[serde(untagged)]
enum RegistryToken {
    Token { token: String }
    AccessToken { access_token: String }
}

impl RegistryToken {
    pub fn token(&self) -> String {
        match self {
            RegistryToken::Token { token } => token,
            RegistryToken::AccessToken { access_token }  => access_token
        }
    }
}
```

`serde` would match in order of the enum (so DockerHub would match on `RegistryToken::Token`), and of course fail if neither are present.

crates/oci-distribution/src/client.rs
```diff
@@ -751,12 +753,88 @@ impl ClientProtocol {
...
+
+            fn visit_map<V>(self, mut map: V) -> Result<Self::Value, V::Error>
+            where
+                V: de::MapAccess<'de>,
+            {
+                let mut token = String::new();
+                loop {
+                    let next_key_result = map.next_key::<String>();
+                    if next_key_result.is_err() {
+                        return Err(de::Error::custom("Unable to parse JSON"));
```
itowlson, line 795: I believe these three lines could be written:

```
let next_key_option = map.next_key::<String>().map_err(|_| -> de::Error::custom("Unable to parse JSON"))?;
```

crates/oci-distribution/src/client.rs
```diff
@@ -751,12 +753,88 @@ impl ClientProtocol {
...
+                                REGISTRY_TOKEN_JSON_SECONDARY_FIELD
+                            );
+                            if is_secondary_field {
+                                continue; // do not overwrite value from primary with secondary
+                            }
+                        }
+                        token = val;
+                    }
+                }
+                if token.is_empty() {
```
itowlson, line 824: Are you using the empty string for "we haven't found a token"?  If so, would it be appropriate to use an `Option` instead?  Or is there a potential case where you read the token and it's empty?

crates/oci-distribution/src/client.rs
```diff
@@ -751,12 +753,88 @@ impl ClientProtocol {
...
+                        return Err(de::Error::custom("Unable to parse JSON"));
+                    }
+                    let next_key_option = next_key_result.unwrap();
+                    if next_key_option.is_none() {
+                        break; // done reading the map
+                    }
+                    let next_value_result = map.next_value::<String>();
+                    if next_value_result.is_err() {
+                        continue; // does not parse as string-to-string (e.g. 'expires_in' field)
+                    }
```
itowlson, line 804: Instead of reading the key and value separately, is it worth calling `next_entry` to get them as a tuple?  This might be more concise, but I don't know how it plays with your error/exit/continue strategy.

itowlson (commented):
@jdolitsky I don't have the domain expertise to comment properly, but I did leave some thoughts on idioms that might be relevant to `visit_map` since you specifically asked about that.  Hope these are useful even if they end up not applying here!

thomastaylor312:
@jdolitsky Friendly follow up ping

jdolitsky:
sorry about delay.

@kflansburg - that looks like a lot less code, so I am for it. If support were to be added. If wanted to add support for more fields in the future (like `expires_in`), we would just have to add more getters like `fn token` ?

@itowlson I appreciate the thorough rust review!

If there is some urgency on this, feel free to implement @kflansburg 's solution in a separate PR or on top of my branch. Or maybe I can resubmit with just the unit tests.

kflansburg:
Closed in favor of #554

PR #534, opened 2021-03-02
Move to tracing for logging

kflansburg:
This pull request: 

* Converts all library crates to remove `log` and add `tracing`. All `log` macros are replaced with identical `tracing` macros. 
* Converts all binaries to initialize the logger with `tracing-subscriber` (similar to `env-logger`), and use identical environment variable based filters.
* We use `log` feature for `tracing` so that consumers of the crate can use `log` and `env-logger` in their binaries if they wish. 

Functionally this will not make much difference right now (except for maybe slightly different output format), but we should be able to incrementally start using `tracing`'s nicer features now. Ones I'm excited about:

* The obvious ability to use spans. 
* Easy integration with tools like Jaeger.
* Automatic function instrumentation using `instrument` macro.

```rust
#[instrument]
pub fn my_function(my_arg: usize) {
    // This event will be recorded inside a span named `my_function` with the
    // field `my_arg`.
    event!(Level::INFO, "inside my_function!");
    // ...
}
```

* The ability to produce structured logs with additional fields using `tracing`'s really slick extended macro functionality:

```rust
warn!(
    field1,                      // No need to specify field name, it is the same as the Rust variable name
    ?field2,                     // Use Debug to stringify field
    %field3,                     // Use Display to stringify field
    "message: {}",
    value                        // The usual format string arguments
);
```

`</evangelism>`


crates/kubelet/src/webserver/mod.rs
```diff
@@ -4,12 +4,12 @@ use crate::provider::{NotImplementedError, Provider};
 use http::status::StatusCode;
 use http::Response;
 use hyper::Body;
+use std::convert::Infallible;
+use std::sync::Arc;
```
itowlson, line 8: This is inherited from existing code but is it worth tidying up - the doc comments in the middle of the imports?

itowlson (approved):
Looks good - thanks!  I did have one tidying suggestion (of existing layout) where you were already moving some stuff around - definitely not a must-do though.

kflansburg:
@itowlson Yep, I can get that docs comment, I'm not sure how that got there. 

Tests seem to be failing right now :(.

kflansburg:
@itowlson I've reverted to using `env_logger` in the binary, which appears to resolve the e2e test problem. I think this is fine to merge now, and I will do some follow up research on why `tracing-subscriber` was introducing a performance regression. Requested another review to confirm that this is ok.

itowlson:
This still looks okay with the WASI fix, but how does having `env_logger` in the WASI binary play with using `tracing` elsewhere?  Will this still open up the future benefits you cited?  Do we need to advise people to use `env_logger` rather than `tracing` in their own binaries (but not provider crates?)?

kflansburg:
> This still looks okay with the WASI fix, but how does having `env_logger` in the WASI binary play with using `tracing` elsewhere? Will this still open up the future benefits you cited? Do we need to advise people to use `env_logger` rather than `tracing` in their own binaries (but not provider crates?)?

It should work fine with `tracing` + `env-logger`, that was my intent when enabling the `log` feature for `tracing`, which provides interoperability with crates using `log` and `env-logger`. We should be able to start using these features, although spans will not be emitted I believe. Without knowing why `env-logger` works and `tracing-subscriber` doesn't, I'm not sure if we need to tell users to use `env-logger`. My hope is get `tracing-subscriber` working shortly.

thomastaylor312 (approved):
LGTM. Let me know if you need any more help with figuring out why the tracing-subscriber is causing issues for you

PR #497, opened 2021-01-26
Running Krustlet on Linode (LKE)

DazWilkin:
I guess I need to update the referring docs to point to this and the DO instructions too?

PR #535, opened 2021-03-02
DigitalOcean docs use public networking; remove

DazWilkin:
Per: https://github.com/deislabs/krustlet/pull/497#issuecomment-782246494


bacongobbler:
Hi @DazWilkin! I appreciate you taking the time to write up a PR for this. However I don't think we should purge all this hard work from the documentation! It's still quite useful for those who want to experiment with Krustlet in a development environment.

What if I wrote some additional documentation around this howto guide explaining it's is a work-in-progress, point out where we need to address some of the security issues, and suggest that community contributions are welcome?

DazWilkin:
Your earlier feedback was well taken.

I think it's correct (and understandable particularly for Microsoft) that you have the requirement of more stringent security.

I don't mind at all if these are deleted; I think I blogged about running Krustlet on DO and so it's available there.

I agree that it would be better to have someone who's more proficient in DO and Linode to write and maintain (!) these docs.

DazWilkin:
... and thank you for taking the time to comment.

thomastaylor312:
I agree with @bacongobbler here! I think the work you did is great, we just need to add a bit to clarify that these are completely for development purposes. Someone will surely find these handy

PR #538, opened 2021-03-06
Use WatchStream for Manifest.

### LINKED ISSUE #536: [krator/master] Watch on Manifest not working correctly anymore
Hi,
currently using krator for building an operator. Before the big Tokio 1.0 update, I was able to watch for modification events with code more or less similar to this:

https://github.com/deislabs/krustlet/blob/master/crates/kubelet/src/container/state.rs#L45-L48

... short snippet I used for testing:
```
        let mut m = manifest.clone();
        while let Some(current) = m.next().await {
            info!(
                "manifest for {} was updated",
                current.metadata.name.unwrap()
            );
        }
```

this used to work, but now it does not receive an event when I update the resource.

kflansburg:
Closes #536 

@thomastaylor312 I know you didn't want to add all of these trait constraints, but this feels like the most stable solution given the trouble we've had with this (presumably `WatchStream` will be kept up to date and reliable by the Tokio team). In practice, this only adds one new trait to `Operator::Manifest`: `Unpin`, and required no changes outside of Krator.

### DIFF
--- a/crates/krator/src/manifest.rs
+++ b/crates/krator/src/manifest.rs
@@ -1,20 +1,30 @@
 use core::pin::Pin;
 use core::task::{Context, Poll};
 use tokio::sync::watch::{channel, Receiver, Sender};
-use tokio_stream::Stream;
+use tokio_stream::{wrappers::WatchStream, Stream};
 
 /// Wrapper for `ObjectState::Manifest` type which reflects
 /// the latest version of the object's manifest.
-#[derive(Clone)]
-pub struct Manifest<T: Clone> {
+pub struct Manifest<T: Clone + Sync + Send + std::marker::Unpin + 'static> {
     rx: Receiver<T>,
+    stream: WatchStream<T>,
 }
 
-impl<T: Clone> Manifest<T> {
+impl<T: Clone + Sync + Send + std::marker::Unpin + 'static> Clone for Manifest<T> {
+    fn clone(self: &Manifest<T>) -> Manifest<T> {
+        Manifest {
+            rx: self.rx.clone(),
+            stream: WatchStream::new(self.rx.clone()),
+        }
+    }
+}
+
+impl<T: Clone + Sync + Send + std::marker::Unpin + 'static> Manifest<T> {
     /// Create a new Manifest wrapper from the initial object manifest.
     pub fn new(inner: T) -> (Sender<T>, Self) {
         let (tx, rx) = channel(inner);
-        (tx, Manifest { rx })
+        let stream = WatchStream::new(rx.clone());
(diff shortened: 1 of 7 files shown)


crates/krator/src/manifest.rs
```diff
@@ -1,20 +1,30 @@
...
     rx: Receiver<T>,
+    stream: WatchStream<T>,
 }
 
-impl<T: Clone> Manifest<T> {
+impl<T: Clone + Sync + Send + std::marker::Unpin + 'static> Clone for Manifest<T> {
+    fn clone(self: &Manifest<T>) -> Manifest<T> {
+        Manifest {
+            rx: self.rx.clone(),
+            stream: WatchStream::new(self.rx.clone()),
```
kflansburg, line 17: I know this looks super hacky, but there does not appear to be a way to get the `latest()` value from a `WatchStream` because the `Receiver` is stored in a private field. `Receiver` should be pretty lightweight to clone, and calling `borrow()` does not appear to update the `version` stored, so there shouldn't be any issue with these two being "out of sync".

crates/krator/src/runtime.rs
```diff
@@ -155,8 +155,8 @@ impl<O: Operator> OperatorRuntime<O> {
                         }
                         match manifest_tx.send(manifest) {
                             Ok(()) => (),
-                            Err(e) => {
-                                warn!("Unable to broadcast manifest update: {:?}", e);
+                            Err(_) => {
+                                debug!("Manifest receiver hung up, exiting.");
```
kflansburg, line 159: It turned out that this never happened because all of our receivers were stuck. Now that they exit properly, this happens, but should not be considered an error. I updated the three places where this type of behavior was logged.

crates/krator/src/manifest.rs
thomastaylor312: Can you add this as a comment in the code? I think this is useful context to have down the road

crates/krator/src/manifest.rs
```diff
@@ -1,20 +1,30 @@
...
 /// the latest version of the object's manifest.
-#[derive(Clone)]
-pub struct Manifest<T: Clone> {
+pub struct Manifest<T: Clone + Sync + Send + std::marker::Unpin + 'static> {
     rx: Receiver<T>,
+    stream: WatchStream<T>,
 }
 
-impl<T: Clone> Manifest<T> {
+impl<T: Clone + Sync + Send + std::marker::Unpin + 'static> Clone for Manifest<T> {
```
thomastaylor312, line 13: Nit: This is getting a little long, we should probably move it to a `where` clause in all places where it occurs

thomastaylor312 (approved):

PR #539, opened 2021-03-06
RFC: Krator Controller Manager API

kflansburg:
This is trying to scope out what I think the ideal API for configuring multiple controllers and watches with Krator would be. It draws from similar APIs used by the Go Controller Runtime and `kube_runtime`. 

There are still a lot of question marks here:
* There is probably going to be a lot of dynamic dispatch of object manifests (casting to `Reference` and back) with this model. I don't really love that and I've also identified some places where I will expect runtime (rather than compile time) errors that will be annoying.
* How are objects made available in the state machine? I'm thinking a `Store` type that has a method `get<R: Resource>(name: String, namespace: Option<String>) -> Option<R>` (and possibly a stream equivalent). This might be attached to `Manifest` via an `Arc`.
* Owned objects (watched objects with `OwnerRef` pointing to a given managed object) need to generate notifications to the state machine. 
* More broadly, we probably need a better way for generating notifications for changes to the managed object as well. Should this interrupt the state machine? I think that is dangerous. Right now every state is responsible for checking for changes when able and transitioning to new states accordingly. This is most explicit but not very ergonomic. I think this is the biggest issue for our API right now. 
* Finally, notifications should probably be filtered (i.e. there are a lot of updates related to our own status patches), so some hook for deciding if an object update can be ignored.


crates/krator/src/manager.rs
```diff
@@ -0,0 +1,193 @@
...
+            _controller: controller,
+            watches: vec![],
+            owns: vec![],
+            namespace: None,
+            list_params: Default::default(),
+        }
+    }
+
+    /// Restrict controller to manage a specific namespace.
+    pub fn namespaced(&mut self, namespace: &str) {
```
thomastaylor312, line 61: Nit: Shouldn't the builder pattern all return `Self`?

crates/krator/src/store.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Stores or caches arbitrary Kubernetes objects.
+pub struct Store {
+    objects: RwLock<ResourceMap>,
+}
+
+impl Store {
+    /// Insert a new object.
+    pub async fn insert<R: 'static + k8s_openapi::Resource>(
+        &self,
+        namespace: Option<String>,
```
thomastaylor312, line 41: Can't you get both `namespace` and `name` from a `Resource`?

crates/krator/src/store.rs
```diff
@@ -0,0 +1,82 @@
...
+type ResourceMap = HashMap<ResourceKey, HashMap<ObjectKey, Box<dyn Any>>>;
+
+/// Stores or caches arbitrary Kubernetes objects.
+pub struct Store {
+    objects: RwLock<ResourceMap>,
+}
+
+impl Store {
+    /// Insert a new object.
+    pub async fn insert<R: 'static + k8s_openapi::Resource>(
```
thomastaylor312, line 39: You could avoid `async` on both functions and a `RwLock` if you make this an `&mut self`

thomastaylor312:
Ok, let me take a stab at your questions:

> There is probably going to be a lot of dynamic dispatch of object manifests (casting to Reference and back) with this model. I don't really love that and I've also identified some places where I will expect runtime (rather than compile time) errors that will be annoying.

I don't really see this yet, but based on how we have studiously avoided this, if we can keep avoiding it that would be good. However, if this is the eventual concession we have to make and what people are looking for, I think that will be ok as Rust should allow us to handle the errors well

> How are objects made available in the state machine? I'm thinking a Store type that has a method get<R: Resource>(name: String, namespace: Option<String>) -> Option<R> (and possibly a stream equivalent). This might be attached to Manifest via an Arc.

I made a suggestion in my comments to make this even simpler, but isn't this also just a reflector? Maybe it would be better to use a reflector from the `kube_runtime` crate?

> More broadly, we probably need a better way for generating notifications for changes to the managed object as well. Should this interrupt the state machine? I think that is dangerous. Right now every state is responsible for checking for changes when able and transitioning to new states accordingly. This is most explicit but not very ergonomic. I think this is the biggest issue for our API right now.

I think I need a bit more of an idea of how you think you would do this otherwise. I try to always keep in mind (and I think we feel the same way here) that state machines should always be driven by comparing desired state with current state in the machine rather than the Go method of having an old and a new object.

> Finally, notifications should probably be filtered (i.e. there are a lot of updates related to our own status patches), so some hook for deciding if an object update can be ignored.

If the same types of notifications will always be filtered (i.e. if we can detect that it was from a status patch), we should just build that in

crates/krator/src/store.rs
kflansburg: I'm expecting `Store` will be stored in an `Arc`, so I would prefer to only require `&self` and hide the `RwLock` access inside `insert` and `get`.

crates/krator/src/store.rs
kflansburg: I believe you are thinking of `kube::Resource`. `k8s_openapi::Resource` only has group, version, kind, api_version.

kflansburg:
Thanks for your comments. 

> I don't really see this yet, but based on how we have studiously avoided this, if we can keep avoiding it that would be good. However, if this is the eventual concession we have to make and what people are looking for, I think that will be ok as Rust should allow us to handle the errors well

I don't really think this is avoidable. `Store` is essentially the implementation of what I'm referring to here, where objects are stored as `Arc<dyn Any>`, and then cast back with error checking. I think since we are managing the internal `HashMap`, we can ensure that objects of a given type are always placed under the correct (group, version, kind) key in the `HashMap`. What do you think of this?


> I made a suggestion in my comments to make this even simpler, but isn't this also just a reflector? Maybe it would be better to use a reflector from the `kube_runtime` crate?

I _think_ a reflector just operates on a single resource kind. But it is a possibility. I think my question was written before I fully developed the `Store` type, and now it is mostly about whether to add _another_ argument to `State::next` or to add functionality to `Manifest` to be a one-stop-shop for the state of all the resources you are watching ( I think I prefer this option). 


> I think I need a bit more of an idea of how you think you would do this otherwise. I try to always keep in mind (and I think we feel the same way here) that state machines should always be driven by comparing desired state with current state in the machine rather than the Go method of having an old and a new object.

> If the same types of notifications will always be filtered (i.e. if we can detect that it was from a status patch), we should just build that in

I'm not really sure, I think the problem right now is that state machine authors must implement the logic to watch for changes, and decide if those need to be reacted to. This is also typically something that is only done in one or two states and not all of them. For instance, we would not plan to watch for `Pod` changes during `ImagePull`, even though that could impact which images to pull and it could take up to a minute to get to a state where we _are_ watching for changes.

I agree with you that state machines are not generally reactive, so I don't have a good answer for this. Regarding the second response above, my thought was that the user could supply a hook which would be used to insert custom logic to filter object changes (I think @soenkeliebau was asking for this). They could also define an enum with all of the supported object mutations (for instance on Pods we might only support a changed image), and the hook could emit this enum. They could then potentially write some logic for how every state should respond to every mutation. This sounds super complex, but I cant think of a more rigorous way to deal with this issue.

crates/krator/src/store.rs
thomastaylor312: Ok, sounds good to me

thomastaylor312:
> I don't really think this is avoidable. Store is essentially the implementation of what I'm referring to here, where objects are stored as Arc<dyn Any>, and then cast back with error checking. I think since we are managing the internal HashMap, we can ensure that objects of a given type are always placed under the correct (group, version, kind) key in the HashMap. What do you think of this?

> I think a reflector just operates on a single resource kind. But it is a possibility. I think my question was written before I fully developed the Store type, and now it is mostly about whether to add another argument to State::next or to add functionality to Manifest to be a one-stop-shop for the state of all the resources you are watching ( I think I prefer this option).

This is basically what _every_ k8s implementation has to do so it would be fine. I am still not understanding why `Store` needs to handle more than a single object type though. Can you explain what a controller would look like that handles more than a single kind? I figure you'd have a controller per type anyway as mixing and matching violates separation of concerns. Basically, it is getting a bit too abstract for me, could you give a simple, but concrete example to reason through with you here?

> I'm not really sure, I think the problem right now is that state machine authors must implement the logic to watch for changes, and decide if those need to be reacted to. This is also typically something that is only done in one or two states and not all of them. For instance, we would not plan to watch for Pod changes during ImagePull, even though that could impact which images to pull and it could take up to a minute to get to a state where we are watching for changes.

Random start of the week thought here: What if required state machine to be idempotent? So every time there is a change, it reruns through the graph, reconciling things at each step. So if it gets to a step like volume management and the volume is already attached, then it just moves on.

soenkeliebau:
Sorry, a bit late to the party here, but I'll briefly add my two cents - full disclosure though, have not looked at any code yet!

On a different note, Lars and I will try and join the weekly call later today and be happy to talk a bit about the route that we took with our operators.

> Random start of the week thought here: What if required state machine to be idempotent? So every time there is a change, it reruns through the graph, reconciling things at each step. So if it gets to a step like volume management and the volume is already attached, then it just moves on.

This is pretty much what our provider tries to do and which works fairly well in principle, but also triggered my wish to filter updates back in the day. I updated the PodState to reflect the state that the engine was currently in upon entering that state - and that status change then triggered the next reconciliation. Of course there are a number of fairly simple solutions for that, chief among them: simply don't update the status if no change is performed :)
In our kubelet provider we more or less sidestepped the issue, as we always delete and recreate a pod, which causes a new state machine to be created.


> More broadly, we probably need a better way for generating notifications for changes to the managed object as well. Should this interrupt the state machine? I think that is dangerous. Right now every state is responsible for checking for changes when able and transitioning to new states accordingly. This is most explicit but not very ergonomic. I think this is the biggest issue for our API right now.

For _real_ operators we took a somewhat different approach without a state machine.,It looks similar to what the operator sdk does, but offers some more tooling to assist with common tasks.

An example of the main reconcile function of one of our operators looks like this:
```rust
self.check_zookeeper_reference()
                .await?
                .then(self.context.delete_illegal_pods(...))
                .await?
                .then(self.context.wait_for_terminating_pods(...))
                .await?
                .then(self.context.delete_excess_pods(...))
                .await?
                .then(self.create_missing_pods(...))
                .await
```

The return type of each of these functions can be one of three things:
- Continue with the next function
- Stop execution and requeue in Kubernetes with a delay - this will cause the reconcile function to be started again after a delay
- Done, finish execution and don't requeue

Also, you can see that some functions are provided by the context, these are the ones which are implemented in our operator library, the _self._ methods are operator specific.

This gives us a fairly flexible coding model where we can dynamically react to what happens during the execution of these functions:
- found an illegal pod and deleted it? -> requeue to wait for deletion to finalize
- created a new pod? -> requeue
- no new pods needed to be created? -> continue to the next step

I'm not sure how much sense this makes, but I'll be happy to discuss later on the call or answer questions here.

kflansburg:
> This is basically what _every_ k8s implementation has to do so it would be fine. I am still not understanding why `Store` needs to handle more than a single object type though. Can you explain what a controller would look like that handles more than a single kind? I figure you'd have a controller per type anyway as mixing and matching violates separation of concerns. Basically, it is getting a bit too abstract for me, could you give a simple, but concrete example to reason through with you here?

I was hoping to cover this in my Rust Day talk, but since I had to cut it, I will outline the categories here. I'm not sure how familiar you ware with the Go Operator SDK, so apologies if this isn't new information. 

* A controller has a single resource `Kind` that it manages (I think this is what you were referring to). 
* A controller can have any number of resource `Kinds` that it _owns_. This would be in a case where the Operator creates other Kubernetes resources on behalf of a specific managed resource object. In this case the owned resources have an `OwnerRef` which points back at the object. The operator would watch this owned resource, and changes to it would generate notifications to the state machine of the specific object referenced by the `OwnerRef`. 
* Finally, an operator may have watched resources. These could be resources like `ConfigMaps` which are used to configure the Operator, or related resources (like I believe that service mesh controllers watch `Services`, even though they aren't directly managing / owning them). This allows you to ad-hoc query for these resources without hitting the Kubernetes API. 

So based on this, I think its reasonable to support the dynamic typing I'm suggesting (and sounds like is common practice from what you say). 


> Random start of the week thought here: What if required state machine to be idempotent? So every time there is a change, it reruns through the graph, reconciling things at each step. So if it gets to a step like volume management and the volume is already attached, then it just moves on.

Hmmm, this is pretty interesting. I have been starting to think of the state machine API as simply being an opinionated way to define the `reconcile` method from the Go Operator SDK. In that framework, idempotency makes sense. I think this also addresses classes of errors that I think we currently have where, say, we get to `Running`, encounter an `Error` and then move back to `Starting`, possibly re-running `init` containers without intending to (this is just an example, idk if that actually happens). 

I'm thinking the model would be that there is a hook on the `Operator` called `should_reconcile`, which is notified whenever the managed object or an owned object is changed. If it returns `true`, then Krator restarts the state machine from the `InitialState`. The only concern I have remaining is whether its ok to interrupt an in-progress state handler. We felt ok for delete events because the object is going away anyway, but I'm not so sure in this case.

kflansburg:
> The return type of each of these functions can be one of three things:
> 
>     * Continue with the next function
> 
>     * Stop execution and requeue in Kubernetes with a delay - this will cause the reconcile function to be started again after a delay
> 
>     * Done, finish execution and don't requeue
> 
> 
> Also, you can see that some functions are provided by the context, these are the ones which are implemented in our operator library, the _self._ methods are operator specific.
> 
> This gives us a fairly flexible coding model where we can dynamically react to what happens during the execution of these functions:
> 
>     * found an illegal pod and deleted it? -> requeue to wait for deletion to finalize
> 
>     * created a new pod? -> requeue
> 
>     * no new pods needed to be created? -> continue to the next step
> 
> 
> I'm not sure how much sense this makes, but I'll be happy to discuss later on the call or answer questions here.

This is pretty interesting to hear. Are you building controllers using `kube-runtime` for this pattern? I think this also hints at the possible need for a type of higher-level DSL for common operator tasks (although that is a bit of a rabbit hole).  

Do you think that the state machine model I described above would work for this? I would think each of those steps could be a node in the state graph, and you would be able to filter which change events from Kubernetes would restart the state machine from the beginning. 

Do you have any thoughts on interrupting the running state machine? My understanding is that Tokio would allows us to cancel a task, but it would only stop the task once it has `awaited`. So this might interrupt the task while its waiting for an API response or something.

soenkeliebau:
> This is pretty interesting to hear. Are you building controllers using kube-runtime for this pattern? I think this also hints at the possible need for a type of higher-level DSL for common operator tasks (although that is a bit of a rabbit hole).

Our implementation is based on the [controller](https://docs.rs/kube-runtime/0.51.0/kube_runtime/controller/struct.Controller.html) from kube-runtime, yes. Since we decided to stick with a _pure_ reconciliation approach this has allowed us to skip a bit of boilerplate code around managing whom to notify, deduplication and things like that. 
As for the higher level DSL, I totally agree that this would be a rabbit hole. We are absolutely not claiming to solve this for everybody here! What we are doing is creating a very opinionated abstraction layer tailored to our needs - with a little polish applied afterwards to try and keep it usable for a wider audience.

Our current operating model is that we will code in parallel on a few operators for a while and then take a step back and look at which code looks to be duplicated and then evaluate whether this makes sense to pull out into the framework. So far this has worked ok for us, but it is still somewhat early days, may still break :)

This process is the origin of the methods you see in my snippet above, all our operators needed to find eligible nodes that were missing a pod, find pods that are on nodes which should not have one anymore, etc.

>Do you think that the state machine model I described above would work for this? I would think each of those steps could be a node in the state graph, and you would be able to filter which change events from Kubernetes would restart the state machine from the beginning.

We discussed this a lot when we started looking into operators. Technically it looks like a great fit, simply add the option to jump out of the state machine and requeue on state transitions and everything is golden.

I think that a state machine doesn't add much value in this case though, I'm sorry to say. In principle you should throw away the entire state machine between reconciliations and start a fresh one. You could get away with just transitioning to the initial state when you get an update, but that leaves you with recovering on restart/crash as an issue.
It feels to me as if the state machine in this model is simply _degraded_ to just handling flow control, which seems like a bit of a waste.

>Do you have any thoughts on interrupting the running state machine? My understanding is that Tokio would allows us to cancel a task, but it would only stop the task once it has awaited. So this might interrupt the task while its waiting for an API response or something.

I don't think I'd simply interrupt the state machine at random moments during execution, rather the operator implementation should have specific breakpoints, where it is safe to jump out. In our model this is when _Requeue_ is returned.
We try to implement our operators in a way that they take very small steps toward the target state and requeue as soon as a change was made.
To spin up a three node cluster for example the operator would start its reconciliation, see that three pods need creating, write one pod and the config maps, requeue and upon restart it would notice that there is a pod that is not ready yet and requeue again until this shows as ready. Then it would create the second pod ... rinse and repeat...

This creates fairly small steps towards the end goal and I wouldn't feel comfortable interrupting inside of such a small step tbh. 


All of what I am talking about here is open source btw, so please feel free to take a look!

The operator framework can be found at https://github.com/stackabletech/operator-rs

Many of the _value add_ functions are currently in review, so you'd need to check out this [PR](https://github.com/stackabletech/operator-rs/pull/78)

The [ZooKeeper operator](https://github.com/stackabletech/zookeeper-operator) would be one example that uses this framework

We have not yet gotten around to writting to many docs for this, it is high on our list, but at the moment it would be a bit of a moving target to be honest - we will get around to it though!
In the meantime, feel free to ask as many questions as you want either here or on Slack, I'll be happy to answer everything I can and get Lars for those I cannot answer :)

kflansburg:
@soenkeliebau Very interesting. Thanks for this information. I'll give that PR you shared a look.

kflansburg:
Hmmm, deleting the base branch accidentally closed the PR.

kflansburg:
@thomastaylor312 I think this may be at a good stopping point for this PR, I have basic functionality working and want to keep the scope of the PR sane.

What works:
* You can configure an arbitrary number of controllers and register them with the manager.
* You can configure watchers for additional resources, with optional namespace or list params.
* `Store` is available as a public field `store` on `Manifest` to allow querying of watched resources.
* Moose example and Krustlet use the new API. 

What doesn't work:
* Webhook API is not implemented. I have made it so you cannot accidentally compile the new manager with the webhook feature enabled. For now the original `OperatorRuntime` API should be used. #601 
* We still haven't settled on an API for notifications so managed and owned objects do not generate notifications on change. #600 
* I'd like to expand the documentation with some more examples.
* Managed objects do not get cached in the store, but at some point we may want to eliminate the `Manifest` type and consolidate that behavior (once notifications are figured out). 
* Currently, a watcher is created for each configured watched resource, I think in the future we can deduplicate these (with varying degrees of complexity) #602

crates/krator/src/store.rs
```diff
@@ -0,0 +1,136 @@
...
+        &self,
+        namespace: Option<&str>,
+        name: &str,
+        object: R,
+    ) {
+        let mut objects = self.objects.write().await;
+        let key = GroupVersionKind::gvk(R::GROUP, R::VERSION, R::KIND).unwrap();
+        let resource_objects = (*objects).entry(key).or_insert_with(HashMap::new);
+        let object_key = ObjectKey::new(namespace.map(|s| s.to_string()), name.to_string());
+        resource_objects.insert(object_key, serde_json::to_value(&object).unwrap());
```
thomastaylor312, line 67: Just to clarify, are you storing these as JSON values to avoid having to use `Box<dyn Any>`? Are there going to be any performance problems to worry about?

crates/krator/src/store.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Clear cache for specified object kind.
+    pub async fn reset(&self, gvk: &GroupVersionKind) {
+        let mut objects = self.objects.write().await;
+        let key = gvk.clone();
+        let resource_objects = (*objects).entry(key).or_insert_with(HashMap::new);
+        resource_objects.clear();
+    }
+
+    /// Delete a cached object.
+    pub async fn delete_any(
```
thomastaylor312, line 79: Nit: Maybe just call this `delete`? `delete_any` made me think I'd find a `delete` as well

crates/krator/src/store.rs
```diff
@@ -0,0 +1,136 @@
...
+                            R::GROUP,
+                            R::VERSION,
+                            R::KIND,
+                            e
+                        );
+                    }
+                },
+                None => Ok(None),
+            },
+            // TODO: Should this be an error since we probably arent tracking that resource type?
```
thomastaylor312, line 132: I think it is an appropriate behavior as `HashMap` does the same thing. I would document this behavior in the function documentation though, as well as describing when an error will be returned

crates/krator/src/util.rs
```diff
@@ -0,0 +1,123 @@
...
+/// # let pod = Pod::default();
+/// # let dynamic_object = serde_json::from_str::<DynamicObject>(&serde_json::to_string(&pod).unwrap()).unwrap();
+/// let pod = concrete_object::<Pod>(dynamic_object).unwrap();
+/// ```
+pub fn concrete_object<R>(dynamic_object: DynamicObject) -> anyhow::Result<R>
+where
+    R: DeserializeOwned,
+{
+    // TODO: This sucks
+    let manifest = serde_json::to_string(&dynamic_object)?;
```
thomastaylor312, line 86: Can't you use `to_value` like you did in `Store`? Also, A+ TODO comment

thomastaylor312 (commented):
I actually really like how this all turned out. Mostly some nits with a few items of discussion before I LGTM.

@bacongobbler should probably take a look at this too due to its size.

@soenkeliebau any other feedback here since you've been involved in the discussion for a while?

crates/krator/src/store.rs
thomastaylor312: *an appropriate behavior to return `None`. The store doesn't care about missing types it isn't tracking, just the controller does

crates/krator/src/manager/controller.rs
```diff
@@ -0,0 +1,197 @@
...
+    T: Resource<DynamicType = ()> + serde::de::DeserializeOwned + Clone + Send + 'static
+{
+}
+
+impl<O: Operator> ControllerBuilder<O> {
+    /// Create builder from operator singleton.
+    pub fn new(operator: O) -> Self {
+        ControllerBuilder {
+            controller: operator,
+            watches: vec![],
```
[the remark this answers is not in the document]
kflansburg: So, I think this requires `Default` to be implemented for `ControllerBuilder`, which I don't think is easy because of the `controller` field (certainly can't derive). Am I missing something?

crates/krator/src/manager/controller.rs
```diff
@@ -0,0 +1,197 @@
...
+        self
+    }
+
+    /// Watch all objects of given kind R. Cluster scoped and no list param
+    /// restrictions.
+    pub fn watches<R>(mut self) -> Self
+    where
+        R: Watchable,
+    {
+        self.watches.push(Watch::new::<R>(None, Default::default()));
```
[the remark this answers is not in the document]
kflansburg: I think for the sake of consistency I'm going to leave it. I don't want these methods to silently depend on each other.

crates/krator/src/manager/tasks.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Task for executing a single Controller / Operator. Listens for
+/// [DynamicEvent](crate::util::DynamicEvent) on a
+/// [channel](tokio::sync::mpsc::channel) and forwards them to a Krator
+/// [OperatorRuntime](crate::OperatorRuntime).
+///
+/// # Errors
+///
+/// A warning will be logged if a `DynamicEvent` cannot be converted to a
+/// concrete `Event<O::Manifest>`.
+pub async fn launch_runtime<O: Operator>(
```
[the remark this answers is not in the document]
kflansburg: These don't need to be exported, but I wanted them to be in the docs. I see thats actually now possible with `--document-private-items` so I will make these private again.

crates/krator/src/manager/watch.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl Watch {
+    pub fn new<
+        R: Resource<DynamicType = ()> + serde::de::DeserializeOwned + Clone + Send + 'static,
+    >(
+        namespace: Option<String>,
+        list_params: ListParams,
+    ) -> Self {
+        let gvk = GroupVersionKind::gvk(&R::group(&()), &R::version(&()), &R::kind(&())).unwrap();
```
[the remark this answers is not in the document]
kflansburg: This only errors if `version` or `kind` are empty strings. I think since we are getting those values from a type that implements `Resource`, and not just arbitrary input, its reasonable to expect them to be well-formed. Thoughts?

crates/krator/src/runtime.rs
```diff
@@ -149,7 +139,7 @@ impl<O: Operator> OperatorRuntime<O> {
                 Ok(())
             }
             ObjectEvent::Deleted { name, namespace } => {
-                let key = ObjectKey::new(&name, &namespace);
+                let key = ObjectKey::new(namespace.clone(), name.clone());
```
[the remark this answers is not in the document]
kflansburg: I need them for constructing the `ObjectEvent` below.

crates/krator/src/store.rs
kflansburg: I was originally planning to store a `Box<dyn Any>`, but there is kind of a convoluted reason this won't work:

1. I start with `Watch` which just has the `GroupVersionKind` information for the resource and not the actual type. 
2. I construct a `kube_runtime::Watcher` using this which yeilds `DynamicObject`.
3. `DynamicObject` stores the data as `serde_json::Value`. 
4. I won't have the type information again to convert back until the user queries it. 

It seemed reasonable to use `serde_json::Value` since that is how `kube` appears to be trending for storing type-erased objects and we are tightly integrated with that crate. I'm not totally sure about performance but I definitely dont like the serialization and de serialization that happens when the user fetches an object. 

There may also be some changes in the pipeline for the `kube` crate that will simplify certain things.

crates/krator/src/store.rs
kflansburg: My convention here is that `_any` indicates that they only have the `GroupVersionKind` and not the concrete type, a remnant from when I was using `Box<dyn Any>` to store things. This isn't really intended to be a public API so I will just make it private to avoid confusion.

kflansburg: I will also rename to `_gvk`.

crates/krator/src/store.rs
kflansburg: Ok, added documentation, and also realized there is a bug :( #603

crates/krator/src/util.rs
kflansburg: Ok, yeah that is probably slightly better.

crates/krator/src/manager/controller.rs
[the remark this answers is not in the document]
thomastaylor312: Oh right, because it is calling the parent default method.

crates/krator/src/manager/watch.rs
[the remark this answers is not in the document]
thomastaylor312: Only if it is an unrecoverable error. I can just imagine a case where there could have just been data corruption or something else on k8s side that could cause an error that doesn't necessarily need to panic, but I'll leave it up to you

thomastaylor312 (approved):

crates/krator/src/manager/watch.rs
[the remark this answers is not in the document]
bacongobbler: Did we come to a consensus on this answer? Just re-checking as I noticed @thomastaylor312 approved the PR

crates/krator/src/runtime.rs
```diff
@@ -102,6 +72,26 @@ impl<O: Operator> OperatorRuntime<O> {
             operator: Arc::new(operator),
             list_params,
             signal: None,
+            store: Store::new(),
+        }
+    }
+
+    pub(crate) fn new_with_store(
```
bacongobbler, line 79: Instead of a new constructor, would it make sense to follow a builder pattern here? As in should we provide a new OperatorBuilder type that looks like this:

```
let builder = Operator::builder().add_store(Store::new()).build()?;
```

Kinda like https://docs.rs/reqwest/0.11.2/reqwest/struct.ClientBuilder.html

That way we won't have 1,000,001 different `new_with_*` functions as part of the API.

I see we're already following this pattern with the ControllerBuilder API.

bacongobbler (commented):
Left a few open questions. Code here looks great!

bacongobbler (approved):
oops forgot to LGTM

crates/krator/examples/moose.rs
```diff
@@ -528,6 +525,21 @@ Running moose example. Try to install some of the manifests provided in examples
     
     "#
     );
-    runtime.start().await;
+
+    // New API does not currently support Webhooks, so use legacy API if enabled.
```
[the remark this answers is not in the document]
kflansburg: I think the webhooks will always be optional since they bring in some dependencies. The `OperatorRuntime` API will probably be removed once I get webhooks working with the new API. I think Krator is still a ways away from 1.0, so that will probably happen before 1.0.

crates/krator/src/manager/watch.rs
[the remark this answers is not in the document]
kflansburg: I'd like to leave this for now. I don't see a scenario where this would be malformed and also something we could recover from.

crates/krator/src/runtime.rs
kflansburg: I would normally agree, but I think the public API for this will disappear soon, and the `new` method will be changed require the `Store` argument as well.

crates/krator/src/manager/watch.rs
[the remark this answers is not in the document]
kflansburg: Actually, the real answer here is that I'm opening a PR for `kube` 0.54 in a few minutes, and this method no longer returns a result starting in 0.53.

PR #547, opened 2021-03-22
Pod conditions added to the StatusBuilder in kubelet

### LINKED ISSUE #546: Allow adding pod conditions to the pod status
The [StatusBuilder](https://github.com/deislabs/krustlet/blob/v0.6.0/crates/kubelet/src/pod/status.rs#L155-L205) for the pod status in kubelet does not provide a method to add [pod conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions).

We would like to use pod conditions to have more insight into the status of our pods.

siegfriedweber:
Closes #546 

I did not create helper functions like `make_status_with_conditions` or `make_status_with_containers_and_conditions` because the builder already provides a convenient way to create a status with whatever information is at hand.


kflansburg (approved):
LGTM

PR #548, opened 2021-03-24
fix(bootstrap): Handle busybox date

yanganto:
- fix bootstrap script with busybox date
- the script is tested on NixOS with BusyBox v1.32.0


thomastaylor312 (approved):
Awesome! Thank you so much for fixing this

PR #554, opened 2021-03-30
Support DockerHub OCI Authentication

### LINKED ISSUE #495: [OCI Distribution] Serde fails to decode auth response from Docker Hub.

kflansburg:
Closes #495 

I've preserved the commit history and tests from #525, but I cannot push to that branch so I'm opening a new PR. 

cc @jdolitsky

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -17,8 +17,10 @@ use futures_util::future;
 use futures_util::stream::StreamExt;
 use hyperx::header::Header;
 use reqwest::header::HeaderMap;
+use serde::de;
 use sha2::Digest;
 use std::collections::HashMap;
+use std::fmt;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tracing::debug;
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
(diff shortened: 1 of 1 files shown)


crates/oci-distribution/src/client.rs
```diff
@@ -751,15 +752,24 @@ impl ClientProtocol {
...
 /// A token granted during the OAuth2-like workflow for OCI registries.
-#[derive(serde::Deserialize, Default)]
-struct RegistryToken {
-    #[serde(alias = "access_token")]
-    token: String,
+#[derive(Deserialize)]
+#[serde(untagged)]
+#[serde(rename_all = "snake_case")]
+enum RegistryToken {
+    Token { token: String },
```
kflansburg, line 759: Note: I believe some registries send _both_ fields. I originally tried 

```rust
#[derive(Deserialize)]
#[serde(rename_all = "snake_case")]
enum RegistryToken {
    Token(String),
    AccessToken(String)
}
```

but this would cause an error if both fields are present. The `untagged` approach that I use here simply selects the first match without error.

thomastaylor312:
I'd prefer that @jdolitsky, @radu-matei, or @bacongobbler take a look at this given their experience with OCI

jdolitsky (approved):
If those tests are passing, I'd say this is a good solution

botieking98:
Hi guys! Will this PR should be merged?

kflansburg:
> Hi guys! Will this PR should be merged?

@itowlson please review.

crates/oci-distribution/src/client.rs
bacongobbler: It's an odd case but not unheard of, yes.

bacongobbler (approved):
Looks great. Thank you for the extensive list of test cases.

PR #557, opened 2021-03-31
Clean-up phase in krator fixed where a channel was closed prematurely

### LINKED ISSUE #556: Channel in krator is closed prematurely
When a resource is deleted then krator logs the following warning:

```
[WARN  krator::runtime] Error dispatching object event. error=channel closed
```

The manifest receiver in `krator::runtime::run_object_task` is consumed on cleanup and the channel is closed ([krator/src/runtime.rs#L378](https://github.com/deislabs/krustlet/blob/v0.7.0/crates/krator/src/runtime.rs#L378)). The resource is deleted afterwards ([krator/src/runtime.rs#L396](https://github.com/deislabs/krustlet/blob/v0.7.0/crates/krator/src/runtime.rs#L396)) which emits an `Apply` event with the new `deletion_timestamp` and a `Deleted` event. When the `Apply` event is handled then the resource task is terminated because the "Manifest receiver hung up" ([krator/src/runtime.rs#L191-L192](https://github.com/deislabs/krustlet/blob/v0.7.0/crates/krator/src/runtime.rs#L191-L192)). When the `Deleted` event is handled then the warning is logged because the channel to the resource task is already closed.

A possible solution would be to clone the manifest receiver ([krator/src/runtime.rs#L378](https://github.com/deislabs/krustlet/blob/v0.7.0/crates/krator/src/runtime.rs#L378)):

```
match operator.deregistration_hook(manifest.clone()).await {
```

Then the resource task does not terminate prematurely, can process the `Deleted` event and terminates afterwards.

siegfriedweber:
Fixes #556

### DIFF
--- a/crates/krator/src/runtime.rs
+++ b/crates/krator/src/runtime.rs
@@ -156,6 +156,7 @@ impl<O: Operator> OperatorRuntime<O> {
         let (sender, mut receiver) = tokio::sync::mpsc::channel::<Event<O::Manifest>>(128);
 
         let deleted = Arc::new(Notify::new());
+        let deleted_event = Arc::new(Notify::new());
 
         let (manifest, object_state) = match initial_event {
             Event::Applied(manifest) => {
@@ -167,6 +168,7 @@ impl<O: Operator> OperatorRuntime<O> {
 
         let (manifest_tx, manifest_rx) = Manifest::new(manifest);
         let reflector_deleted = Arc::clone(&deleted);
+        let reflector_deleted_event = Arc::clone(&deleted);
 
         // Two tasks are spawned for each resource. The first updates shared state (manifest and
         // deleted flag) while the second awaits on the actual state machine, interrupts it on
@@ -206,6 +208,7 @@ impl<O: Operator> OperatorRuntime<O> {
                             manifest.namespace()
                         );
                         reflector_deleted.notify_one();
+                        reflector_deleted_event.notify_one();
                         match manifest_tx.send(manifest) {
                             Ok(()) => (),
                             Err(_) => {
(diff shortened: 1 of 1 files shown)


kflansburg:
Thanks for catching this. It sounds like keeping the manifest object around until the end of `run_object_task` has fixed the issue for you, but I would expect there to still be a race condition here. Do you think that we should `await` at the end of `run_object_task` until the `Deleted` event arrives?

siegfriedweber:
I am fairly new to Rust so treat my proposals with caution.

We have the following options:

- Do not send a manifest after the resource was deleted, i.e. remove [krator/src/runtime.rs#L207-L213](https://github.com/deislabs/krustlet/blob/v0.7.0/crates/krator/src/runtime.rs#L207-L213).
- Add another `Notify` channel, e.g. `resource_task_terminated` besides `deleted`, which is awaited at the end of `run_object_task`. The `deleted` channel is actually notified very often:
  - in the Running phase when the `deletion_timestamp` is applied
  - in the Terminated phase when the new status is applied
  - when the `run_object_task` deletes the resource and the `deletion_timestamp` without a grace period is applied
  - when the `Deleted` event is handled

kflansburg:
I think its important that `Deleted` events get sent to the channel in case one arrives when Krator was not already shutting down. I think this happens when someone runs `kubectl delete --force --grace-period=0`. 

I think my suggestion was that `Manifest` would have a new method on it that allows you to `await` for it to be deleted.

kflansburg:
@siegfriedweber Let me know what you think of my commit. I think your solution makes sense, I just want to make sure that the `Manifest` isn't dropped until the `Deleted` event is received. 

We now have a lot of `Notifies`, but I think that can be refactored later.

siegfriedweber:
I fixed a copy&paste error and moved the notification of `deleted_event` after the manifest was sent.

kflansburg (approved):
LGTM

PR #558, opened 2021-03-31
chore(*): Updates openapi objects to k8s 1.20

### LINKED ISSUE #555: Use the latest version of kubernetes
Right now our API objects are pinned to the 1.18 definitions. We should update these to the latest definitions before hitting 1.0.

thomastaylor312:
Closes #555

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ rustls-tls = [
 anyhow = "1.0"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "time"] }
 kube = { version= "0.51", default-features = false }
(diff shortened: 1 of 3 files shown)


kflansburg (approved):
LGTM, my only thought is, would we maximize compatibility by compiling against the second-latest version, rather than the latest (I believe that would guarantee support for 3 minor versions, rather than 2).

thomastaylor312:
We could actually add a feature that allows you to enable the underlying features for the SDK users. For the binaries, that could be more difficult. Ideally we should support n-2 k8s versions. I think those can be added post 1.0 without breaking anything as we see how people use it

PR #561, opened 2021-04-02
Kube 0.52

kflansburg:
Looking forward to using `DynamicObject` and `GroupVersionKind` for the new watcher API.


crates/krator/src/state.rs
```diff
@@ -1,7 +1,7 @@
 //! Used to define a state machine.
 
-use k8s_openapi::Resource;
-use kube::api::{Meta, PatchParams};
+// use k8s_openapi::Resource;
```
thomastaylor312, line 3: Left over debug?

thomastaylor312 (approved):

crates/krator/src/state.rs
kflansburg: Good catch.

PR #562, opened 2021-04-03
Change master to main in Actions Config

kflansburg:
After GitHub's automatic branch renaming, actions were not running on `main`. This should fix that.


kflansburg:
I used `main` as the name in my forked branch, which seems to be causing some problems. Closing in favor of #563

PR #568, opened 2021-04-08
Replaces the dependency on tempdir with tempfile (RUSTSEC-2018-0017)

### LINKED ISSUE #567: tempdir dependency has been marked as _unmaintained_ in RUSTSEC-2018-0017
The kubelet crate has a dependency on the tempdir crate, which has been marked as unmaintained in the [RUSTSEC](https://rustsec.org/advisories/RUSTSEC-2018-0017.html) db.

The recommendation is to use [tempfile](https://github.com/Stebalien/tempfile) instead.

This has been [flagged](https://github.com/stackabletech/agent/issues/121) for us as well, since we depend on the kubelet.

soenkeliebau:
Removes the dependency on tempdir which has been flagged as unmaintained in RUSTSEC-2018-0017.

Uses the tempfile crate instead, which has been in the dependency tree anyway, as warp depends on it.

Calls to TempDir were replaced with their equivalent from the tempfile crate.

The [documentation](https://docs.rs/tempfile/3.2.0/tempfile/#resource-leaking) for tempfile states the following: 

> tempfile will (almost) never fail to cleanup temporary resources, but TempDir and NamedTempFile will if their destructors don't run. This is because tempfile relies on the OS to cleanup the underlying file, while TempDir and NamedTempFile rely on their destructors to do so.

Which to me sounds like calling .tempfile() should be the preferred way. However as far as I can tell this does not offer creating temporary directories with a prefix. For this first iteration I kept the code functionally as is. 

I'll leave it to someone with more background knowledge of why the prefixes were chosen to evaluate whether these are actually needed?

Most occurrences of this seem to be in Windows specific code, so I was not really able to test this, `cargo check` runs fine for now.. I'll create this PR and see what CI has to say about it :)

fixes #567

### DIFF
--- a/crates/kubelet/src/volume/persistentvolumeclaim.rs
+++ b/crates/kubelet/src/volume/persistentvolumeclaim.rs
@@ -210,7 +210,7 @@ pub(crate) async fn populate(
     // TODO(bacongobbler): implement node_unstage_volume(). We'll need to
     // persist the staging_path somewhere so we can recall that information
     // during unpopulate()
-    let staging_path = Builder::new().prefix("&csi.volume_handle").tempdir()?;
+    let staging_path = Builder::new().prefix(&csi.volume_handle).tempdir()?;
     if stage_unstage_volume {
         stage_volume(&mut csi_client, &csi, staging_path.path()).await?;
     }
--- a/crates/kubelet/src/volume/persistentvolumeclaim.rs
+++ b/crates/kubelet/src/volume/persistentvolumeclaim.rs
@@ -19,7 +19,7 @@ use k8s_openapi::api::core::v1::{
 use k8s_openapi::api::storage::v1::StorageClass;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::LabelSelector;
 
-use tempdir::TempDir;
+use tempfile::Builder;
 use thiserror::Error;
 
 use crate::grpc_sock;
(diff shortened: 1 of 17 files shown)


soenkeliebau:
Not sure about that test failure, could very well be related, it fails when running `cat *.txt` with "no such file". That could very well be because of a missing tempdir ... I'm not really sure which line 27 is referred to there...

crates/kubelet/src/volume/persistentvolumeclaim.rs
```diff
@@ -210,7 +210,7 @@ pub(crate) async fn populate(
     // TODO(bacongobbler): implement node_unstage_volume(). We'll need to
     // persist the staging_path somewhere so we can recall that information
     // during unpopulate()
-    let staging_path = TempDir::new(&csi.volume_handle)?;
+    let staging_path = Builder::new().prefix("&csi.volume_handle").tempdir()?;
```
bacongobbler, line 213: We should probably be using `tokio::fs` here. That way we're not blocking the thread. Everything else is fine - mio_uds_windows is third-party code.

Also why are we quoting the literal string `&csi.volume_handle` here? Shouldn't the temporary directory use the literal volume handle name as a prefix instead?

soenkeliebau: We absolutely should not be quoting that, you are right. The danger of performing multiple **almost** identical changes in short order.

I'm afraid I don't fully understand your tokio::fs comment though. Afaik that offers no temp dir functionality, are you suggesting to create and cleanup a temp dir "manually" ?

thomastaylor312:
@bacongobbler would you also like to review again or can we merge?

soenkeliebau:
> Thanks for catching this @soenkeliebau. This is what we get for adopting feral code (that was unfortunately very needed)

It wasn't even me who caught this :)
We run the [audit-check](https://github.com/actions-rs/audit-check) action on all our repos [once a day](https://github.com/stackabletech/agent/blob/main/.github/workflows/daily_security.yml) and that automatically opens issues whenever any dependency has a RUSTSEC advisory filed against it - I just followed the dependency chain here ;)

bacongobbler (approved):

PR #570, opened 2021-04-09
Wrappes creation of temporary directory with tokio::task::spawn_blocking to avoid blocking IO in an async context

### LINKED ISSUE #569: switch blocking tempfile call to non-blocking
From https://github.com/deislabs/krustlet/pull/568#discussion_r610028439

> Underneath the hood this is doing I/O operations with the std:fs functions. However, we are in an async block, which means we shouldn't do any operations that could block. This specific line should be wrapped in a tokio::task::spawn_blocking and awaited so we don't block a running task on accident

Marking as a good first issue for people looking to learn some Rust.

soenkeliebau:
fixes #569


thomastaylor312 (approved):
I think the test failure is a flake based on what I am seeing on other PRs. I'll merge this as soon as tests pass

PR #571, opened 2021-04-09
Allow providers to implement custom shutdown behavior

### LINKED ISSUE #523: Make pod eviction on shutdown optional
Currently the Krustlet automatically cordons the node and evicts all pods when a shutdown signal is received, which for most use cases following the K8S philosophy makes total sense.

For our use-case, in which we hand pods off to systemd to run it is quite valid to keep pods around even though the node they are assigned to is not online at the moment. The most relevant use case would be a restart of the krustlet.

I am happy to create a small pull request to introduce a parameter that allows disabling this behavior if that is something the team would consider?

soenkeliebau:
This is a very simple first attempt at allowing providers to implement custom shutdown behavior.

I tried to be as minimally invasive as possible for the time being. I pass the node name into the shutdown function because I don't think the provider ever has access to the _real_ node object in a proper fashion, but I may be mistaken.
The only way I could find was to take it from the builder in the [node()](https://github.com/deislabs/krustlet/blob/b50bc34c1eafb59b59bc3eff39f643a213bea98e/crates/kubelet/src/provider/mod.rs#L105) function, but that would mean relying on the implementation detail that the kubelet sets the name on the builder before and not after passing the builder to the provider.

We could make this much more complex if we wanted, some ideas I had along the way and just want to mention here: 
- track shutdown reason (Ctrl-C, SIGINT, SIGTERM, Host shutting down, panic) and pass this into the shutdown function (this might become extremely difficult with cross plattform compatibility)
- add a shutdownstate type to the provider trait and optionally have all pods transition to this state when shutting down 

fixes #523

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -101,13 +101,7 @@ impl<P: Provider> Kubelet<P> {
         });
 
         // Periodically checks for shutdown signal and cleans up resources gracefully if caught.
-        let signal_handler = start_signal_handler(
-            Arc::clone(&signal),
-            client.clone(),
-            self.config.node_name.clone(),
-        )
-        .fuse()
-        .boxed();
+        let signal_handler = start_signal_handler(Arc::clone(&signal)).fuse().boxed();
 
         let operator = PodOperator::new(Arc::clone(&self.provider), client.clone());
         let node_selector = format!("spec.nodeName={}", &self.config.node_name);
@@ -189,16 +183,11 @@ async fn start_node_updater(client: kube::Client, node_name: String) -> anyhow::
 }
 
 /// Checks for shutdown signal and cleans up resources gracefully.
-async fn start_signal_handler(
-    signal: Arc<AtomicBool>,
-    client: kube::Client,
-    node_name: String,
-) -> anyhow::Result<()> {
+async fn start_signal_handler(signal: Arc<AtomicBool>) -> anyhow::Result<()> {
     let duration = std::time::Duration::from_millis(100);
     loop {
         if signal.load(Ordering::Relaxed) {
             info!("Signal caught.");
-            //node::drain(&client, &node_name).await?;
             // When the signal was caught we simply exit the loop here,
             // handling is done outside of this task to avoid having to
             // pass a reference to the provider here
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -117,14 +117,16 @@ impl<P: Provider> Kubelet<P> {
         };
         let mut operator_runtime = OperatorRuntime::new(&self.kube_config, operator, Some(params));
         let operator_task = operator_runtime.start().fuse().boxed();
-
         // These must all be running for graceful shutdown. An error here exits ungracefully.
         let core = Box::pin(async {
             tokio::select! {
-                res = signal_handler => res.map_err(|e| {
-                    error!("Signal handler task joined with error {:?}", &e);
-                    e
-                }),
+                res = signal_handler => match res {
+                    Ok(()) => self.provider.shutdown(&self.config.node_name).await,
+                    Err(e) => {
+                        error!("Signal handler task joined with error {:?}", &e);
+                        Err(e)
+                    }
+                },
                 _ = operator_task => {
                     warn!("Pod operator has completed");
                     Ok(())
@@ -196,7 +198,10 @@ async fn start_signal_handler(
     loop {
         if signal.load(Ordering::Relaxed) {
             info!("Signal caught.");
-            node::drain(&client, &node_name).await?;
+            //node::drain(&client, &node_name).await?;
+            // When the signal was caught we simply exit the loop here,
+            // handling is done outside of this task to avoid having to
+            // pass a reference to the provider here
             break Ok(());
         }
         tokio::time::sleep(duration).await;
--- a/crates/kubelet/src/provider/mod.rs
+++ b/crates/kubelet/src/provider/mod.rs
@@ -124,16 +124,6 @@ pub trait Provider: Sized + Send + Sync + 'static {
     ///
     /// * `node_name` - The name of the node object that was created by this Kubelet instance
     ///
-    /// # Examples
-    /// This code has been taken from the wasi provider and evicts all currently running pods
-    /// from this node upon shutdown.
-    ///
-    /// ```no_run
-    /// async fn shutdown(&self, node_name: &str) -> anyhow::Result<()> {
-    //     node::drain(&self.shared.client, &node_name).await?;
-    //     Ok(())
-    //  }
-    /// ```
     async fn shutdown(&self, node_name: &str) -> anyhow::Result<()> {
         info!("Shutdown triggered for node {}, since no custom shutdown behavior was implemented Kubelet will simply shut down now.", node_name);
         Ok(())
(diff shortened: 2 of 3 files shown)


botieking98:
I have cherry pick your patch and rebuild krustlet code, but I find that wasm instance is still stopped when krustlet process stopped. So How can I do that to keep wasm instance running always ?

kflansburg:
> I have cherry pick your patch and rebuild krustlet code, but I find that wasm instance is still stopped when krustlet process stopped. So How can I do that to keep wasm instance running always ?

I don't believe this will be possible. Krustlet executes Wasm modules inside of its own Tokio runtime threads. If Krustlet shuts down these will necessarily stop as well.

botieking98:
> > I have cherry pick your patch and rebuild krustlet code, but I find that wasm instance is still stopped when krustlet process stopped. So How can I do that to keep wasm instance running always ?
> 
> I don't believe this will be possible. Krustlet executes Wasm modules inside of its own Tokio runtime threads. If Krustlet shuts down these will necessarily stop as well.

Thank you for reply ! But I still want to know how to manage the wasm instances (threads) created by wasmtime, as we know kubelet  can do it but krustlet cannot.

bacongobbler:
See https://github.com/deislabs/krustlet/issues/572#issuecomment-816693168. Let’s continue the conversation there so we can focus on the review. Thanks.

crates/kubelet/src/kubelet.rs
```diff
@@ -196,7 +198,10 @@ async fn start_signal_handler(
     loop {
         if signal.load(Ordering::Relaxed) {
             info!("Signal caught.");
-            node::drain(&client, &node_name).await?;
+            //node::drain(&client, &node_name).await?;
```
thomastaylor312, line 201: Nit: left over code?

thomastaylor312 (approved):
Leaving for @kflansburg to approve as well since he was involved in the original discussion

crates/kubelet/src/kubelet.rs
```diff
@@ -187,16 +183,14 @@ async fn start_node_updater(client: kube::Client, node_name: String) -> anyhow::
 }
 
 /// Checks for shutdown signal and cleans up resources gracefully.
-async fn start_signal_handler(
-    signal: Arc<AtomicBool>,
-    client: kube::Client,
-    node_name: String,
-) -> anyhow::Result<()> {
+async fn start_signal_handler(signal: Arc<AtomicBool>) -> anyhow::Result<()> {
```
kflansburg, line 186: I think this task should be removed entirely?

crates/kubelet/src/kubelet.rs
```diff
@@ -117,14 +111,16 @@ impl<P: Provider> Kubelet<P> {
...
         let operator_task = operator_runtime.start().fuse().boxed();
-
         // These must all be running for graceful shutdown. An error here exits ungracefully.
         let core = Box::pin(async {
             tokio::select! {
-                res = signal_handler => res.map_err(|e| {
-                    error!("Signal handler task joined with error {:?}", &e);
-                    e
-                }),
+                res = signal_handler => match res {
```
kflansburg, line 117: I think its weird to have a task that just waits and does the actual work here. We are also implicitly requiring that none of the other arms of this `select!` will ever complete first. 

Can we run `provider.shutdown()` after the `try_join!` on 134?.

crates/wasi-provider/src/lib.rs
```diff
@@ -183,6 +184,12 @@ impl Provider for WasiProvider {
     fn volume_path(&self) -> Option<PathBuf> {
         Some(self.shared.volume_path())
     }
+
+    // Evict all pods upon shutdown
+    async fn shutdown(&self, node_name: &str) -> anyhow::Result<()> {
```
kflansburg, line 189: @thomastaylor312 Do you think we should just never evict pods to match the official Kubelet, or are we acknowledging here that our Kubelets function differently because the tasks go away when we shut down (unlike containers which may continue running).

crates/kubelet/src/provider/mod.rs
```diff
@@ -110,6 +110,25 @@ pub trait Provider: Sized + Send + Sync + 'static {
...
+    ///
+    /// There are currently no mechanisms in place to propagate the shutdown trigger to the state
+    /// machine, providers will have to implement this themselves via a [`std::sync::mpsc::channel`]
+    /// or the shared state or something similar.
+    ///
+    /// # Arguments
+    ///
+    /// * `node_name` - The name of the node object that was created by this Kubelet instance
+    ///
+    async fn shutdown(&self, node_name: &str) -> anyhow::Result<()> {
```
kflansburg, line 127: So if we wanted to implement something like the new behavior in 1.21, we could send the `stop()` command to our Pod handles, without instructing the Pods to shut down via the Kubernetes API. We probably can't wait for the state machines to exit though...

kflansburg (changes requested):
I think I'd just like to review the necessity of that one task, and here some feedback on the behavior for the Wasi provider. Otherwise looks good.

crates/kubelet/src/kubelet.rs
soenkeliebau: I wondered about that as well but figured it was maybe in order to be able to add additional shutdown handling down the line.
Splitting this out into a handler communicating via a signal would allow implementing additional handlers that watch for a shutdown CRD, look for local files, offer a rest API or other stuff fairly easily. 
Just start an additional task with the new handling that sets the shutdown AtomicBool to true.

Come to think of it, we could allow providers to define a closure that can handle custom shutdown triggers and spin up a task running this in the Kubelet.  I'm not advocating for this, just a thought that came to me pretty much while writing this :)

crates/kubelet/src/kubelet.rs
soenkeliebau: I wondered about that as well but figured it was maybe in order to be able to add additional shutdown handling down the line.
Splitting this out into a handler communicating via a signal would allow implementing additional handlers that watch for a shutdown CRD, look for local files, offer a rest API or other stuff fairly easily.
Just start an additional task with the new handling that sets the shutdown AtomicBool to true.

Come to think of it, we could allow providers to define a closure that can handle custom shutdown triggers and spin up a task running this in the Kubelet. I'm not advocating for this, just a thought that came to me pretty much while writing this :)

crates/wasi-provider/src/lib.rs
thomastaylor312: The latter. Each one can do something different. I was going to change this provider later to not evict anything as we don't need to. However, your OCI provider will definitely want to do some cleanup

crates/kubelet/src/provider/mod.rs
thomastaylor312: I mean, you technically could wait on the state machines. But it depends on how you have things wired up

crates/kubelet/src/kubelet.rs
soenkeliebau: Just for shiggles I've written down my idea here:
https://github.com/soenkeliebau/krustlet/commit/7c8d904c4f732b2cfec72771fb7bfa0094a98e60

kflansburg (approved):
Sorry for the delay, work has been crazy. 

Ok, I think this can be merged. I'm actually thinking I will move the "shutdown hook" into Krator at some point because I think that's very useful. 

I will need to play with the Wasi provider with eviction turned off because I'm still not convinced that is desirable. Let me know if you want to discuss @thomastaylor312 (or maybe its none of my business).

thomastaylor312:
> I will need to play with the Wasi provider with eviction turned off because I'm still not convinced that is desirable. Let me know if you want to discuss @thomastaylor312 (or maybe its none of my business).

FWIW this should be fine. In fact, if all the node does is restart, we want those pods to remain there and pick up where they left off (by restarting the modules when the node sees them). K8s tries to move them if the node is down too long

Issue #573, opened 2021-04-12
[krator] krator with admission-webhook feature does not compile anymore


kesselborn:
I am preparing a PR which will need to fix this

```
$ cargo build --features=admission-webhook
   Compiling tokio-io-timeout v1.1.1
   Compiling h2 v0.3.1
   Compiling tower v0.4.6
   Compiling k8s-openapi v0.11.0
   Compiling tokio-tungstenite v0.13.0
   Compiling hyper v0.14.4
   Compiling hyper-tls v0.5.0
   Compiling hyper-timeout v0.4.1
   Compiling warp v0.3.0
   Compiling kube v0.52.0
   Compiling kube-runtime v0.52.0
   Compiling krator v0.2.0 (/Users/kesselborn/dev/innoq/krustlet/crates/krator)
error[E0432]: unresolved import `kube::api::Meta`
 --> crates/krator/src/admission.rs:4:5
  |
4 | use kube::api::Meta;
  |     ^^^^^^^^^^^^^^^ no `Meta` in `api`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `krator`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed

```

kflansburg:
Thanks for finding this. I missed this when updating the `kube` crate. I will try fix it later today.

PR #574, opened 2021-04-12
[wip] Adjust code to changed kube_derive

kesselborn:
This tries to fix deislabs/krustlet#573

The kube-derive macro now implements `kube::Resource` instead of
`k8s_openapi::Resource` and and `k8s_openapi::Metadata`.

I think these adjustments should be ok, `however: kube::Resource` does not
have the method `metadata_mut()` and I couldn't fix the `resync` method
@kflansburg

I added the compilation of the example app to the github actions
workflow so this kind of error get get caught early

### DIFF
--- a/crates/krator/src/runtime.rs
+++ b/crates/krator/src/runtime.rs
@@ -9,7 +9,6 @@ use tokio::sync::Notify;
 use tracing::{debug, error, info, trace, warn};
 
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
-use k8s_openapi::Metadata;
 use kube::{
     api::{Api, ListParams, Resource},
     Client,
--- a/crates/krator/src/runtime.rs
+++ b/crates/krator/src/runtime.rs
@@ -8,7 +8,6 @@ use tokio::sync::mpsc::Sender;
 use tokio::sync::Notify;
 use tracing::{debug, error, info, trace, warn};
 
-use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use kube::{
     api::{Api, ListParams, Resource},
     Client,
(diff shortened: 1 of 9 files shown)


kflansburg:
Thanks for tackling this. Didn't realize it would be this complex. 

* Good call adding it to the CI tests.
* Looks like the `metadata_mut` issue will be addressed in upcoming `kube` release https://github.com/clux/kube-rs/pull/486. 
* Good observation that we don't really need `k8s_openapi::Resource` trait bound anymore. 

I'll take a crack at eliminating the `metadata_mut` entirely.

kflansburg (approved):
@kesselborn I have removed the need for `metadata_mut()` by simplifying the channel used in the runtime module. 

If this looks good to you, I think we can merge.

crates/krator/src/runtime.rs
```diff
@@ -90,16 +113,16 @@ impl<O: Operator> OperatorRuntime<O> {
...
-        match &event {
-            Event::Applied(object) => {
-                let key: ObjectKey = object.into();
+    async fn dispatch(&mut self, event: ObjectEvent<O::Manifest>) -> anyhow::Result<()> {
+        match event {
+            ObjectEvent::Applied(object) => {
+                let key: ObjectKey = (&object).into();
                 // We are explicitly not using the entry api here to insert to avoid the need for a
                 // mutex
                 match self.handlers.get_mut(&key) {
```
kesselborn, line 122: a `get` should be enough here, shouldn't it?

crates/krator/src/runtime.rs
```diff
@@ -55,13 +54,37 @@ impl<R: Resource> From<&Event<R>> for PrettyEvent {
...
+                namespace: object.namespace(),
+            },
+            ObjectEvent::Deleted { name, namespace } => PrettyEvent::Deleted {
+                name: name.to_string(),
+                namespace: namespace.clone(),
+            },
+        }
+    }
+}
+
```
kesselborn, line 80: I know it wasn't introduced here, but what is `PrettyEvent` actually used for? Is this just for tracing purposes?

kesselborn (commented):
lgtm

crates/krator/src/runtime.rs
kflansburg: Yes, just makes it easier to debug events without dumping the entire object spec.

crates/krator/src/runtime.rs
kflansburg: I think you are right, Tokio 0.3 made it so you don't need a mutable reference when calling `Sender::send`. Surprised Clippy doesnt complain.

PR #575, opened 2021-04-12
fix(e2e): Removes old log cat step

thomastaylor312:
This was causing a spurious failure when e2e tests failed


kflansburg:
Still flaking?

thomastaylor312:
This is the same flake I saw on another PR. The logs are not quite clear as to what is happening. The krustlet process is dying, but the logs aren't clear why

technosophos (approved):
This looks fine to me.

PR #577, opened 2021-04-14
Feature/admission webhook improvements

kesselborn:
This PR adds some conveniences around the admission webhook -- everytime I needed to implement a webhook, I was quite annoyed about the many moving parts that have to be coordinated.
This looks like a lot of code but there is quite a bit of documentation and test code included + the derive code is more or less trivial but annoying to implement for each project individually.
It's probably easiest to look at each commit indivually as they are split up between `krator` and `krator_derive`.
This is still blocked until #574  (or the bug it that one tries to fix) is fixed, but perhaps people interested can already have a look at it.
/cc @kflansburg


kflansburg:
Hey @kesselborn thanks for putting this together. Something like this was definitely on my roadmap. 

This will probably take some time to review, but I do have some questions. The thing that originally prevented me from doing something like this in the moose example is that it requires the operator itself to have RBAC permissions to make these configuration changes, which I don't believe is best practice. Does this require those elevated permissions? What are your thoughts on this? I had planned to implement some sort of tooling like a Cargo plugin to deploy the operator and make this configuration.

kesselborn:
yes: if you want to do it the way I added it to the moose example, you would need elevated permissions ... but as it is only an example, I think that's ok + you can see how you can use those functions.
In a current operator I am writing, you can just output the crd when calling it with an option.

That's why I added the Type `WebhookResources` as a wrapper type and implemented the `Display` crate. This is analagous to the `crd()` command from the `kube_derive` macro ... so you can either just output the manifests as yaml or write code that installs those resources. The code can than be run locally or within the cluster.

With this approach, you have total flexibility when you provide an operator:
- you could just auto-generate the files in a pre-commit hook and add the yaml-files to your repository
- you could create a binary that installs the files and can be run once before deploying the operator
- you could auto-generate a very specific RBAC-file which just allows to install exactly these resources
- you could document how users can create the yaml files

... the quick win is that working code and documentation don't diverge.

While I like cargo plugins, this would make Rust a dependency for people who want to use the operator and I think Rust is not (yet) very common for many people. As an Operator-Provider, I would like people to be able to use the operator without the need to use something other than the normal Kubernetes-Tooling

kesselborn:
I am currently thinking, whether it would make sense to extract the derive macro into it's own little crate. It's not really related to krator but could be used independently as well.
The only remaining part would be the extraction of the certificate and the private key -- what do you think @kflansburg 

I'll think about it and probably adjust the pull request accordingly

crates/krator-derive/src/admission.rs
```diff
@@ -0,0 +1,273 @@
...
+                    _ => (),
+                };
+            }
+        }
+        cri.name = name.expect("kube macro must have property name set");
+
+        Ok(cri)
+    }
+
+    fn emit(self) -> Result<proc_macro2::TokenStream> {
```
kflansburg, line 106: There is a lot of stuff here that I think we may want to be able to configure via attributes.

crates/krator/examples/moose.rs
```diff
@@ -331,6 +362,70 @@ impl Operator for MooseTracker {
...
+        created_crd = api.replace(&crd_name, &PostParams::default(), &crd).await?;
+    } else {
+        created_crd = api.create(&PostParams::default(), &crd).await?;
+    }
+
+    Ok(created_crd)
+}
+
+#[cfg(feature = "admission-webhook")]
+async fn setup_cluster(client: &kube::Client) -> anyhow::Result<()> {
```
kflansburg, line 398: I think this would ideally be run automatically by `krator` when a given feature flag is enabled.

kflansburg (changes requested):
I think this looks good, albeit very complex. IMO the best practice for this would be to emit manifests / kustomize + Makefile in a similar fashion to `kubebuilder`, with auto-apply being an opt-in feature flag. I do think this gets us closer to that so I'm in favor of merging after some of my comments are addressed. 


@thomastaylor312 I think you should take a look at this since it touches a lot of stuff in `krator-derive`. I'm not sure how common it is to have so many dependencies in a derive crate but it seems like an antipattern.

crates/krator-derive/Cargo.toml
```diff
@@ -16,20 +16,37 @@ license-file = "../../LICENSE"
...
+[features]
+default = ["kube/native-tls"]
+admission-webhook = ["kube", "k8s-openapi", "rcgen", "anyhow"]
+
 [package.metadata.docs.rs]
 features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
+
+[dev-dependencies]
+kube = { version = "0.52", default-features = false, features = ["derive"] }
```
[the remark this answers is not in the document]
kesselborn: ah, probably not ... at first I only needed it in test and forgot to remove it

crates/krator/Cargo.toml
```diff
@@ -30,24 +26,32 @@ default = ["kube-native-tls"]
...
 [dependencies]
 async-trait = "0.1"
 anyhow = "1.0"
-tokio  = { version = "1.0", features = ["fs", "macros", "signal"] }
+tokio = { version = "1.0", features = ["fs", "macros", "signal"] }
 tokio-stream = { version = "0.1", features = ['sync'] }
 k8s-openapi = { version = "0.11", default-features = false, features = ["v1_20"] }
 kube = { version = "0.52", default-features = false }
-kube-runtime = { version= "0.52", default-features = false }
+kube-runtime = { version = "0.52", default-features = false }
```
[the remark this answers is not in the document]
kesselborn: oh, no: this was probably a toml plugin in vscode

crates/krator/examples/moose.rs
```diff
@@ -274,16 +303,18 @@ impl State<MooseState> for Released {
 
 struct SharedMooseState {
     friends: HashMap<String, HashSet<String>>,
+    client: kube::Client,
```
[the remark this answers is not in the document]
kesselborn: yeah, I am aware or that ... I thought it would be ok in an example ... could fix that though

crates/krator/examples/moose.rs
```diff
@@ -331,6 +362,70 @@ impl Operator for MooseTracker {
             }),
         }
     }
+
+    #[cfg(feature = "admission-webhook")]
+    async fn admission_hook_tls(&self) -> anyhow::Result<krator::admission::AdmissionTls> {
+        let client = self.shared.read().await.client.clone();
+        let secret_name = Moose::admission_webhook_secret_name();
+
+        let secret = kube::Api::<Secret>::namespaced(client, "default")
```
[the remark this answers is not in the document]
kesselborn: yup ... again: for the example I made some choices that I wouldn't do for a proper project and I thought about whether this is ok. But I think you are right: hard-coding default is probably not a good idea.

As to being more secure when mounting the secret: yes, you are right. Again, for the example here, we don't really know, whether the example is being executed locally or within the cluster -- reading the secret from the API works in both cases. Perhaps a comment explaining the consequence would be sufficient here?

crates/krator/examples/moose.rs
```diff
@@ -331,6 +362,70 @@ impl Operator for MooseTracker {
...
+
+async fn apply_crd(client: &kube::Client) -> anyhow::Result<CustomResourceDefinition> {
+    info!("installing moose crd");
+    let mut crd = Moose::crd();
+    let crd_name = crd.metadata.name.as_ref().unwrap();
+
+    let api = kube::Api::<CustomResourceDefinition>::all(client.to_owned());
+
+    let created_crd;
+    if let Ok(existing_crd) = api.get(crd_name).await {
```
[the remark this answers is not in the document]
kesselborn: oh, nice one ... thanks. I was looking for this but didn't find it.

crates/krator-derive/src/admission.rs
kesselborn: attributes to the feature? good idea! Can you switch on/off dependencies with this? The only real gain would probably less dependencies if, for example, you are not interested in the automatic generation of the certificate & private key.

crates/krator-derive/Cargo.toml
```diff
@@ -16,20 +16,37 @@ license-file = "../../LICENSE"
...
+kube-runtime = { version = "0.52", default-features = false }
+kube-derive = "0.52"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+serde_yaml = "0.8"
+schemars = "0.8.0"
+krator = { default-features = false, path = "../krator" }
```
thomastaylor312, line 52: This creates a circular dependency when trying to push the crate. We had this in here before (I think)

crates/krator-derive/src/admission.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl CustomDerive for CustomResourceInfos {
+    fn parse(input: DeriveInput) -> Result<Self> {
+        let ident = input.ident;
+
+        // Limit derive to structs
+        let _s = match input.data {
+            Data::Struct(ref s) => s,
+            _ => {
+                return Err(r#"Enums or Unions can not #[derive(AdmisstionWebhook)"#)
```
thomastaylor312, line 52: Nit (spelling): Admission

crates/krator-derive/src/admission.rs
```diff
@@ -0,0 +1,273 @@
...
+                /// Creates a MutatingWebhookConfiguration using the certificate from the given service and the service
+                /// of the given service as configuration
+                pub fn admission_webhook_configuration(service: k8s_openapi::api::core::v1::Service, secret: k8s_openapi::api::core::v1::Secret) -> anyhow::Result<k8s_openapi::api::admissionregistration::v1::MutatingWebhookConfiguration> {
+                   use anyhow::Context;
+                   let crd = #name_identifier::crd();
+
+                   let webhook_name = #name_identifier::admission_webhook_configuration_name() ;
+                   let versions: Vec<String> = crd.spec.versions.into_iter().map(|v| v.name).collect();
+
+                   anyhow::ensure!(secret.type_ == Some("tls".to_string()), format!("secret with name {} is not a tls secret", secret.metadata.name.unwrap()));
```
thomastaylor312, line 216: Are `unwrap`s the correct thing here? Seeing as these are errors, we should probably do an `unwrap_or_default`

kesselborn:
@kflansburg sorry for the late reploy and thanks for taking the time to review this rather big PR -- very much appreciated.

Before fixing the comments I would like to discuss again, if it makes sense to make this part of `krator_derive`. I implemented it there as I was using `krator` and was missing the functionality. When thinking about it again however, I think none of the additions are related to `krator` and could be used as well when writing an operator w/o using `krator` ... whereas pulling in `krator_derive` without using `krator` makes little sense.

I think extracting this as an individual crate is probably cleaner? What do you think?

Regarding pulling in a lot of additional dependencies via a proc macro: I was wondering this myself as well, and would be interested in @thomastaylor312 opinion here as well.

crates/krator-derive/src/admission.rs
```diff
@@ -0,0 +1,273 @@
...
+                            type_: Some("ClusterIP".to_string()),
+                            ..Default::default()
+                        }),
+                        status: None
+                   }
+                }
+
+                /// Creates a MutatingWebhookConfiguration using the certificate from the given service and the service
+                /// of the given service as configuration
+                pub fn admission_webhook_configuration(service: k8s_openapi::api::core::v1::Service, secret: k8s_openapi::api::core::v1::Secret) -> anyhow::Result<k8s_openapi::api::admissionregistration::v1::MutatingWebhookConfiguration> {
```
thomastaylor312, line 209: I think all of the error types should be `Box<dyn Error + Send + Sync>` for maximum capability. We do not want to force people deriving things to use `anyhow`. If we already are with krator, then this is fine for now

crates/krator-derive/src/lib.rs
```diff
@@ -1,96 +1,104 @@
...
+//! A crate for deriving state machine traits in Kubelet.
+//!
+//! Right now this crate consists of a [TransitionTo] derive macro for the `TransitionTo` trait.
+//!
+//! In addition to the `derive` attribute, this macro
 //! also requires the use of a custom attribute called `transition_to` that specifies the types that
 //! can be transitioned to. Not specifying this attribute will result in a compile time error.
-
+//!
+//! If the feature `admission-webhook` is enabled, this crate provides a [AdmissionWebhook] derive macro that
```
thomastaylor312, line 9: This should probably contain the list of dependencies the generated code uses, which I think are `kube`, `k8s_openapi`, and `anyhow`.

crates/krator-derive/Cargo.toml
```diff
@@ -16,20 +16,37 @@ license-file = "../../LICENSE"
...
+keywords = ["wasm", "wasi", "webassembly", "kubernetes"]
 
 [lib]
 proc-macro = true
-
 [dependencies]
 syn = "1.0"
 quote = "1.0"
 
+anyhow = { version = "1.0.40", optional = true }
```
thomastaylor312, line 27: After reading through the new code, I don't think we need most of these new dependencies. The code in the macro is dependent on the project using `krator-derive` to import the correct dependencies

crates/krator-derive/src/lib.rs
```diff
@@ -1,96 +1,104 @@
...
+/// - `admission_webhook_secret(namespace: &str) -> Secret`: Creates a Kubernetes secret of type `tls` that contains a
+///    certificate and a private key and can be used for the admission webhook service
+/// - `admission_webhook_service(namespace: &str) -> Service`: Returns a service that forwards to pods where label `app`
+///    has the value returned by the function `admission_webhook_service_app_selector()`. It exposes port `443`
+///    (necessary for webhooks) and listens to the pod's port `8443`
+/// - `admission_webhook_configuration(service: Service, secret: Secret) -> Result<MutatingWebhookConfiguration>`:  Creates a MutatingWebhookConfiguration using
+///    the certificate from the given service and the service of the given service as configuration
+///
+/// Example
+/// ```
```
thomastaylor312, line 46: This is a great example! The one problem (like in a previous comment) that this code cannot depend on `krator` or we get a weird circular dependency issue. If we can get it so that isn't needed in `dev-dependencies`, this will work fine. If it does need a dependency on `krator`, we'll want to move this example there

crates/krator/examples/moose.rs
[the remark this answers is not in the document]
thomastaylor312: Yeah, I think we can just pop an allow unused on there or conditionally include this parameter when the admission webhook feature isn't enabled

thomastaylor312 (changes requested):
This is looking pretty good. Thank you for adding tests and cleaning up the derive stuff.! Insofar as I understand, I don't think we need to import any `dependencies` for the derive macros to work, but I could be wrong

kflansburg:
@kesselborn Given that `krator-derive` should not depend on `krator`, does that address your concerns about splitting it out into a new crate? I think for now it makes most sense to me in `krator-derive`. I think this could still be used without `krator`.

crates/krator-derive/Cargo.toml
```diff
@@ -16,20 +16,37 @@ license-file = "../../LICENSE"
...
+k8s-openapi = { version = "0.11", default-features = false, features = ["v1_20"], optional = true }
```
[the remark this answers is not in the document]
kesselborn: [the default feature settings](https://github.com/deislabs/krustlet/blob/7c47872f33108b05d95c126c7820a2882dd7d68b/crates/krator-derive/Cargo.toml#L37) does this, doesn't it?

crates/krator-derive/Cargo.toml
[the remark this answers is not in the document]
kesselborn: fixed in commit acc48b7

crates/krator-derive/src/lib.rs
```diff
@@ -1,96 +1,104 @@
...
+//!
+//! Right now this crate consists of a [TransitionTo] derive macro for the `TransitionTo` trait.
+//!
+//! In addition to the `derive` attribute, this macro
 //! also requires the use of a custom attribute called `transition_to` that specifies the types that
 //! can be transitioned to. Not specifying this attribute will result in a compile time error.
-
+//!
+//! If the feature `admission-webhook` is enabled, this crate provides a [AdmissionWebhook] derive macro that
+//! provides functions for creating necessary resources for running a admission webhook.
```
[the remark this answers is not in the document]
kesselborn: these details are available if you click on the link [AdmissionWebhook]

image

I can move it to crate level as well -- not sure what is more idiomatic

crates/krator-derive/tests/admission.rs
```diff
@@ -0,0 +1,107 @@
...
+)]
+#[kube(group = "example.com", version = "v1", kind = "MyCr")]
+pub struct CrSpec {
+    pub name: String,
+}
+
+#[test]
+fn it_has_a_function_for_creating_admission_webhook_tls_secret() {
+    let secret: k8s_openapi::api::core::v1::Secret = MyCr::admission_webhook_secret("default");
+    // let secret = foo("default");
```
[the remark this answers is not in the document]
kesselborn: smuggled it into commit 9362f6f

crates/krator/examples/moose.rs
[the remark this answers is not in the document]
kesselborn: I changed the code in commit 9362f6f and this change is not necessary anymore

crates/krator/examples/moose.rs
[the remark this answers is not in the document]
kesselborn: made namespace configurable in commit 926d048

crates/krator/examples/moose.rs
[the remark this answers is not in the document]
kesselborn: fixed in commit 5afaea5

crates/krator/examples/moose.rs
kesselborn: I don't think I really agree here ... there are many different way of installing CRD and the other resources, especially due to different RBAC permissions. The function itself is kind of trivial when used like that and as it ships within the example, users can copy it if they like the approach.

crates/krator-derive/Cargo.toml
kesselborn: good catch ... although I think it's fine as it is only in `dev-dependencies`.
I removed it as it is not needed anymore (it was part of a test before)

removed in commit c3391b9

crates/krator-derive/src/admission.rs
kesselborn: fixed in commit 6e5fd8b

crates/krator-derive/src/admission.rs
kesselborn: 🤯 I got to admit, that I don't understand this as I am kind of new to Rust ... if you give me a pointer, I am happy to adjust the code accordingly

crates/krator-derive/src/admission.rs
kesselborn: fixed in commit 5d54163 ... I am catching errors early with `ensure!` to have more speaking error messages

kesselborn:
@kflansburg regarding the splitting out into a separate crate: the transition macro creates a 

```rust
impl krator::TransitionTo
```

block ... doesn't that make it kind of bound to krator?
However: happy to stand corrected, no too familiar what is most common

crates/krator-derive/Cargo.toml
kesselborn: yeah ... my hope was that the dependencies get pulled in as a transient dependency. 
Is there a good way to solve this? For example: there is a method creating a secret with a certificate and a private key using the `rcgen` crate. If this is not present, the user of `krator-derive` will get an error that `rcgen` is missing pointing to the derive macro ... is this something that users of macros usually have to solve themselves and is simply documented?

crates/krator-derive/src/lib.rs
kesselborn: that makes sense, yes

crates/krator-derive/src/admission.rs
kesselborn: I will have to look into how to parse those attributes

crates/krator-derive/Cargo.toml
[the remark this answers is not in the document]
kesselborn: ok: just saw the windows error: 

```
   Compiling num_cpus v1.13.0
error: failed to run custom build command for `openssl-sys v0.9.60`

Caused by:
  process didn't exit successfully: `D:\a\krustlet\krustlet\target\debug\build\openssl-sys-8ab17b7cd200101e\build-script-main` (exit code: 101)
  --- stdout
```

crates/krator-derive/Cargo.toml
kesselborn: I cleaned up the dependecies in commit dec3236

crates/krator-derive/src/admission.rs
thomastaylor312: Hahah no problem. So when you are trying to implement errors in the most generic way possible (meaning that you aren't going to require the use of a specific library), you can change return type to `Result<k8s_openapi::api::admissionregistration::v1::MutatingWebhookConfiguration, Box<dyn Error + Send + Sync>>`. Boxing a type that implements a trait in rust is how dynamic dispatch works but it is also essentially the "generic error type" or lowest common denominator in Rust. We also require the `Send + Sync` trait bounds as all of these are being used in an async context and will likely need to be sent across `await`s.

Hopefully that makes more sense? Honestly, if you don't feel comfortable doing it, we can handle it in a follow up too.

crates/krator-derive/Cargo.toml
thomastaylor312: Yeah it is generally documented and not pulled into a derive crate. Insofar as I understand things, derive macros are purely meant as transparent code generation

thomastaylor312:
To clarify, it doesn't depend on krator in the Cargo dependency sense. I think my comment [here](https://github.com/deislabs/krustlet/pull/577#discussion_r617690699) explains that a bit more

crates/krator-derive/src/admission.rs
kesselborn: this is implemented in commit 193b6ac

crates/krator-derive/src/lib.rs
kesselborn: done in commit 193b6ac

crates/krator-derive/src/admission.rs
kesselborn: @thomastaylor312 thanks for taking the time explaining this. That is kind of what I thought this is about. My remaining problem was on how to create this kind of error.
It took me a while to find out but finally a colleague poked me to `Err("foo".into())`  ... this should be mentioned in every Rust-error handling tutorial / texts / docs ;) 

I did this in commit e8a64e2

crates/krator-derive/src/lib.rs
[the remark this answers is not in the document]
kesselborn: I am resolving this for now, feel free to unresolve if you disagree.
The only thing that we have to make sure is that the feature `admission-webhook` is used when creating the docs -- can this be configured somehow?

crates/krator-derive/src/lib.rs
kesselborn: is it really a problem in `dev-dependencies` as well? I got problems when having circular dependencies in `dependencies` but no complains in `dev-dependencies` ... but that's perhaps just a cheap `works for me` ;)

crates/krator-derive/Cargo.toml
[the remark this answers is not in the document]
kesselborn: these dependencies are removed now and the Windows build works

kesselborn:
@kflansburg / @thomastaylor312 : can we move this forward somehow? I know, it's quite a lot of comments and therefore some work to review.

I resolved all comments apart from the ones, where I had a follow-up

kflansburg:
> @kflansburg / @thomastaylor312 : can we move this forward somehow? I know, it's quite a lot of comments and therefore some work to review.
> 
> 
> 
> I resolved all comments apart from the ones, where I had a follow-up

Apologies, I've been very busy this last week. I will try to take a final look at this soon.

bacongobbler:
Hey so I'm trying to run the examples, but I'm running into an issue with the admission webhook being unreachable:

```
><> k create -f manfred.yaml
Error from server (InternalError): error when creating "manfred.yaml": Internal error occurred: failed calling webhook "mooses.animals.com": Post "https://mooses-animals-com-admission-webhook.default.svc:443/?timeout=10s": dial tcp 10.96.19.121:443: connect: connection refused
```

I'm trying to follow the guide as directed in the logs:

```
May 05 09:36:15.807 moose: 
    
Running moose example. Try to install some of the manifests provided in examples/assets
```

This is with a KinD cluster, in case that matters. I'll try with AKS and see if my results differ.

bacongobbler:
Oddly I receive a different result with AKS:

```
><> k create -f manfred.yaml
Error from server (InternalError): error when creating "manfred.yaml": Internal error occurred: failed calling webhook "mooses.animals.com": Post "https://mooses-animals-com-admission-webhook.default.svc:443/?timeout=10s": read unix @->/socket/konnectivity: read: connection reset by peer
```

bacongobbler:
Things work just fine without the admission webhook feature enabled.

```
><> k create -f manfred.yaml
moose.animals.com/manfred created
><> k create -f melissa.yaml
moose.animals.com/melissa created
```

```
Running moose example. Try to install some of the manifests provided in examples/assets
    
    
    at crates/krator/examples/moose.rs:529

  May 05 10:36:29.473 krator::runtime: Got a watch restart. Resyncing queue...
    at crates/krator/src/runtime.rs:297

  May 05 10:36:29.473 krator::runtime: Finished resync of objects.
    at crates/krator/src/runtime.rs:300

  May 05 10:36:53.484 moose: Found new moose named manfred!
    at crates/krator/examples/moose.rs:127

  May 05 10:37:00.263 moose: Found new moose named melissa!
    at crates/krator/examples/moose.rs:127
```

kesselborn:
hey @bacongobbler : do you run the moose example outside the cluster or do you deploy it into the cluster? If you run it outside the cluster, you need to do some "hacking", so that the example can process the mutation requests ... if you look further up in the log it should have respective instructions:

```
If you run this example outside of Kubernetes (i.e. with `cargo run`), you need to make the webhook available.
Try the script example/assets/use-external-endpoint.sh to redirect webhook traffic to this process. If this
operator runs within Kubernetes and you use the webhook resources provided by the admission-webhook macro, 
make sure your deployment has the following labels set: ...
```

I tested locally using kind as well ... if you test it with AKS, you would have to make sure, that the Kubernets-Cluster can talk to your process

kesselborn:
so, just to give more context: these admission-webhooks are really just that: webhooks. So, in order to provide a webhook, the example app provides an https-endpoint where Kubernetes will post `moose` manifests whenever you create a new `moose`. This endpoint needs to be available through the service `mooses-animals-com-admission-webhook`. When you run the executable outside the cluster, you need to make the endpoint available to this service ... this is what the script ` script example/assets/use-external-endpoint.sh` will do for local setups like kind.

If you use AKS, you should deploy the moose example into the cluster setting the appropriate labels (they show up in the log as well), as this script will most probably not work.

bacongobbler:
Got it. Thanks for the context.

However, even with the script it still doesn't seem to be working. I'm not sure if the admission webhook is picking up on the new endpoint.

```
><> ./examples/assets/use-external-endpoint.sh
service/mooses-animals-com-admission-webhook patched
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
endpoints/mooses-animals-com-admission-webhook configured
><> k create -f examples/assets/manfred.yaml
Error from server (InternalError): error when creating "examples/assets/manfred.yaml": Internal error occurred: failed calling webhook "mooses.animals.com": Post "https://mooses-animals-com-admission-webhook.default.svc:443/?timeout=10s": dial tcp 10.96.78.189:443: connect: connection refused
```

Even if it did work, I don't think hard-coding to 172.17.0.1 will work for WSL2 as the network bridge from KinD to the host network is different for that operating system. We documented this process in the [HOWTO guide for Krustlet](https://github.com/deislabs/krustlet/blob/main/docs/howto/krustlet-on-kind.md#step-2-determine-the-default-gateway) if that helps.

For reference: the patch succeeded, and the endpoint exists.

```
><> k get svc mooses-animals-com-admission-webhook -o yaml | tail -n 12
spec:
  clusterIP: 10.96.78.189
  clusterIPs:
  - 10.96.78.189
  ports:
  - port: 443
    protocol: TCP
    targetPort: 8443
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
><> k get endpoints
NAME                                   ENDPOINTS         AGE
kubernetes                             172.18.0.2:6443   13m
mooses-animals-com-admission-webhook   172.17.0.1:8443   9m1s
```

bacongobbler:
As a note: I don't think my comments here are blocking this from going through. I'm just providing my (relatively naive) feedback here to inform you how someone new to the feature is experiementing. Doc/tooling fixes shouldn't block this from going through IMO.

That being said, the concerns raised earlier from @kflansburg about requiring escalated privileges to register an admission webhook seems concerning. If one were to find a security flaw in an operator's admission webhook, the attacker now would have escalated cluster privileges. That seems like a very concerning attack vector.

Are there ways to avoid the need for privilege escalation? How do other operator frameworks register admission webhooks?

kesselborn:
yeah ... testing this outside the cluster is a work around to speed up development and there will never be a really good solution for it. I just tried to reach parity with the [existing scripts in the current example code](https://github.com/deislabs/krustlet/blob/22486344c8a6a5924d8a8910f25bc1f171efc3bd/crates/krator/examples/assets/deploy.sh#L31-L39).

The correct way to test this is to deploy the moose operator into a Kubernetes cluster.

Regarding the security concerns: this is an example operator which is optimized for ease of use. You only need more privileges if you want to install necessary resources with the operator binary -- but that's not necessary. There are actually specific functions that will print out the manifests for you, so you can install the resources with `kubectl`, `helm`, etc.

Letting the operator produce these manifests is something I like as the operator and the documentation are not diverging -- you can always create the manifests that fit your operator.

kflansburg (changes requested):
@kesselborn We have discussed and believe that it is not a good idea to include this security vulnerability as part of the demo, as it will encourage bad behavior. If you can change it to generate the YAML for the user to manually apply, and update the README, then I will approve. If you really want to keep this functionality, then I would ask that this specific behavior is gated behind a specific feature with `danger` in the name. 

@bacongobbler Were you able to get the Endpoint working? Was the issue with the gateway being used and WSL?

@thomastaylor312 You has also requested changes. Can you indicate whether those have been resolved?

bacongobbler:
Thanks for the ping. In the end I was able to get it to work - just a quick misunderstanding of the networking setup on my end.

kesselborn:
@kflansburg I do agree that this sets a bad precedent. I adjusted the example accordingly and added a comment to to the apply functions to note, that they should not be called in the operator.

bacongobbler:
Providing some additional context as discussed offline...

We (Taylor and I) are maintainers on several large open source projects. Helm being one of them. In Helm 2, our stance was to provide a permissive default configuration. This allowed first-time users to start experimenting with Helm and Kubernetes without having to dive headfirst into the security controls. Unfortunately, this permissive configuration could grant a user a broad range of permissions they weren't intended to have. And for the most part users ignored the security documentation, instead choosing to criticize the project for "not doing the right thing" by default.

Learning from our mistakes, it's best to demonstrate "best practices" as the default rather than skewing towards ease-of-use. Especially when "ease-of-use" means skirting certain security controls.

Another great example of providing insecure defaults can be shown in [this article for MongoDB instances being hacked](https://www.zdnet.com/article/mongodb-ransacked-now-27000-databases-hit-in-mass-ransom-attacks/). The attackers hacked instances that did not have password-protected admin accounts. As in, they attacked MongoDB instances using the default settings.

As the Moose demonstration is the de-facto example for krator, I would not be surprised that many users will copy the example and apply it to their projects. Hence the concern.

crates/krator-derive/src/lib.rs
thomastaylor312: I think we are ok for now. I'll do some double checking later

kesselborn:
@bacongobbler thanks for the explanation -- I absolutely understand those points and do agree that this would have set a bad example :) 

Looking in the failed checks now

crates/krator/examples/moose.rs
kesselborn: resolving this as this would mean pulling the operations, that need a lot of permissions, into krator

kflansburg (approved):
LGTM, thanks for your patience!

PR #578, opened 2021-04-14
fix(bootstrap): Ensures directories are created before writing

thomastaylor312:
Closes #527

PR #582, opened 2021-04-18
Fix E2E Logs Flake

kflansburg:


thomastaylor312:
Got this working in #578! Thanks for doing the heavy lifting of debugging

PR #584, opened 2021-04-19
ref(volume): Adds new `Mountable` trait

### LINKED ISSUE #482: ref(pkg/volume): create a Mountable trait
Right now, `volumes_from_pod` takes a pod spec, parses the manifest to determine [which volumes need to be mounted](https://github.com/deislabs/krustlet/blob/461a8e9ff393c09dba53894a51bcdbd3393b3c20/crates/kubelet/src/volume/mod.rs#L42-L64) and mount them [by calling `.configure`](https://github.com/deislabs/krustlet/blob/461a8e9ff393c09dba53894a51bcdbd3393b3c20/crates/kubelet/src/volume/mod.rs#L47) for each volume type.

A better system would be to create separate structs for each type (PVC, Secret, ConfigMap, etc.) which implements a `Mountable` trait:

```
trait Mountable {
    fn mount(path: &PathBuf) -> Result<(), MountError>
}
```

thomastaylor312:
This adds a new `Mountable` trait that any volume can implement. Things
are better than before but not perfect and I think further refactor of
how we store references to volumes should be done before we hit 1.0.

Closes #482

### DIFF
--- a/crates/kubelet/src/volume/configmap.rs
+++ b/crates/kubelet/src/volume/configmap.rs
@@ -1,42 +1,77 @@
 use std::path::Path;
 
 use k8s_openapi::api::core::v1::{ConfigMap, KeyToPath};
+use k8s_openapi::ByteString;
 
 use super::*;
 
-pub(crate) async fn populate(
-    config_map: ConfigMap,
-    path: &Path,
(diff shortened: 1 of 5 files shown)


bacongobbler (approved):
so much better. LGTM

PR #588, opened 2021-04-21
chore(*): Updates the kubelet crate to use tracing

### LINKED ISSUE #583: Switch out traditional log lines for tracing in kubelet
Currently we just dropped in `tracing` to kubelet and continued using it like `log`. In preparation for the 1.0 release, we need to instrument major functions and use structured fields

thomastaylor312:
We had already dropped in the tracing crate, but were not using structured
fields. This adds structured fields to all log messages and switches from
the `env_logger` crate to the fmt subscriber.

Closes #583

### DIFF
--- a/crates/wasi-provider/src/states/pod/initializing.rs
+++ b/crates/wasi-provider/src/states/pod/initializing.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use tracing::{error, info};
+use tracing::{error, info, instrument};
 
 use kubelet::backoff::BackoffStrategy;
 use kubelet::container::state::run_to_completion;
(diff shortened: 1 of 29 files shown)


kflansburg:
@thomastaylor312 I believe this is running into the issue that we never solved from before where if we use `tracing` in the `krustlet-wasi` binary, it fails E2E tests. I wish we could figure it out, but probably not necessary in this PR.

crates/wasi-provider/src/states/pod/initializing.rs
```diff
@@ -30,16 +35,17 @@ impl State<PodState> for Initializing {
         let pod_rx = pod.clone();
         let pod = pod.latest();
 
+        tracing::Span::current().record("pod_name", &pod.name());
```
kflansburg, line 38: Cool

kflansburg (approved):
LGTM

PR #591, opened 2021-04-28
Added parsing of more parameters to the internal processing of a request for logs

### LINKED ISSUE #590: Extract and pass along all available parameters from the logs endpoint
The log tailing function in Kubelet currently only extracts the `--tail` and `--follow` parameters from the request served by the webserver.

Ideally all available parameters should be added to the Config struct as different providers will have different requirements here.

At a first glance the following ones sound useful:
- all-containers
- limit-bytes
- timestamps
- previous
- since
- since-time

There are more parameters that are supported by kubectl, but some of them are handled by kubectl/api-server  I think, not sure about all of them:
- prefix
- insecure-skip-tls-verify-backend
- pod-running-timeout
- max-log-requests
- selector

soenkeliebau:
Added some parameters to the internal processing of a request for logs:
- since
- since_seconds
- previous
- timestamps
- limit_bytes

More details on these can be found at https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#logs

fixes #590

### DIFF
--- a/crates/kubelet/src/log/mod.rs
+++ b/crates/kubelet/src/log/mod.rs
@@ -1,6 +1,8 @@
 //! `log` contains convenient wrappers around fetching logs from the Kubernetes API.
 use anyhow::bail;
+use chrono::{DateTime, Utc};
 use serde::Deserialize;
+use std::time::Duration;
 use tokio::io::{AsyncBufReadExt, AsyncRead};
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
LGTM! Thanks for the help with this one

PR #592, opened 2021-05-06
feat(provider): Adds traits for volume and plugin stuff

### LINKED ISSUE #524: Create trait for volume_path an plugin_registry.
Identified in #519 as an issue. 

We would like to add a trait bound on the shared provider state to require the methods for exposing `volume_path` and `plugin_registry`. This can provide default implementations, the goal is just to ensure there is only one place where this needs to be overridden to support CSI. 

cc @bacongobbler

thomastaylor312:
This still isn't super pretty (as you can't overload impls in Rust), but
at least things are consolidated. I can combine these traits if desired
but I left them separate in case we wanted to improve ergonomics further
in the future

Closes #524

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -4,7 +4,7 @@ use crate::config::Config;
 use crate::node;
 use crate::operator::PodOperator;
 use crate::plugin_watcher::PluginRegistry;
-use crate::provider::Provider;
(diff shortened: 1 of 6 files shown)


bacongobbler (approved):

technosophos (approved):

thomastaylor312:
Labeled as breaking so we don't forget to note the API change in the release notes

PR #593, opened 2021-05-10
oci-distribution: OCI error detail is optional

ereslibre:
According to the [OCI distribution
specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes),
both the `detail` and `message` fields are optional. Some
registries (such as `us.gcr.io`) are not reporting the `detail` field
back in light of certain errors. The `message` field could also be
omitted while still following the spec.

Implement serde defaulting using `Default::default()` if any of these
attributes are missing on the wire.

### DIFF
--- a/crates/oci-distribution/src/errors.rs
+++ b/crates/oci-distribution/src/errors.rs
@@ -8,9 +8,11 @@
 pub struct OciError {
     /// The error code
     pub code: OciErrorCode,
-    /// A message associated with the error
+    /// An optional message associated with the error
+    #[serde(default)]
     pub message: String,
-    /// Unstructured data associated with the error
+    /// Unstructured optional data associated with the error
+    #[serde(default)]
     pub detail: serde_json::Value,
 }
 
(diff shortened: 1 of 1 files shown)


crates/oci-distribution/src/errors.rs
```diff
@@ -8,9 +8,11 @@
...
     pub code: OciErrorCode,
-    /// A message associated with the error
+    /// An optional message associated with the error
+    #[serde(default)]
     pub message: String,
-    /// Unstructured data associated with the error
+    /// Unstructured optional data associated with the error
+    #[serde(default)]
     pub detail: serde_json::Value,
 }
```
ereslibre, line 17: The strategy of using serde defaulting to `Default::default()` is such that there's no cascading refactor on `krustlet`. However, if you are good with it, I can make `message` and `detail` be an `Option` that I believe is closer to the spec wording.

bacongobbler (approved):

PR #595, opened 2021-05-10
feat(tests): Adds e2e tests for CSI

### LINKED ISSUE #579: Add e2e tests for CSI
Right now we are not currently testing the CSI code in our e2e tests. This should be added before 1.0

thomastaylor312:
This PR adds e2e tests for the CSI code as part of our polishing for 1.0. Right now these tests only run on linux as some of the dependent components don't yet run on Mac. Also, the mock driver would need to use all of our ugly Windows socket code so we don't test it there either. Basically this is all a bit ugly, but it does exercise basic testing of the CSI components of the system. I also updated `podsmiter` to account for the new resources we are creating.

Closes #579

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,15 +57,19 @@ dirs = { package = "dirs-next", version = "2.0.0" }
 hostname = "0.3"
 regex = "1.3"
 tracing-subscriber = "0.2"
+serde = "1.0"
 
 [dev-dependencies]
 serde_derive = "1.0"
 serde_json = "1.0"
-serde = "1.0"
 reqwest = { version = "0.11", default-features = false }
-tempfile = "3.1"
 kube-runtime = { version = "0.55", default-features = false }
 compiletest_rs = "0.6"
+tempfile = "3.2"
+k8s-csi = "0.3"
+tonic = "0.4"
+async-trait = "0.1"
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "time", "process"] }
 
 [workspace]
 members = [
--- a/src/krustlet-wasi.rs
+++ b/src/krustlet-wasi.rs
@@ -15,6 +15,7 @@ async fn main() -> anyhow::Result<()> {
     // Initialize the logger
     tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
         .init();
 
     let kubeconfig = kubelet::bootstrap(&config, &config.bootstrap_file, notify_bootstrap).await?;
--- a/tests/csi/README.md
+++ b/tests/csi/README.md
@@ -1,0 +1,18 @@
+# CSI Test directory
+
+This directory contains the config for CSI RBAC (necessary if you are using
+KinD). Required binaries for the test are downloaded automatically when you use
+the `justfile`. Right now we only run these things on Linux because the
+Registrar is still missing [MacOS
+support](https://github.com/kubernetes-csi/node-driver-registrar/pull/133) and
+the mock driver would need to handle Windows, which would reuse our ugly code.
+It may be worth making some sort of more fully featured and separate mock driver
+in the future
+
+## Why not use the Docker images for these?
+
+Those would work here for our e2e tests since we are running a KinD node. However, we want these
+tests to reflect more real world usage and there is no guarantee that a Krustlet node will have a
+container runtime (nor do we want to require one). There is also the additional benefit of having
+this be some simple instructions of how you can build some of the key components to get CSI running
+for a Real World™ cluster.
--- a/tests/test_resource_manager.rs
+++ b/tests/test_resource_manager.rs
@@ -1,5 +1,7 @@
 use futures::StreamExt;
 use k8s_openapi::api::core::v1::{ConfigMap, Namespace, Pod, Secret};
+#[cfg(target_os = "linux")]
+use k8s_openapi::api::{core::v1::PersistentVolumeClaim, storage::v1::StorageClass};
 use kube::api::{Api, DeleteParams, PostParams};
 use serde_json::json;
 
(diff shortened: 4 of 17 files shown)


Cargo.toml
```diff
@@ -45,7 +45,7 @@ rustls-tls = [
 
 [dependencies]
 anyhow = "1.0"
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "time"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "time", "process"] }
```
thomastaylor312, line 48: The new features and dependencies added here are for things in the unit tests and associated binaries (like podsmiter)

src/krustlet-wasi.rs
```diff
@@ -15,6 +15,7 @@ async fn main() -> anyhow::Result<()> {
     // Initialize the logger
     tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
```
thomastaylor312, line 18: Yeah, we weren't using `RUST_LOG` properly. Whoops

Cargo.toml
kflansburg: Is this something that can be set in `dev-dependencies`?

tests/csi/README.md
```diff
@@ -0,0 +1,24 @@
+# CSI Binaries directory
```
kflansburg, line 1: I believe that these should either leverage [git LFS](https://git-lfs.github.com/), or be hosted somewhere else. I would prefer not to download 70 MB when cloning the repository. (Maybe it already is and I just cant tell from the PR).

tests/test_resource_manager.rs
```diff
@@ -180,16 +187,79 @@ impl TestResourceManager {
...
+                                "storageClassName": storage_class
+                            }
+                        }))?,
+                    )
+                    .await?;
+                self.push(TestResource::Pvc(resource_name.to_owned()));
+            }
         }
 
+        //         apiVersion: v1
```
kflansburg, line 239: Should this be cleaned up?

kflansburg (changes requested):
The code itself looks reasonable, this looks like it was a huge effort!

I have a few administrative requests.

Cargo.toml
thomastaylor312: Nope. There isn't a way to turn on the feature conditionally. I don't think you can have a duplicate dep in `dev-dependencies` can you?

tests/csi/README.md
thomastaylor312: I didn't do this as it looked like it was gonna be a hassle. Let me give it a try

thomastaylor312: Wait, wouldn't this mean we'd have to set up LFS on every build run and anyone who wanted to run the tests locally would have to have it as well? I feel like I'd prefer storing the binary somewhere and downloading it on demand

kflansburg: Hmmm, yes but this is pretty standard best practice, and easy to do. It’s pretty rude to have large binaries stored in git itself.

Cargo.toml
kflansburg: I'm pretty sure `kube` does this for `k8s-openapi` https://github.com/clux/kube-rs/blob/a6cc8995183a9ff9b0d7deb2ebad11dfa4ca31bb/kube/Cargo.toml#L86

tests/csi/README.md
thomastaylor312: Yes I agree. But I think that downloading is preferable to having git LFS for 2 files only used during tests. If these were key components, maybe. But I hesitate to add Yet Another Step™ to getting someone going with krustlet

tests/test_resource_manager.rs
thomastaylor312: Yeah, I'll take care of that shortly

tests/csi/README.md
kflansburg: Oh, I see, downloading is fine. I thought you were proposing keeping it in git.

kflansburg: FWIW, I'm pretty sure if its stored in LFS, you can still clone the repo, you would only need to set up LFS to obtain the binaries.

thomastaylor312: Basically I am debating between having an on-demand GH action that builds these binaries and pushes them into the blobstore or just manually pushing them into the blob store. I'll probably revisit on Monday once I get some more random opinions from the internet. But I don't think git LFS is the right choice here. Removing them from git definitely is

Cargo.toml
thomastaylor312: Fixed!

tests/csi/README.md
bacongobbler: It looks like [csi-release-tools](https://github.com/kubernetes-csi/csi-release-tools) only provides Docker images.

Have you considered pulling the binary out of the container image? That would alleviate the need to compile from source.

```
><> id=$(docker create k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.2.0)
><> docker cp $id:/csi-node-driver-registrar - | tar xf - > ./csi-node-driver-registrar
><> chmod 755 csi-node-driver-registrar
><> ./csi-node-driver-registrar --help
Usage of ./csi-node-driver-registrar:
  -add_dir_header
        If true, adds the file directory to the header of the log messages
```

thomastaylor312: That could work but adds a hard dependency on docker. I think pretty much everyone develops using `kind`, which means docker would be installed, but we haven't required that as of yet. I like keeping the set of tools small, but if this is easiest, we can do it this way

bacongobbler: If you're hard set on either github actions vs git-lfs, I would prefer setting up a GitHub Action that builds the binaries from source, then pushes them up as a Github Release. That way we can keep CSI artifacts and Krustlet artifacts separate. Then in Krustlet we'd download those artifacts during a CI run.

You could set up a separate repository for that... csi-release-action would seem like a good descriptive name.

thomastaylor312: Ok, think I am going to keep it simple and have an on demand job that uploads them to the krustlet blobstore, purely for testing purposes. If people find having the compiled binaries around useful, we can break out a separate project

thomastaylor312: https://github.com/deislabs/krustlet/pull/614

thomastaylor312: Ok this is addressed

thomastaylor312:
Ok @kflansburg and @bacongobbler. I added the new way of fetching the test binaries and squashed all the commits to remove the binaries from the branch

kflansburg (approved):
LGTM

PR #607, opened 2021-05-20
Fix typo: AdmissionTLs -> AdmissionTLS

kesselborn:
Cases don't make sense that way ;) ... oversaw that one until I used it in a project

### DIFF
--- a/crates/krator/examples/moose.rs
+++ b/crates/krator/examples/moose.rs
@@ -372,7 +372,7 @@ impl Operator for MooseTracker {
     }
 
     #[cfg(feature = "admission-webhook")]
-    async fn admission_hook_tls(&self) -> anyhow::Result<krator::admission::AdmissionTls> {
+    async fn admission_hook_tls(&self) -> anyhow::Result<krator::admission::AdmissionTLS> {
         let client = self.shared.read().await.client.clone();
         let secret_name = Moose::admission_webhook_secret_name();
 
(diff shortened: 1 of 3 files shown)


kflansburg:
While I agree that it’s not correct English, the partially lowercase acronym is idiomatic Rust. If you run clippy on your branch, it should complain about the all uppercase acronym.

kesselborn:
But then it should at least bei AdmissionTls, nicht AdmissionTLs

Making TLS all capital is probably something I got from Go.
Either way is fine, but not TLs

bacongobbler:
@kesselborn im not sure where you see it as AdmissionTLs.If you look at the code more closely the L is lowercases to follow Rust’s convention of AdmissionTls. Please re-review more closely :)

From https://doc.rust-lang.org/1.0.0/style/style/naming/README.html:

> In CamelCase, acronyms count as one word: use Uuid rather than UUID. In snake_case, acronyms are lower-cased: is_xid_start.

This follows Rust’s convention, so the current casing is correct. Because of that I’m going ahead and closing this as the current styling confirms to standard Rust style convention.

kesselborn:
Uhm 🙈,
super sorry: I should sleep more 😀

I was 100% convinced that it was wrong but looking a the change proves me kind of wrong 😃

Sorry for the noise, don’t know what was wrong with me this morning.

Cheers,
Daniel

Issue #463, opened 2020-12-01
ConfigMap/Secret volumes should be mounted as read-only


bacongobbler:
As I'm reading through the `volume` module, it appears as though we do not prepare the ConfigMap and Secret host mounts as read-only. We just write the data to a filepath in the volume and provide that directory to the Provider.

https://github.com/deislabs/krustlet/blob/0b99368e91934d489ade88be99188ec632509475/crates/kubelet/src/volume/mod.rs#L41

https://github.com/deislabs/krustlet/blob/0b99368e91934d489ade88be99188ec632509475/crates/kubelet/src/volume/mod.rs#L163

In Kubernetes, those volumes are mounted as read-only.

thomastaylor312:
@bacongobbler After looking at this some more, I think this is a per provider implementation detail (as we might not be actually "mounting" something depending on what it is). I don't think this is currently possible from the WASI side of things. However, we could set the file permission modes to readonly if you think that might help as well. But at least from the WASI side, if they have full access to the directory, they should be able to just modify permissions back

bacongobbler:
> However, we could set the file permission modes to readonly if you think that might help as well.

I believe this is how Kubernetes handles read-only permissions, though I believe they would use cgroups and vfs mounts to set up read/write permissions.

https://docs.oracle.com/en/operating-systems/oracle-linux/6/adminsg/ol_cgconfig_cgroups.html

Here's where the kubelet sets up standard volume mounts:

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/kubelet_pods.go#L257-L266

And here's where the kubelet sets up raw block volumes:

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/kubelet_pods.go#L129-L135

thomastaylor312:
Generally the providers don't mount the volume themselves, the CSI plugin does when it is called. But in the configmap/secrets case, we aren't mounting anything. We are just preopening the file from WASI. A container provider will be loading the local file system path as a volume into the container, but there is no such thing for WASI, so this is a little unclear. Perhaps we should sync up on this at the next meeting

PR #615, opened 2021-05-27
fix(#463): read-only configmap/secret mount directory

fibonacci1729:


bacongobbler (approved):
LGTM! This oughta do it.

PR #617, opened 2021-06-01
oci-distribution: fix push to ghcr.io

flavio:
The oci-distribution crate is currently pushing layers using the "monolithic upload" API offered by distribution v2[1].

According to the official specification, the layer has to be uploaded using this API:

```
PUT /v2/<name>/blobs/uploads/<uuid>?digest=<digest>
```

Prior to this commit, the code was building the destination url using simple string interpolation.
That resulted in making a request like:

```
PUT /v2/<name>/blobs/uploads/<uuid>&digest=<digest>
```

Interestingly enough, this didn't seem to be a problem to the distribution servers. Apparently the Go implementation is more
relaxed when it comes to parsing an incoming request.

The GitHub Package registry (ghcr.io), on the other hand rejects these requests with a 404 response.

This commits fixes the problem by using the `Reqwest::Url` structure to create the target url.

[1] https://github.com/distribution/distribution/blob/v2.7.1/docs/spec/api.md#monolithic-upload

### DIFF
--- a/crates/oci-distribution/src/client.rs
+++ b/crates/oci-distribution/src/client.rs
@@ -16,7 +16,7 @@ use anyhow::Context;
 use futures_util::future;
 use futures_util::stream::StreamExt;
 use hyperx::header::Header;
-use reqwest::header::HeaderMap;
+use reqwest::{header::HeaderMap, Url};
 use serde::Deserialize;
 use sha2::Digest;
 use std::collections::HashMap;
@@ -573,11 +573,16 @@ impl Client {
         image: &Reference,
         digest: &str,
     ) -> anyhow::Result<String> {
-        let url = format!("{}&digest={}", location, digest);
+        let url = Url::parse_with_params(location, &[("digest", digest)])?;
         let mut close_headers = self.auth_headers(image);
         close_headers.insert("Content-Length", "0".parse().unwrap());
 
-        let res = self.client.put(&url).headers(close_headers).send().await?;
+        let res = self
+            .client
+            .put(url.as_str())
+            .headers(close_headers)
+            .send()
+            .await?;
         self.extract_location_header(&image, res, &reqwest::StatusCode::CREATED)
             .await
     }


bacongobbler:
Thanks @flavio. What's the best way to test this PR? Is there a way we can add a unit test here to catch future regressions?

flavio:
Sorry, yesterday was a bank holiday in Italy. I'll think about how to get the fix covered by unit tests :+1:

flavio:
> Thanks @flavio. What's the best way to test this PR? Is there a way we can add a unit test here to catch future regressions?

I think the most reliable way would be to add a new test that attempts to push something to ghcr.
[Recently GitHub made some changes](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry) that allow push/pull operations to be done from a GitHub Action using the `GITHUB_TOKEN` that is added automatically each run.

The summarize:

  1. Enable ghcr for the krustlet project
  2. Write a unit test that pushes a simple Wasm file to ghcr:
     2.1. By default use something like `ghcr.io/deislabs/krustlet/ci/oci-distribution-test-suite:latest` as target
     2.2. Allow this target to be changed (dunno, maybe via an environment variable?)
  3. Mark this unit test as `#[ignore]`, so that on local machines is not executed
  4. Ensure the test is run inside of the GitHub Actions

Point 2.2 allows regular developers, who do not have the privileges to push contents to `ghcr.io/deislabs/krustlet`, to execute the test against a ghcr repository they own.

What do you think about this proposal?

bacongobbler (approved):
LGTM. Thanks!

bacongobbler:
Oops... Would you mind rebasing this?

ereslibre:
Note: https://github.com/krustlet/krustlet/pull/670 contains the commit on this PR and is rebased ;)

flavio:
Let's close this in favor of #670

PR #620, opened 2021-06-02
Remove Krator source and use crates.io version

kflansburg:


thomastaylor312 (approved):
Look at Krator, all grown up and going out on its own 😢

Issue #589, opened 2021-04-23
Check that parameters for certificate and private key are not set to the same file


soenkeliebau:
When --cert-file and --private-key-file are set to the same file one overwrites the other and TLS fails with a parse error (which makes sense).

A quick check in the config that these parameters do not have the same value would be a useful addition I think.

sladyn98:
@thomastaylor312 Hey i am just getting started with rust and thought this was a good way to get started, any pointers on where I can begin ?

soenkeliebau:
@sladyn98 
I haven't spent much time looking, but I think this should be a good starting point: https://github.com/deislabs/krustlet/blob/c781a9816858607647108d2157c7c5c1ca004a35/crates/kubelet/src/config.rs#L338

That code should be common to most ways of building a config and can return an error during initialization, which should be suitable for this check.

thomastaylor312:
Hi @sladyn98, thanks for offering to help! So when I was double checking things, I actually think we'd want to verify this in the `Kubelet::new` function: https://github.com/deislabs/krustlet/blob/20c00115d3af460fe7740d6a07589b25578fcaef/crates/kubelet/src/kubelet.rs#L42

The reason being is that the config file and command line flag stuff is optional, which means a user can generate a `Config` on their own and have it not be valid. I think at minimum, it should check the two pieces of data specified in this issue. A stretch goal (but definitely not required for a first time issue like this) would be to add a `validate` function to the `Config` so any future data validation can be added there.

Hope that helps! And thanks again for the help

yunkunrao:
Hi, I found this issue was not updated since June 5th and I'm been following Krustlet community. So I submit a new PR, could you please help review this PR when it is convenient. Thanks.

PR #623, opened 2021-06-04
Validate config properties

sladyn98:
Added a check to validate if the config's private key is equal to cert file

Closes: https://github.com/deislabs/krustlet/issues/589

### DIFF
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -46,7 +46,7 @@ impl<P: Provider> Kubelet<P> {
     ) -> anyhow::Result<Self> {
 
         if config.server_config.private_key_file == config.server_config.cert_file {
-            // throw error ?
+            panic!("The config private key file and server config certificate file are the same");
         }
 
         Ok(Self {
--- a/crates/kubelet/src/kubelet.rs
+++ b/crates/kubelet/src/kubelet.rs
@@ -44,6 +44,11 @@ impl<P: Provider> Kubelet<P> {
         kube_config: kube::Config,
         config: Config,
     ) -> anyhow::Result<Self> {
+
+        if config.server_config.private_key_file == config.server_config.cert_file {
+            // throw error ?
+        }
+
         Ok(Self {
             provider: Arc::new(provider),
             kube_config,


crates/kubelet/src/kubelet.rs
```diff
@@ -44,6 +44,11 @@ impl<P: Provider> Kubelet<P> {
         kube_config: kube::Config,
         config: Config,
     ) -> anyhow::Result<Self> {
+
+        if config.server_config.private_key_file == config.server_config.cert_file {
+            // throw error ?
```
sladyn98, line 49: @thomastaylor312  Would we like to throw an error here ?

bacongobbler:
Hi @sladyn98. Can you please provide a unit test to cover this issue? An explanation discussing what issue you are experiencing and why this PR fixes it would also be helpful to others reviewing this code. Thanks!

sladyn98:
@bacongobbler hey, I attached the issue number

bacongobbler:
Oh! Thank you.

Yes, in this case it would be proper to throw an error.

sladyn98:
@bacongobbler Does this look okay ?

crates/kubelet/src/kubelet.rs
```diff
@@ -44,6 +44,11 @@ impl<P: Provider> Kubelet<P> {
         kube_config: kube::Config,
         config: Config,
     ) -> anyhow::Result<Self> {
+
+        if config.server_config.private_key_file == config.server_config.cert_file {
+            panic!("The config private key file and server config certificate file are the same");
```
bacongobbler, line 49: `panic` probably isn't the right choice here... That will cause the program to exit abruptly. `panic` should only be used when a program reaches an unrecoverable state.

It might be better to instead use the `anyhow::Error` return type. Something like this...

```
return Err(anyhow!("The config private key file and server config certificate file are the same"));
```

bacongobbler (changes requested):

bacongobbler:
Closing in favour of https://github.com/krustlet/krustlet/pull/683 which more closely matches with the expected behaviour (and includes tests to verify). Thank you!

PR #627, opened 2021-06-15
to fix: #625 - Windows cannot cleanup volumes on unmount 

### LINKED ISSUE #625: Windows cannot cleanup volumes on unmount
It looks like secret and configmap volumes are having issues cleaning up on Windows. This could be due to an open file handle we aren't cleaning up or we may need to use a different cleanup command for Windows. This problem manifests itself in the logs:

```
wasi_provider::states::pod: Unable to unmount volume error=Access is denied. (os error 5) volume_name=default-token-pxmd7
```

VishnuJin:
@thomastaylor312 
i have added the steps to resolve #625 unmount issue in windows using [remove_dir_all](https://crates.io/crates/remove_dir_all) crate only for windows and unix family will continue to have std::fs implementation as usual.

I'm unable to test this completely as i don't have Windows pc, I'm on M1 mac and elapsed parallels trial already :(
Compilation and just windows test has been passed.
Please let me know you thoughts !!

### DIFF
--- a/crates/kubelet/Cargo.toml
+++ b/crates/kubelet/Cargo.toml
@@ -45,7 +45,7 @@ serde_json = "1.0"
 serde_yaml = "0.8"
 hyper = { version = "0.14", default-features = false, features = ["stream"] }
 reqwest = { version = "0.11", default-features = false, features = ["json", "stream"]}
-tokio  = { version = "1.0", features = ["fs", "macros", "signal", "net","rt"] }
+tokio  = { version = "1.0", features = ["fs", "macros", "signal", "net"] }
 tokio-stream = { version="0.1", features = ["fs", "net"] }
 kube = { version = "0.55", default-features = false, features = ["jsonpatch"] }
 kube-runtime = { version= "0.55", default-features = false }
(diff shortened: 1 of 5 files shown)


crates/kubelet/Cargo.toml
```diff
@@ -45,7 +45,7 @@ serde_json = "1.0"
 serde_yaml = "0.8"
 hyper = { version = "0.14", default-features = false, features = ["stream"] }
 reqwest = { version = "0.11", default-features = false, features = ["json", "stream"]}
-tokio  = { version = "1.0", features = ["fs", "macros", "signal", "net"] }
+tokio  = { version = "1.0", features = ["fs", "macros", "signal", "net","rt"] }
```
thomastaylor312, line 48: Is this `rt` required? We use spawn_blocking in the `wasi-provider` and it doesn't need the `rt` feature

thomastaylor312 (commented):
Thank you for working on this! This is looking pretty good. One minor comment. I'll come back to re-review once tests pass

crates/kubelet/Cargo.toml
VishnuJin: not necessary!! just saw in the tokio documentation and added this, i have removed this in the next commit

thomastaylor312 (approved):

thomastaylor312:
Thanks again @VishnuJin!

VishnuJin:
@thomastaylor312 I'm very happy, thank you ☺️

Issue #540, opened 2021-03-08
Use the new async support in wasmtime


thomastaylor312:
Wasmtime 0.24 added support for [async functions](https://github.com/bytecodealliance/rfcs/blob/main/accepted/async-imports.md). This should allow us to move away from using `spawn_blocking`.

thomastaylor312:
Based on the documentation [here](https://docs.rs/wasmtime/0.26.0/wasmtime/struct.Config.html#asynchronous-wasm), I think it would be best if we maintain our current workflow as it is one of the suggested ways we do it. `Store` is still not Send, which will likely cause problems returning it from the thread. So I am popping this off of 1.0 and deferring until we get to further maturity

peterhuene:
We've also added additional [guidance regarding Wasmtime and multi-threading safety in Rust](https://github.com/bytecodealliance/wasmtime/blob/main/docs/examples-rust-multithreading.md).

I've also got an [example wrapping a `Future`](https://github.com/peterhuene/wasmtime-functions/blob/main/crates/runtime/src/server.rs#L103-L133) that solely contains the `Store` and all related objects to make it "safely" marked as `Send` for use on a multi-threaded executor.

thomastaylor312:
That is great! Thanks @peterhuene

PR #630, opened 2021-06-19
Update Wasmtime to v0.28

radu-matei:
This commit updates Wasmtime to v0.28 and simplifies the way modules are instantiated using the linker.

### DIFF
--- a/crates/wasi-provider/Cargo.toml
+++ b/crates/wasi-provider/Cargo.toml
@@ -24,10 +24,10 @@ anyhow = "1.0"
 async-trait = "0.1"
 backtrace = "0.3"
 kube = { version = "0.55", default-features = false }
-wasmtime = "0.25"
-wasmtime-wasi = "0.25"
-wasi-common = "0.25"
-wasi-cap-std-sync = "0.25"
+wasmtime = { version = "0.28", default-features = true }
+wasmtime-wasi = "0.28"
+wasi-common = "0.28"
+wasi-cap-std-sync = "0.28"
 cap-std = "0.13"
 tempfile = "3.1"
 serde = "1.0"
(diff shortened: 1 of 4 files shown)


crates/wasi-provider/Cargo.toml
```diff
@@ -24,18 +24,18 @@ anyhow = "1.0"
 async-trait = "0.1"
 backtrace = "0.3"
 kube = { version = "0.55", default-features = false }
-wasmtime = "0.25"
-wasmtime-wasi = "0.25"
-wasi-common = "0.25"
-wasi-cap-std-sync = "0.25"
+wasmtime = { version = "0.28", default-features = true }
```
thomastaylor312, line 27: Super nit: Isn't default features...well...the default?

crates/wasi-provider/Cargo.toml
```diff
@@ -24,18 +24,18 @@ anyhow = "1.0"
...
+wasi-cap-std-sync = "0.28"
 cap-std = "0.13"
 tempfile = "3.1"
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 kubelet = { path = "../kubelet", version = "0.7", default-features = false, features = ["derive"] }
 krator = { version = "0.3", default-features = false, features = ["derive"] }
-wat = "1.0"
+wat = "1.0.38"
```
thomastaylor312, line 38: Is this pin necessary or can we use anything from the 1.0 series? I know it will select higher, but I was just curious

thomastaylor312 (commented):
Code is looking good. However, is it possible to move all of this wasmtime stuff out of `spawn_blocking` now that it is `Send + Sync`? This would close #540 and is part of the motivation for having this in wasmtime

crates/wasi-provider/Cargo.toml
radu-matei: It is, Wasmtime fails to compile with anything lower than 1.36 I think?

thomastaylor312 (approved):

PR #633, opened 2021-06-23
Updated default aks version to 1.20.7

cmhamakawa:
1.18.10 is deprecated, so the demo fails by default right now.


bacongobbler (approved):
Thank you for the version bump!

PR #641, opened 2021-07-14
Add github actions to run cargo-audit on pull requests and daily

### LINKED ISSUE #640: Rust security audit should be run regularly to catch outdated or orphaned dependencies
Krustlet currently does not appear to regularly run `cargo audit` to check for [security advisories](https://rustsec.org/) against its dependencies. 

We currently have an issue against our agent which is caused by a dependency that Krustlet pulls in
https://github.com/stackabletech/agent/issues/216 which was caused by our [daily audit check](https://github.com/stackabletech/agent/blob/main/.github/workflows/daily_security.yml).

It might be worth adding this or something similar to Krustlet as well.

soenkeliebau:
Adds workflows to automatically run cargo-audit [1] to check for security advisories against dependencies of this project.

This is set to run on two occasions:
- Every day at midnight - to catch new advisories for existing dependencies
- Upon a new pull request or push to the main branch - to ensure no new dependencies are added which have existing advisories against them

Additionally we could consider making this a required check for pull requests.

fixes #640 

[1] https://github.com/RustSec/rustsec


.github/workflows/build.yml
```diff
@@ -169,3 +169,27 @@ jobs:
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  cargo-deny:
+    name: Run cargo deny
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
```
thomastaylor312, line 189: Because this project is headed to the CNCF, they have their own license scanning process. So I think we should drop this from here for now, but keep the other ones

thomastaylor312 (changes requested):
This is looking great! Just one change requested here. Also, if we merge this, I assume we will follow up on the advisories in another PR, correct? If so, we should probably create an issue and add it to the 1.0 milestone so we make sure to address it.

.github/workflows/build.yml
radu-matei: I am not familiar with the Snyk license scanner, is it as thorough as a Cargo specific one? Does it even have support for Rust?

My experience with FOSSA (the alternative CNCF license scanner) is not that great, so I wouldn't dismiss this one built specifically for Rust.

thomastaylor312: It might not be. I'm fine keeping this around.

thomastaylor312 (approved):
@soenkeliebau This is good to go! But before I merge, would you mind opening some issues to handle any of the advisory checks that aren't passing? I can fix them soon, but I wan to make sure we are tracking them

thomastaylor312:
Oh, you'll also need to signoff (`git commit --amend --no-edit -s`) your latest commit now that this repo is under the CNCF DCO

.github/workflows/build.yml
soenkeliebau: I completely agree on the FOSSA experience, we gave that a try as well. We had a call with them about our experience and they were quite open about Rust not being a priority for them because they see next to no demand for it. Which is fair enough, I'm not complaining, just giving context :)

soenkeliebau:
> @soenkeliebau This is good to go! But before I merge, would you mind opening some issues to handle any of the advisory checks that aren't passing? I can fix them soon, but I wan to make sure we are tracking them

In principle I'm happy to, but the audit action should actually create issues itself once it is run not from a fork and allowed to do so (those should be the same advisories that the deny action finds). That would most probably duplicate anything I open.. 

Regarding the license failures, so far in the `deny.toml` file I've simply copied what we have decided on for our project, not sure if that is also suitable for you / the CNCF.  Might be worthwhile you having a quick look in there.

I've had a quick look at what the license check currently complains about: 

```
error[L003]: wasi-provider = 0.7.0 is unlicensed
  ┌─ wasi-provider 0.7.0 (path+file:///github/workspace/crates/wasi-provider):2:9
  │
  │ name = "wasi-provider"
  │         ^^^^^^^^^^^^^ a valid license expression could not be retrieved for the crate
  │ version = "0.7.0"
  │ license = ""
  │            - license expression was not specified
```
I'll open a ticket for this.

```
error[L001]: failed to satisfy license requirements
  ┌─ mach 0.3.2 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
  │ license = "BSD-2-Clause"
  │            ^^^^^^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: not explicitly allowed
```
Should we add this to the list of allowed licenses?



```
error[L003]: krator-derive = 0.2.0 is unlicensed
  ┌─ krator-derive 0.2.0 (registry+https://github.com/rust-lang/crates.io-index):2:9
  │
  │ name = "krator-derive"
  │         ^^^^^^^^^^^^^ a valid license expression could not be retrieved for the crate
  │ version = "0.2.0"
  │ license = ""
  │            - license expression was not specified
  │ license-files = [
  │     { path = "LICENSE", hash = 0x62365415, score = 1.00, license = "Apache-2.0" },
  │                                                    ---- low confidence in the license text
```
I'll open a ticket for this.

```
  ┌─ webpki-roots 0.21.0 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
  │ license = "MPL-2.0"
  │            ^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: license is considered copyleft
```
I've had a look at the MPL faq on their page and think this paragraph would be applicable:

> Q8: I want to distribute (outside my organization) executable programs or libraries that I have compiled from someone else's unchanged MPL-licensed source code, either standalone or part of a larger work. What do I have to do?
>
> You must inform the recipients where they can get the source for the MPLed code in the executable program or library you are distributing (i.e., you must comply with Section 3.2). You may distribute any executables you create under a license of your choosing, as long as that license does not interfere with the recipients' rights to the source under the terms of the MPL.

Sounds to me like this should be included in something like a NOTICE file and we should be good. Should I open a ticket for this?

thomastaylor312:
Oh I forgot that the action does that for you!

> Sounds to me like this should be included in something like a NOTICE file and we should be good. Should I open a ticket for this?

Yeah, let's go ahead and open a ticket for it.

PR #646, opened 2021-07-19
Feat/k8s1.21 support

### LINKED ISSUE #618: Fix volume handling for 1.21
Looks like k8s 1.21 added a new `projected` volume type that amalgamates all the functionality of the downward API stuff as well as secret/configmap volumes(see https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/). However, our current volume logic doesn't know how to handle it. It looks like the default service account token and certificate are automatically generated using this `projected` volume type like so:

```yaml
- name: kube-api-access-b59rx
      projected:
        defaultMode: 420
        sources:
        - serviceAccountToken:
            expirationSeconds: 3607
            path: token
        - configMap:
            items:
            - key: ca.crt
              path: ca.crt
            name: kube-root-ca.crt
        - downwardAPI:
            items:
            - fieldRef:
                apiVersion: v1
                fieldPath: metadata.namespace
```

thomastaylor312:
This does all the work for making Krustlet compatible with k8s 1.21. Unfortunately, this involved a lot more work than intended. You can look at each commit message for a bit more detail, but this PR mainly adds/fixes these things:

- Support for the new projected volume type
- Downward API volumes (added to support the projected volumes)
- Fixes container status patching
- Adds tests for the new volume stuff

Closes #618


Cargo.toml
```diff
@@ -46,10 +44,10 @@ rustls-tls = [
 [dependencies]
 anyhow = "1.0"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "time"] }
-kube = { version = "0.55", default-features = false }
-k8s-openapi = { version = "0.11", default-features = false, features = ["v1_20"] }
+kube = { version = "0.58", default-features = false }
+k8s-openapi = { version = "0.12", default-features = false, features = ["v1_21"] }
 futures = "0.3"
-krator = { version = "0.3", default-features = false }
+krator = { git = "https://github.com/krator-rs/krator", version = "0.3", default-features = false, branch = "main" }
```
thomastaylor312, line 50: Please note that I am working on cutting a new version of krator with the API changes. I will come back and update this, but it should not be merged until I do

thomastaylor312: See: https://github.com/krator-rs/krator/pull/45

crates/kubelet/src/resources/quantity.rs
```diff
@@ -0,0 +1,338 @@
+//! A custom Quantity type that actually parses the `Quantity` type from k8s to something usable
+//! with real numbers
+
+// NOTE: This whole module might be useful to other people doing k8s things in Rust. We may want to
+// eventually export this or create a separate crate if people want to reuse it
+
+use k8s_openapi::apimachinery::pkg::api::resource::Quantity as KubeQuantity;
+
+const KILOBYTE: f64 = 1000.0;
```
bacongobbler, line 9: How come they need to be floating point values? Shouldn't these be of type `u32`?

I can understand the need for floating point values for `QuantityType::Cpu`. but shouldn't we use whole numbers for `QuantityType:Memory`?

crates/kubelet/src/volume/projected.rs
```diff
@@ -0,0 +1,250 @@
...
+    client: kube::Client,
+    audience: String,
+    expiration_time: i64,
+    pod_name: String,
+    pod_uid: String,
+}
+
+impl ServiceAccountSource {
+    async fn mount_at(&mut self, path: impl AsRef<Path>) -> anyhow::Result<()> {
+        // As far as I can tell, this is the only way to access the token subresource on service accounts
```
bacongobbler, line 35: Yeah, this comment from kubeflow seems to support that statement (short of storing this information in another service account): https://github.com/kubeflow/pipelines/issues/5138#issuecomment-845846555

crates/kubelet/src/volume/projected.rs
```diff
@@ -0,0 +1,250 @@
...
+        let (req, _) = TokenRequest::create_namespaced_service_account_token(
```
bacongobbler, line 36: Should the second return parameter be handled? It appears to return type StatusCode... I'm wondering if we need to handle that as well as the Result (which we implicitly handle on line 53).

https://docs.rs/k8s-openapi/0.12.0/k8s_openapi/api/authentication/v1/struct.TokenRequest.html

bacongobbler (approved):
Great work. Left a few comments just to clarify a few things, but the code here looks good to me.

crates/kubelet/src/resources/quantity.rs
thomastaylor312: Welcome to the terrible world of `Quantity`! So according to the spec, you can technically make a memory request of `1.5Gi` which is a float. Basically we convert everything to a float to make the operations easier, hence why these values are floats as well

crates/kubelet/src/volume/projected.rs
thomastaylor312: Nope, it took me a sec to grok this as well. It returns a function that can create an error type from a status code. All I need here is the raw request I can pass into the `kube` library, which does all of its own error conversion stuff for me

crates/kubelet/src/resources/quantity.rs
bacongobbler: Wonderful.

PR #647, opened 2021-07-22
Restrict permissions of private key file

### LINKED ISSUE #639: Private key file is created with default umask
When using bootstrapping for krustlet, the private key file (and certificate, but not really critical for this) is created without specifying a umask [1].

On most systems this will default to being world readable, which is not ideal for something this critical.

We should look into making this readable for the user running krustlet only.

[1] https://github.com/deislabs/krustlet/blob/e802c9c8b176fc9300d411386e16cc56869887ad/crates/kubelet/src/bootstrapping/mod.rs#L274

siegfriedweber:
On Unix file systems the mode of the private key file is set to 0o600. On other file systems the permissions remain unchanged.

Fixes #639

### DIFF
--- a/crates/kubelet/src/bootstrapping/mod.rs
+++ b/crates/kubelet/src/bootstrapping/mod.rs
@@ -1,4 +1,4 @@
-use std::{convert::TryFrom, env, path::Path, str};
+use std::{convert::TryFrom, env, io, path::Path, str};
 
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):

thomastaylor312:
Thank you for the fix!

PR #648, opened 2021-07-22
Add experimental WASI HTTP support

fibonacci1729:
This PR adds:

- Support for the experimental HTTP library in the wasi provider.
- A simple demo to showcase the enhancement.


demos/wasi/postman-echo/k8s.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: postman-echo
+  labels:
+    app: postman-echo
+  annotations:
+    alpha.wasi.krustlet.dev/allowed-domains: '["https://postman-echo.com"]'
```
radu-matei, line 8: Is this a normal YAML array?

fibonacci1729: What do you mean "normal"?

radu-matei: Looking at the example, I am not sure how to format the YAML value for multiple allowed domains.

Looking at the YAML spec for sequences / collections (https://yaml.org/spec/1.2/spec.html#id2759963), I thought this would be similar to:

```
key:
  - string1
  - string2      
  - string3
  - string4
```
or

```
key: ['string1', 'string2', 'string3', 'string4']
```

This is basically me not understanding how to write YAML.

fibonacci1729: Oh yes. I agree. Annotations in k8s is a map of key-value pairs, presumably of type string:string. So we can't unmarshal a traditional `yaml` list. I.e. The Kubernetes annotation type we are (un)marshaling is "map[string]string" and a `list` to `string` conversion in yaml is not supported AFAIK....Then again, it is yaml.

radu-matei: So we have to serialize the list as a string, that makes sense.

Ok, thanks for the clarification!

tests/integration_tests.rs
```diff
@@ -531,6 +586,22 @@ async fn test_wasi_node_should_verify() -> anyhow::Result<()> {
...
+#[tokio::test]
+async fn test_experimental_wasi_http_pod() -> anyhow::Result<()> {
+    const EXPECTED_POD_LOG: &str = r#"{"args":{},"data":"I'm not superstitious, but I am a little stitious.","files":{},"form":{},"headers":{"accept":"*/*","content-length":"50","content-type":"text/plain","host":"postman-echo.com","x-amzn-trace-id":"Root=1-60f9e64c-36256b4878dc640c5af75f7d","x-forwarded-port":"443","x-forwarded-proto":"https"},"json":null,"url":"https://postman-echo.com/post"}"#;
+
+    let test_ns = "wasi-e2e-experimental-http";
+    let (client, pods, mut resource_manager) = set_up_test(test_ns).await?;
+
+    create_experimental_wasi_http_pod(client.clone(), &pods, &mut resource_manager).await?;
+
+    assert::pod_log_equals(&pods, EXPERIMENTAL_WASI_HTTP_POD, EXPECTED_POD_LOG).await?;
```
thomastaylor312, line 598: It think it is a bit safer to just do `pod_log_contains` here in case any of the other output changes (which I believe is why the e2e tests are failing)

thomastaylor312 (changes requested):
This is all looking good. I think the e2e test problem should be easily fixable

thomastaylor312 (approved):

PR #649, opened 2021-07-26
fix: node selector for greet demo

alexeldeib:
it's missing the arch selector present on the other manifests, so it doesn't always hit Krustlet nodes


thomastaylor312 (approved):
Thank you for catching this!

alexeldeib:
oops, accidentally pushed to wrong branch -_-

PR #655, opened 2021-07-30
Added license property to Cargo.toml to specify Apache License 2.0

### LINKED ISSUE #654: wasi provider doesn't specify license in Cargo.toml
The cargo.toml file for the wasi-provider crate doesn't specify the license property, this should be added.

soenkeliebau:
fixes #654 

Signed-off-by: Sönke Liebau <soenke.liebau@gmail.com>

### DIFF
--- a/crates/wasi-provider/Cargo.toml
+++ b/crates/wasi-provider/Cargo.toml
@@ -10,6 +10,7 @@ authors = [
   "Kevin Flansburg <kevin.flansburg@gmail.com>",
 ]
 edition = "2018"
+license = "Apache-2.0"
 name = "wasi-provider"
 publish = false
 version = "1.0.0-alpha.1"


kflansburg:
This seems reasonable, my only concerns:

- Have we confirmed that the current license file is exactly Apache 2 (I think it is)
- Is there a way to avoid having the license defined in two places?

soenkeliebau:
> This seems reasonable, my only concerns:
> 
> * Have we confirmed that the current license file is exactly Apache 2 (I think it is)
I've compared it with the Apache License from this location and the only change (apart from whitespaces due to formatting issues) was this line:

Original:
`Copyright [yyyy] [name of copyright owner]`

Krustlet: 
`Copyright (c) The Krustlet Authors`

Which seems fair enough to me.

> * Is there a way to avoid having the license defined in two places?
Not sure there is to be honest - quite frankly I'm not even sure what the legally binding factor is that makes Krustlet have the Apache License... the license file needs to be there to redistribute the license, but does it also signify "This uses that license" in a legally binding manner? It is accepted practice I guess .. Just skimmed [this article](https://fossa.com/blog/apply-license-open-source-software-project/), which seems to agree, but without any proof either (it does mention Rust).

The property in the Cargo.toml is more a convenience thing so that crates.io and tools using the crate metadata don't have to deduct the license from the file.

kflansburg (approved):
Sounds good then

PR #656, opened 2021-07-30
Device Plugin Integration Test

kate-goldenring:
Provides an integration test for testing Device Plugin support added in ref #621. The test creates a device plugin and schedules a pod that requests the resource advertised by the device plugin.
Specifically, it runs a device plugin that registers with the kubelet device manager and advertises a mock `example.com/gpu` device. The test asserts that the environment variable requested by the device plugin is inserted into the Pod.

~TODO: The **test does not check that volumes requested by the device plugin are mounted.** This is because, the `resources` module mounts them as `HostPathVolumes` which [do not currently support](https://github.com/krustlet/krustlet/blob/main/crates/kubelet/src/volume/hostpath.rs#L7) the equivalent of `container_path` (or mounted path), so the container_path specified by a device plugin is essentially ignored and only the host path is used. This makes testing with `tempdir` hard because `WasmerciserContainerSpec` takes static string arguments. Instead, a file could be created in the `.krustlet` directory since it has adequate permissions, or the `HostPathVolumes` struct could be expanded. Thoughts?~

Volume mount testing added after the merging of ref #674 

Note: Copies over gRPC socket utils from kubelet crate


tests/integration_tests.rs
```diff
@@ -1012,3 +1033,62 @@ async fn test_pod_mounts_with_pvc() -> anyhow::Result<()> {
...
+
+async fn create_device_plugin_resource_pod(
+    client: kube::Client,
+    pods: &Api<Pod>,
+    resource_manager: &mut TestResourceManager,
+) -> anyhow::Result<()> {
+    use k8s_openapi::apimachinery::pkg::api::resource::Quantity;
+    let containers = vec![
+        WasmerciserContainerSpec::named("device-plugin-test").with_args(&[
+            "write(lit:watermelon)to(stm:stdout)",
```
kate-goldenring, line 1048: If volume testing was in place, this would ideally be:
```
WasmerciserContainerSpec::named("device-plugin-test").with_args(&[
    "write(lit:watermelon)to(file:/brb/general.txt)",
    "read(file:/brb/general.txt)to(var:myfile)",
    "write(var:myfile)to(stm:stdout)",
    "assert_exists(env:DEVICE_PLUGIN_VAR)",
]),
```

Cargo.toml
```diff
@@ -68,6 +68,13 @@ serde_json = "1.0"
 tempfile = "3.2"
 tokio = {version = "1.0", features = ["macros", "rt-multi-thread", "time", "process"]}
 tonic = "0.4"
+tokio-stream = { version="0.1", features = ["fs", "net"] }
```
thomastaylor312, line 71: Do these deps not work in `dev-dependencies`?

build.rs
```diff
@@ -0,0 +1,22 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // We need to build the proto files so that we can use them in the tests
+    println!("cargo:rerun-if-changed=crates/kubelet/proto/deviceplugin/v1beta1/deviceplugin.proto");
+
+    let builder = tonic_build::configure()
+        .format(true)
+        .build_client(true)
+        .build_server(true);
+
+    println!("cargo:warning=Build file ran");
```
thomastaylor312, line 10: Left over debug I think from when we were pairing

tests/device_plugin/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        let (kubelet_update_sender, kubelet_update_receiver) = mpsc::channel(3);
+        let devices = self.devices.clone();
+        tokio::spawn(async move {
+            kubelet_update_sender
+                .send(Ok(ListAndWatchResponse { devices }))
+                .await
+                .unwrap();
+            loop {
+                // List and watch should not end
+                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
thomastaylor312, line 51: Super nit: I think we use the uint max value to basically never wake up, but it really isn't too big of a deal in a test

tests/integration_tests.rs
thomastaylor312: Do we need this `write` line right now if we aren't using it?

tests/integration_tests.rs
```diff
@@ -1012,3 +1032,62 @@ async fn test_pod_mounts_with_pvc() -> anyhow::Result<()> {
...
+    println!("Starting DP test");
+    let test_ns = "wasi-e2e-pod-with-device-plugin-resource";
+    let (client, pods, mut resource_manager) = set_up_test(test_ns).await?;
+
+    device_plugin::launch_device_plugin(RESOURCE_NAME, RESOURCE_ENDPOINT).await?;
+
+    // Create a Pod that requests the DP's resource
+    create_device_plugin_resource_pod(client.clone(), &pods, &mut resource_manager).await?;
+    assert::pod_exited_successfully(&pods, DEVICE_PLUGIN_RESOURCE_POD).await?;
+    assert::pod_container_log_contains(
```
thomastaylor312, line 1084: Yeah, I think we can drop this for now as we get all we need from just checking if the pod exited successfully

thomastaylor312 (changes requested):
This is looking great! Just a few minor questions/comments. Also, it looks like the include for the grpc code isn't working on Windows based on the test output

Cargo.toml
kate-goldenring: They are under `dev-dependencies` i believe

tests/integration_tests.rs
kate-goldenring: Good point. Removed in ce2fcb4531488cd7f4fa7086b9add02d36b21c5e

tests/integration_tests.rs
kate-goldenring: Good point! Removed ce2fcb4531488cd7f4fa7086b9add02d36b21c5e

tests/device_plugin/mod.rs
kate-goldenring: fixed! ec47ee860d808c84f10a4a64381a399688f356b4

build.rs
kate-goldenring: Nice catch! ec47ee860d808c84f10a4a64381a399688f356b4

kate-goldenring:
There are a fair bit o

> This is looking great! Just a few minor questions/comments. Also, it looks like the include for the grpc code isn't working on Windows based on the test output

@thomastaylor312, I am just using the same grpc_sock library yet am getting a lot of of compilations errors such as:
```

error[E0046]: not all trait items implemented, missing: `ConnectInfo`, `connect_info`
  --> tests\device_plugin\..\grpc_sock\windows\mod.rs:84:1
   |
84 | impl Connected for UnixStream {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `ConnectInfo`, `connect_info` in implementation
   |
   = help: implement the missing item: `type ConnectInfo = Type;`
   = help: implement the missing item: `fn connect_info(&self) -> <Self as Connected>::ConnectInfo { todo!() }`

error[E0599]: no method named `compat` found for opaque type `impl futures::Future` in the current scope
   --> tests\device_plugin\..\grpc_sock\windows\mod.rs:26:10
    |
26  |         .compat()
    |          ^^^^^^ method not found in `impl futures::Future`
    | 
   ::: C:\Users\krustletci\.cargo\registry\src\github.com-1ecc6299db9ec823\tokio-compat-02-0.2.0\src\lib.rs:151:8
    |
151 |     fn compat(self) -> TokioContext<Self>
    |        ------ the method is available for `impl futures::Future` here
    |
```
Did dependencies change and the grpc_sock lib needs to be updated for both tests and kubelet crate or is this more likely importing error I made?

Cargo.toml
thomastaylor312: Ugh...I thought this was in the other cargo file for krustlet. Ignore me

thomastaylor312:
So for this issue: 

> This makes testing with tempdir hard because WasmerciserContainerSpec takes static string arguments. Instead, a file could be created in the .krustlet directory since it has adequate permissions, or the HostPathVolumes struct could be expanded.

I assume the [`new_in` function](https://docs.rs/tempfile/3.2.0/tempfile/struct.NamedTempFile.html#method.new_in) from tempfile wouldn't work because you need a static name? I also am guessing you already tried something like `tempdir.path().to_string_lossy.as_str()` (something like that)? If that is the case, I think I might change the Wasmerciser stuff to take normal instead of static lifetimes which is outside of the scope of this PR (and I can do in a follow up). So, tl;dr, just put it where you'd like and just make sure you have something to clean up the test file

thomastaylor312:
Update: I think #663 might help you once merged

thomastaylor312:
#663 has been merged, so a rebase should help you solve that problem with the volumes!

kate-goldenring:
Looks like all tests are passing accept the `cargo deny` workflows. Do those need to be fixed elsewhere before this can be merged?

thomastaylor312:
Yeah, those are known issues we are waiting for fixes on. No need to worry about them here

thomastaylor312 (approved):
🎉

PR #663, opened 2021-08-04
ref(tests): Removes static lifetimes

thomastaylor312:
This removes the static lifetimes on the wasmerciser structs. This
_may_ have been interfering with the ability to use string variables,
but it also was too much of a constraint where an explicit lifetime
would work just fine. This should make it much more flexible to use


kflansburg (approved):
LGTM

bacongobbler (approved):

PR #666, opened 2021-08-09
added steps to support bootstrapping by loading filepath to CA

### LINKED ISSUE #642: bootstrapping fails if using filepath to CA instead of raw content
this is fairly minor, but kind of annoying when you need to inject the CA content into the file, and they might technically be coming from different sources

https://github.com/deislabs/krustlet/blob/9efb832edb288ba86822bd62e80e3f9f5e68e37b/crates/kubelet/src/bootstrapping/mod.rs#L70

It'd be nice if this could try to check for either field (certificate_authority or certificate_authority_data), read the file content if it's the former, and then use whatever it found (or fail if neither).

VishnuJin:
@thomastaylor312 hello there,
tried to fix #642 
please let me know if Im on the right track or totally off ☹️

### DIFF
--- a/crates/kubelet/src/bootstrapping/mod.rs
+++ b/crates/kubelet/src/bootstrapping/mod.rs
@@ -10,7 +10,7 @@ use rcgen::{
     Certificate, CertificateParams, DistinguishedName, DnType, KeyPair, SanType,
     PKCS_ECDSA_P256_SHA256,
 };
-use tokio::fs::{read, read_to_string, write, File};
+use tokio::fs::{read, write, File};
 use tokio::io::AsyncWriteExt;
 use tracing::{debug, info, instrument, trace};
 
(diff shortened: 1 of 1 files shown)


crates/kubelet/src/bootstrapping/mod.rs
```diff
@@ -66,14 +66,23 @@ async fn bootstrap_auth<K: AsRef<Path>>(
...
-            .certificate_authority_data
-            .ok_or_else(|| {
-                anyhow::anyhow!(
-                    "Unable to find certificate authority information in bootstrap config"
-                )
-            })?;
+
+        let ca_data = match named_cluster.cluster.certificate_authority {
+            Some(certificate_authority) => {
+                read_to_string(certificate_authority).await.map_err(|e| {
```
VishnuJin, line 72: do we need to encode the string from the file?

thomastaylor312: If we want to do it this way, then yes, this would need to be base64 encoded

thomastaylor312:
This is looking good! Just the small fix around base64 encoding

crates/kubelet/src/bootstrapping/mod.rs
```diff
@@ -66,14 +66,26 @@ async fn bootstrap_auth<K: AsRef<Path>>(
...
-            .certificate_authority_data
-            .ok_or_else(|| {
-                anyhow::anyhow!(
-                    "Unable to find certificate authority information in bootstrap config"
-                )
-            })?;
+
+        let ca_data = match named_cluster.cluster.certificate_authority {
+            Some(certificate_authority) => base64::encode(
+                read_to_string(certificate_authority)
```
thomastaylor312, line 72: Nit: You can use `read` here instead to avoid the string conversion stuff and just get the bytes

thomastaylor312 (approved):
LGTM. Will merge when tests pass. There is an optional nit if you'd like to address it

crates/kubelet/src/bootstrapping/mod.rs
VishnuJin: oh yeah !! can i update and push now ? 
I just forgot it, at first i wanted this to be string before encoding commit.

PR #667, opened 2021-08-13
Do not fail if Docker-Content-Digest not given.

cdmurph32:
AWS ECR (public.ecr.aws), Red Hat (registry.redhat.io), and likely other
registries do not provide the `Docker-Content-Digest` header. As
discussed in https://github.com/krustlet/krustlet/issues/624, this
header is not required by the OCI
[Spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull).


kflansburg:
Is this addressed by #665?

cdmurph32:
@kflansburg It is. Sorry about that. I did this 2 weeks ago and finally got around to opening it. That PR is much better.

PR #668, opened 2021-08-13
Watch for the service account token to be ready

### LINKED ISSUE #403: Properly wait for namespace creation to stabilize in tests
Currently in our integration tests, we are just doing [a delay](https://github.com/deislabs/krustlet/pull/396#discussion_r495082143) to give the controller enough time to create the token for the `default` service account. However, we should do this more smartly. Ideally, we should do a watch (on exactly what will be part of this issue, probably something like the status of the `ServiceAccount` or the tokens in the namespace) to wait for the token to be created

nitishm:
Instead of using a sleep use an actual watcher to ensure that the token
is ready before proceeding.

Closes #403

Signed-off-by: Nitish Malhotra <nitish.malhotra@gmail.com>


bacongobbler:
Hi @nitishm. Sorry for the late response.

Generally speaking we ask for unit tests to accompany any changes. That way we can ensure future changes won't introduce a code regression, and we have some context around why certain changes were made in the past. Would you mind writing some tests? Thanks!

bacongobbler:
Looking at this closer... It might be too difficult to implement tests for the tests. I think we can make an exception here.

tests/test_resource_manager.rs
```diff
@@ -103,8 +104,26 @@ impl TestResourceManager {
 
         // k8s seems to need a bit of time for namespace permissions to flow
         // through the system.  TODO: make this less worse
-        tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
+        // tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
thomastaylor312, line 107: This should be removed, as well as the TODO in the comment just above

thomastaylor312 (changes requested):
Looking great, just one minor change to remove old commented out code and the associated TODO

Issue #660, opened 2021-08-03
Update k8s-openapi to the latest


thomastaylor312:
So it looks like the issues I had with the status updating were likely caused by https://github.com/Arnavion/k8s-openapi/issues/103. Those changes were reverted in https://github.com/Arnavion/k8s-openapi/commit/f0fda041e401e7d3bbb39b19ff305e258688728d and so we'll need to re-add option handling again once the new release comes out. This will also require an update to krator to pick up those changes as well. I do not think we'll need to remove the work for improving status patching as those were good changes to make anyway.

Please note this is currently blocked until a new release of k8s-openapi. However, I do think we should perform this update before releasing our RCs as there could be some minor blocking changes.

thomastaylor312:
Labeled as breaking for now until we know what the update looks like

PR #671, opened 2021-08-30
Make CSRs compatible with Kubernetes v1.22

siegfriedweber:
The `certificates.k8s.io/v1beta1` API version of `CertificateSigningRequest` is no longer served as of Kubernetes v1.22. The `certificates.k8s.io/v1` API version is used instead which is available since v1.19.

The requirements of the new API version were already met:
> Notable changes in certificates.k8s.io/v1:
> * For API clients requesting certificates:
>   * `spec.signerName` is now required (see known [Kubernetes signers](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers)), and requests for `kubernetes.io/legacy-unknown` are not allowed to be created via the `certificates.k8s.io/v1` API
>   * `spec.usages` is now required, may not contain duplicate values, and must only contain known usages
>
> see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#certificatesigningrequest-v122

This change makes kubelet incompatible with Kubernetes v1.18 and lower.

`k8s-openapi` is currently compiled with the feature `v1_21` which is okay because the new API is available since `v1_19`.

### DIFF
--- a/crates/kubelet/src/bootstrapping/mod.rs
+++ b/crates/kubelet/src/bootstrapping/mod.rs
@@ -1,7 +1,7 @@
 use std::{convert::TryFrom, env, io, path::Path, str};
 
 use futures::{StreamExt, TryStreamExt};
-use k8s_openapi::api::certificates::v1beta1::CertificateSigningRequest;
+use k8s_openapi::api::certificates::v1::CertificateSigningRequest;
 use kube::api::{Api, ListParams, PostParams};
 use kube::config::Kubeconfig;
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Thanks for doing this! Just as a note for anyone else who stumbles onto this PR, #660 will bring us to fully operational ~battlestation~ k8s 1.22 status as we still need to update the underlying libraries as well

PR #674, opened 2021-09-01
Enable Device Plugin Volume Mounts

kate-goldenring:
Currently, the piping is not in place for mounts specified by device plugins to be set in Pods. Nor is there support for mounting a host path at a specific container path.

This adds a new type of `VolumeRef` called `DeviceVolume` for specifying mounting a host path volume at a specific container path, as requested by a device plugin. It also modifies the WASI provider to enable appending volumes during the `Resources` state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3831,6 +3831,7 @@ dependencies = [
  "cap-std",
  "chrono",
  "futures",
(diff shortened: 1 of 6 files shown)


thomastaylor312 (approved):
This is great! Thank you for adding the tests

PR #677, opened 2021-09-10
feat(provider): add wasmer provider

bnjjj:
Here is the draft to implement wasmer as a runtime to execute wasm in krustlet. It's working with demos, excepting the demo with postman as I didn't already implemented the experimental http methods. 

I have few questions regarding my implementation. I added a "provider" crate but it seems all the files are almost the same except the method `spawn_wasmtime` and `spawn_wasmer`. Do you think it's a good idea to keep these 2 separated crates or just have only one crate `wasi_runtime` and put `spawn_wasmer` methods and so one inside ?

Todo:

- [ ]  Add an interrupt mechanism
- [ ]  Add wasmer support for wasi-experimental-http
- [ ]  Use upstream version of wasmer when changes are merged


bacongobbler:
Hey @bnjjj! Thanks for working on this one.

> Do you think it's a good idea to keep these 2 separated crates

In general we've started splitting out all of the providers into their own projects. For example the wasmCloud provider now lives at https://github.com/wasmCloud/krustlet-wasmcloud-provider and the wasm3 provider is over at https://github.com/bacongobbler/krustlet-wasm3. Same for the CRI provider: https://github.com/kflansburg/krustlet-cri

Would you be okay with putting this work in its own repository?

Thanks!

bnjjj:
Yes why not, if it's the way to do that I'm okay. What I suggest is to let this PR as a draft and open for now. I will continue my work and maybe when it's done you will be able to give me a little review/overview before closing this PR and put everything in my own repository. What do you think ?

thomastaylor312:
That sounds like a great idea @bnjjj!

technosophos:
WHOA! This is so cool! It'll be great to have a Wasmer provider!

bnjjj:
If you have time to review feel free, I can't implement for now the http experimental part and the interrupt part as some features/documentation are missing on Wasmer. So I plan to put this code into his own repository after your review and then document all current limitations. I will then implement missing features on Wasmer to have all the features for krustlet.

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+    io::{self, Read, Seek, Write},
+};
+
+use anyhow::Error;
+use serde::{Deserialize, Serialize};
+use tokio::fs::File;
+use wasmer_wasi::{VirtualFile, WasiFsError};
+
+/// For piping stdio. Stores all output / input in a byte-vector.
+pub struct FilePipe {
```
bacongobbler, line 13: Out of curiosity, is there a reason not to use the newtype pattern? i.e.

```
pub struct FilePipe(std::fs::File);
```

https://doc.rust-lang.org/book/ch19-04-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+            .finish()
+    }
+}
+
+impl Serialize for FilePipe {
+    fn serialize<S>(&self, _serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        unimplemented!()
```
bacongobbler, line 30: Isn't this the same as just not including the implementation? Why implement `Serialize` then call `unimplemented!()`?

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+use serde::{Deserialize, Serialize};
+use tokio::fs::File;
+use wasmer_wasi::{VirtualFile, WasiFsError};
+
+/// For piping stdio. Stores all output / input in a byte-vector.
+pub struct FilePipe {
+    file: std::fs::File,
+}
+
+impl Debug for FilePipe {
```
bacongobbler, line 17: Shouldn't `#[derive(Debug)]` be sufficient?

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+        unimplemented!()
+    }
+}
+
+impl<'de> Deserialize<'de> for FilePipe {
+    fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        unimplemented!()
```
bacongobbler, line 39: Same question here... Why define this, then mark it as unimplemented?

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+impl TryFrom<File> for FilePipe {
+    type Error = Error;
+
+    fn try_from(value: File) -> Result<Self, Self::Error> {
+        Ok(Self {
+            file: value
+                .try_into_std()
```
bacongobbler, line 49: So this will immediately attempt to convert a File to a StdFile without allowing async write operations to complete IIRC. Is there a technical reason why we can't implement From instead of TryFrom to allow a cleaner migration?

https://docs.rs/tokio/1.12.0/tokio/fs/struct.File.html#method.try_into_std

crates/wasmer-provider/src/pipes.rs
```diff
@@ -0,0 +1,102 @@
...
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            "can not seek in a FilePipe",
+        ))
+    }
+}
+
+impl VirtualFile for FilePipe {
+    fn last_accessed(&self) -> u64 {
+        0
```
bacongobbler, line 81: is there a reason these are left unimplemented?

crates/kubelet/src/config.rs
```diff
@@ -66,6 +68,8 @@ pub struct Config {
     /// device plugins lives. This is also where device plugins
     /// should host their services.
     pub device_plugins_dir: PathBuf,
+    /// Wasm runtime to use to execute your wasm
+    pub provider: Provider,
```
bacongobbler, line 72: So generally speaking we've avoided adding internal implementation details like a `Provider` type directly into the `kubelet` crate. The goal of this crate is to implement the kubelet API without leaking any implementation details of Krustlet. That way clients can use the `kubelet` crate to interact with the Kubernetes API and impersonate a kubelet any way they choose.

Since you're just introducing this type to check a string value.... This seems more appropriate in the provider code itself.

crates/kubelet/src/config.rs
```diff
@@ -413,6 +424,29 @@ where
...
+}
+impl FromStr for Provider {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "wasmtime" => Ok(Self::WasmTime),
+            "wasmer" => Ok(Self::Wasmer),
+            other => Err(anyhow::anyhow!(
+                "unknown provider '{:?}', please choose between 'wasmer' and 'wasmtime'",
```
bacongobbler, line 443: There are other providers out there than just these two by the way. For example:

https://github.com/bacongobbler/krustlet-wasm3
https://github.com/kflansburg/krustlet-cri
https://github.com/wasmCloud/krustlet-wasmcloud-provider
https://github.com/deislabs/krustlet-wagi-provider
https://github.com/bacongobbler/krustlet-wasmdome
https://github.com/soenkeliebau/stackable_krustlet

These are just some of the examples I could find through Github's search engine.

I don't think we should be in the business of listing "known providers" in the crate anyways.

crates/wasi-provider/src/states/container/waiting.rs
```diff
@@ -17,7 +17,7 @@ use super::running::Running;
 use super::terminated::Terminated;
 use super::ContainerState;
 
-pub const MAX_CONNCURRENT_REQUESTS_ANNOTATION_KEY: &str =
+pub const MAX_CONCURRENT_REQUESTS_ANNOTATION_KEY: &str =
```
bacongobbler, line 20: good catch.

bacongobbler (commented):
Looks like a good first step! Left a bit of feedback around some of the code. Otherwise I think it's in good shape to split out into a separate project. Thanks for your contributions!

crates/wasmer-provider/src/pipes.rs
bnjjj: It should yes, just a bad copy/paste I think

crates/wasmer-provider/src/pipes.rs
bnjjj: Same reason, The VirtualFile trait need to implement Serializer and Deserialize. But in our case we are using VirtualFile as a pipe, then as it's not treated as a real file then Serialize and Deserialize won't be used at all, that's why we don't need to implement it. I think a better approach would be to change a little bit the wasmer api

PR #678, opened 2021-09-14
Update k8s-openapi to 0.13 and k8s 1.22

### LINKED ISSUE #660: Update k8s-openapi to the latest

olivierlemasle:
As k8s-openapi reverted in https://github.com/Arnavion/k8s-openapi/commit/f0fda041e401e7d3bbb39b19ff305e258688728d to its use of `Option<>`, this commit reverts most of 4846df7982fe4649e2e038ed7f69b04f148063c0.

Also update kube and kube-runtime to 0.60.

**Note:** This depends on krator PR: https://github.com/krator-rs/krator/pull/49 so we must wait for this PR to be merged and a krator release done.

Fixes #660

### DIFF
--- a/crates/kubelet/src/container/mod.rs
+++ b/crates/kubelet/src/container/mod.rs
@@ -138,23 +138,23 @@ impl Container {
     }
 
     /// Get arguments of container.
-    pub fn args(&self) -> &Option<Vec<String>> {
-        &self.0.args
+    pub fn args(&self) -> Option<&Vec<String>> {
+        self.0.args.as_ref()
     }
 
     /// Get command of container.
-    pub fn command(&self) -> &Option<Vec<String>> {
-        &self.0.command
+    pub fn command(&self) -> Option<&Vec<String>> {
+        self.0.command.as_ref()
     }
 
     /// Get environment of container.
-    pub fn env(&self) -> &Option<Vec<k8s_openapi::api::core::v1::EnvVar>> {
-        &self.0.env
+    pub fn env(&self) -> Option<&Vec<k8s_openapi::api::core::v1::EnvVar>> {
+        self.0.env.as_ref()
     }
 
     /// Get environment of container.
-    pub fn env_from(&self) -> &Option<Vec<k8s_openapi::api::core::v1::EnvFromSource>> {
-        &self.0.env_from
+    pub fn env_from(&self) -> Option<&Vec<k8s_openapi::api::core::v1::EnvFromSource>> {
+        self.0.env_from.as_ref()
     }
 
     /// Get image of container as `oci_distribution::Reference`.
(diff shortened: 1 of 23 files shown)


thomastaylor312:
Thanks for doing this @olivierlemasle! We'll try to get the krator thing merged as quickly as possible. Just trying to figure out how to get around a breaking change in our tests introduced by the latest rust version

thomastaylor312:
Ok, we just published krator 0.5 with the changes. You should be able to pull that in now

olivierlemasle:
@thomastaylor312 Thanks, I've update the PR.

crates/kubelet/src/container/mod.rs
```diff
@@ -138,22 +138,22 @@ impl Container {
     }
 
     /// Get arguments of container.
-    pub fn args(&self) -> &Vec<String> {
+    pub fn args(&self) -> &Option<Vec<String>> {
```
thomastaylor312, line 141: In the original API all of these functions were `Option<&T>` rather than borrowed options. Mind switching the signatures in this file to use that instead with `as_ref`?

thomastaylor312 (changes requested):
Looking good! Just one minor change I noticed while I was looking it over

crates/kubelet/src/container/mod.rs
olivierlemasle: I don't understand. AFAIK, it was the original signature before 4846df7:

https://github.com/krustlet/krustlet/blob/da85061ec2ef809590cad023ecdefc6891102c34/crates/kubelet/src/container/mod.rs#L141-L143

olivierlemasle: I persist :smiley:: you **did** have `args` in 0.7, and its return type was `&Option<Vec<String>>`: cf https://github.com/krustlet/krustlet/blob/v0.7.0/crates/kubelet/src/container/mod.rs#L141-L143

However, no problem to change it and have the borrowed inner type. I'm updating the PR.

thomastaylor312: Oh that was in container 🤦 . Sorry. We should have caught that earlier. Thanks for doing it anyway!

olivierlemasle:
Rebased on `main` to fix `cargo deny` errors.

thomastaylor312 (approved):

PR #682, opened 2021-09-24
OCI Distribution Improvements

kflansburg:
This PR includes a number of improvements for OCI distribution:

* HEAD requests are attempted for fetching manifest digest, with a fallback to GET (some providers to do not include the digest in the HEAD response). This appears to be preferred by registry providers such as DockerHub, because they do not apply rate limits to HEAD requests. I have confirmed that `fetch_manifest_digest`, including any authentication API calls, does not incur ratelimits. 
* JWT tokens are now cached by (registry, repository, operation), instead of just registry. Previously, requests using the same client to perform different operations, or access different repositories in the same registry would fail, because tokens are scoped to the repository. 
* Tokens are now stored in an expiring cache, which inspects the JWT for its expiration timestamp, and returns `None` if a token is in the cache but is expired. 

Combined, these changes should make it possible to reuse the OCI client over long periods of time, with more efficient registry API usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -541,6 +541,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "crypto-mac"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1d1a86f49236c215f271d40892d5fc950490551400b02ef360692c29815c714"
+dependencies = [
+ "generic-array",
+ "subtle",
+]
+
+[[package]]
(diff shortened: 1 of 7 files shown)


bacongobbler:
These are all great changes, and they map back to OCI's expectations as to how these JWTs should be treated.

Have you had a chance to look at the failing Ci checks yet? Looks like legitimate type type mismatch and compiler errors.

flavio:
I looked at the changes too, this looks good :clap:

PR #694, opened 2021-10-20
run `cargo update`

bacongobbler:


thomastaylor312:
I think I figured it out. Because we are on an alpha build of notify (to pick up the better features in the 5.0 series), the latest update broke it. I am fixing it now and will push to this branch

Issue #690, opened 2021-10-19
Move oci-distribution to its own repository


bacongobbler:
A number of community members have asked if oci-distribution can be maintained as a separate project so that it can cut new releases independently from the krustlet project. Moving the source code into its own repository under the krustlet org would be the simplest option.

To be completed prior to #616 so that a new release can be cut from the new repository

flavio:
Thanks a lot for that! :heart:

bacongobbler:
Can be closed now that #695 has been merged.

PR #695, opened 2021-10-20
remove oci-distribution

bacongobbler:
new location: https://github.com/krustlet/oci-distribution

related to #690


thomastaylor312:
@bacongobbler I think with a rebase this should be good to go

bacongobbler:
After re-fetching from upstream:

```
><> git rebase main
Current branch remove-oci-distribution is up to date.
```

thomastaylor312:
I thought we had fixed the cargo deny stuff recently

thomastaylor312:
@bacongobbler Let's go ahead an ignore the `cargo deny` stuff for now. We should be able to fix that before we release the beta. So this should be good to merge for now

rickrain:
Any thoughts on when this PR can be merged?

I'm working on a custom kubelet implementation and using the `kubelet` crate.  In my `cargo.toml`, I have a git package dependency to this repo instead of a crates.io package dependency so I can pickup the most recent `kubelet` code that has not yet formally released (see #707).

In my implementation, I'm also making use of the `GenericProvider` and therefore leveraging the generic pod state implementations (`Registered`, `ImagePull`, etc.) - _thank you_ for those btw!  And, because `kubelet` is still referencing `oci-distribution` as a git package reference, I had to do the same in my project.

Unfortunately, I ran into the issue regarding _manifest lists_ not being supported yet in `oci-distribution`, and then found [this PR](https://github.com/krustlet/oci-distribution/pull/3/) in the new `oci-distribution` repo that handles this for me.  Hopefully that will merge into main soon too.

The ideal scenario would be (in order):
1. Merge [this PR](https://github.com/krustlet/oci-distribution/pull/3/) in the `oci-distribution` repo and generate a release for crates.io.
2. Update [cargo.toml](https://github.com/krustlet/krustlet/pull/695/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) in this PR that we're discussing with the new release/version of `oci-distribution`.
3. Create a new `kubelet` release for crates.io.  Although, I believe just getting 1 and 2 would unblock me.

I realize I could provide my own `ImagePull` state, but would prefer not to since the work has been done.

thomastaylor312:
Thanks for reaching out @rickrain! So normally our rule is that maintainers merge their own PRs once reviewed. The other PR you referred to is awaiting review from someone with more OCI experience than myself. So I think we can do all of those things.

I'll start out by pinging @bacongobbler and see if he is ok with me merging this. Then we can do a follow up PR with a version bump as soon as the other feature is added

thomastaylor312:
Just heard back from @bacongobbler. Going to go ahead and merge this

bacongobbler:
I'm okay with merging this. Last we revisited this PR, we were waiting on fixing up Windows CI failures, but it looks like that's stalled out. Good to go otherwise

rickrain:
Thanks @thomastaylor312 and @bacongobbler for the quick response and action.

PR #702, opened 2021-11-22
Ensure a more recent version of tokio is used

flavio:
This is required to address [RUSTSEC-2021-0124](https://rustsec.org/advisories/RUSTSEC-2021-0124.html).

This is currently WIP:

  * [ ] wait for https://github.com/krator-rs/krator/pull/63 to be merged
  * [ ] update kubelet depedencies on windows: remove tokio-compat-02 dependency. I suspect this is no longer needed

Address issue https://github.com/krustlet/krustlet/issues/701

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,16 @@ version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
 dependencies = [
- "gimli",
+ "gimli 0.25.0",
+]
+
+[[package]]
+name = "addr2line"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
+dependencies = [
+ "gimli 0.26.1",
 ]
 
 [[package]]
@@ -52,9 +61,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.44"
+version = "1.0.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61604a8f862e1d5c3229fdd78f8b02c68dcf73a4c4b05fd636d12240aaa242c1"
+checksum = "38d9ff5d688f1c13395289f67db01d4826b46dd694e7580accdc3e8430f2d98e"
 
 [[package]]
 name = "async-recursion"
@@ -64,7 +73,7 @@ checksum = "d7d78656ba01f1b93024b7c3a0467f1608e4be67d725749fdcd7d2c7678fd7a2"
 dependencies = [
  "proc-macro2",
  "quote 1.0.10",
- "syn 1.0.80",
+ "syn 1.0.81",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


kflansburg:
Once the following merge I will publish a new crate version for Krator:

- https://github.com/krator-rs/krator/pull/64
- https://github.com/krator-rs/krator/pull/65

kflansburg:
@flavio `krator` v0.6.0 is published with the new version of `tokio`.

kflansburg:
I believe that the advisories are a known issue, but the windows tests are probably caused by the same `native-tls` feature that they were on `krator`.

flavio:
> I believe that the advisories are a known issue, but the windows tests are probably caused by the same `native-tls` feature that they were on `krator`.

I've just tried to build with the latest changes I submitted on a windows box. Everything builds fine.

However, the really old tokio 0.2 crate is used on windows because [kubelet/src/grpc_sock/windows/mod.rs](https://github.com/krustlet/krustlet/blob/main/crates/kubelet/src/grpc_sock/windows/mod.rs) makes heavy use of it.

It looks like this piece of code should need some refactoring...

kflansburg:
> I've just tried to build with the latest changes I submitted on a windows box. Everything builds fine.
> 
> However, the really old tokio 0.2 crate is used on windows because [kubelet/src/grpc_sock/windows/mod.rs](https://github.com/krustlet/krustlet/blob/main/crates/kubelet/src/grpc_sock/windows/mod.rs) makes heavy use of it.
> 
> It looks like this piece of code should need some refactoring...

The Windows failure looks like it is related to the Windows runner and not our e2e tests. Maybe a flake? I will rerun it. 

Regarding Tokio 0.2, yes we had to use the compatibility layer when Tokio 0.3 came out, my understanding is that there were some API changes that would not be easy to work around with that module specifically. Will that prevent us from addressing this CVE? @thomastaylor312 would know more.

kflansburg:
@bacongobbler @thomastaylor312 The rerun of Windows e2e failed. It looks like Docker never becomes ready. I'm not familiar with how this runner is configured or who hosts it.

thomastaylor312:
Yeah, the custom windows runner is currently experiencing issues. @bacongobbler and I are working on fixing it

@flavio The tokio 0.2 stuff is there because there is no easy way to add async unix sockets on Windows. Someone had written something for 0.2, which is what we adopted into the project. Because tokio completely removed `mio` from the public API, we couldn't just update the mio stuff and then wrap it in a `PollEvented` like we used to. I sank a month into it and got nowhere. If someone who knows async on windows and tokio well, then they could probably write a new version of async sockets

bacongobbler:
What if we just dropped Windows support?

thomastaylor312:
Like, entirely? I mean, I'd love to, but that kinda defeats the purpose of having this do Wasm on any platform

bacongobbler:
> that kinda defeats the purpose of having this do Wasm on any platform

I agree. That being said, our early experiments with Krustlet have already proven that. We proved that an app compiled to WebAssembly can run on any platform and we have demonstrations/recordings on YouTube to support that statement. It's not like dropping Windows support due to outdated library support or CI build failures disproves that point.

From my point of view, I'm more concerned about letting others build their own providers on top of core, and I'd hate to see new community members writing providers on top of a core with major security flaws. I'd rather us move to resolve the CVEs in core first, then come back and re-introduce Windows support if there's enough demand for it again.

Thoughts?

thomastaylor312:
Let's open up a separate issue so as to not clog up this PR further and then discuss at the community meeting

flavio:
Dropping the windows support would definitely make this security fix easier, and also prevent over flaky tests I've seen on other PRs.

I cannot attend the community meetings due to timezone issues. Please me posted here or just write on the slack channel the final outcome of the discussion.

In the meantime, I'm changing the status of this PR from "draft" to "reviewable".

PR #705, opened 2021-12-08
feature #369 add steps to Cleanup on bootstrap fail

### LINKED ISSUE #369: Cleanup on bootstrap fail
Right now, if the bootstrap process fails part of the way through, it leaves a bunch of invalid resources around in the cluster (like the old CSR). As a new user, it would be helpful if I didn't have to clean these up manually after a bootstrap fail.

Requirements:
- Clean up auth certificate CSR on failure

Questions:
- Should we clean up anything else? Cleaning up local config sounds dangerous, as it still could be valid and would force bootstrapping again

VishnuJin:
fixes #369 
@thomastaylor312 is this the right approach ?


VishnuJin:
Im not sure why e2e fails, It passes on my local 😕
image

thomastaylor312:
@VishnuJin you can download the logs from the actions pipeline. Don't worry about the windows-e2e as that has been having issues

PR #714, opened 2022-02-09
Modify 'std::u64::MAX' to 'u64::MAX'

zyy17:
'u64::MAX' is the recommended way to represent the largest number of u64 type(reference: https://doc.rust-lang.org/std/u64/constant.MAX.html) and 'std::u64::MAX' is now deprecated in a future Rust version.

Signed-off-by: zyy17 <zyylsxm@gmail.com>


bacongobbler (approved):
LGTM

PR #740, opened 2023-02-24
changed k8s api endpoints from v1beta1 to v1 in order to work on k8s>…

doagl:
I noticed that krustlet doesn't start on newer kubernetes versions (>1.21), due to deprecated api endpoints (to be more specific: v1beta1 is EOL for CertificateSigningRequest on k8s 1.22, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#certificatesigningrequest-v122).

With these changes i was able to deploy the greet-pod on krustlet with a local k3d and kubernetes 1.25.

Thanks for your cool tool!


thomastaylor312:
Hi there @doagl, thanks for the contribution! So currently Krustlet is pretty much unmaintained. https://github.com/containerd/runwasi has become the more recommended way to enable running wasm inside of k8s!

ClintCombs:
@thomastaylor312 would it make sense to add a note about the state of this project and link to https://github.com/containerd/runwasi in the README?

doagl:
A note about the projects status sound's good. The PR might be worth pulling nonetheless, because then krustlet at least works with v1 k8s APIs.

thomastaylor312:
Yeah, makes sense. Could you please sign your commit off for DCO real quick?

